PR #78, opened 2014-08-08
Added mapped remote filesystem support for workspace browsing.

dpospisil:
This pull request adds capabilty to browse workspaces of jobs being built using docker containers whose workspace is available on master node (via mapped volumes, network shares, etc.).

User just need to specify in image template the directory on master node where the container remote FS is mapped.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
@@ -192,7 +192,7 @@ public class DockerSlave extends AbstractCloudSlave {
      * @throws IOException
      */
     private void addJenkinsAction(String tag_image) throws IOException {
-        theRun.addAction( new DockerBuildAction(getCloud().serverUrl, containerId, tag_image) );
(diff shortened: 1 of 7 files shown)


nixilla:
Can you explain how this works, please? I'm having problems accessing workspace, after build is finished (during build is OK). This is the problem I'm experiencing is described here: http://stackoverflow.com/questions/29296102/jenkins-getting-test-results-from-docker-slave

KostyaSha:
```User just need to specify in image template the directory on master node where the container remote FS is mapped.```
@nixilla could you be more specific in what exactly not clear?
You need share directory from master docker-slaves and then point docker containers on this directory.

danielkraaijbax:
This is perhaps the wrong issue to address this. But what @nixilla is saying is still correct. As stated by @dpospisil during the build the workspace is available. But after this is not. 

Using the copy-to-slave plugin to copy the whole workspace from the slave to the master jenkins also does not work.

KostyaSha:
@danielkraaijbax you duplicates this thread https://github.com/jenkinsci/docker-plugin/issues/195#issuecomment-89530682

danielkraaijbax:
@KostyaSha Ah, sorry. Didn't catch that one. It clears up some stuff.

Issue #91, opened 2014-08-26
Commit / push if failure


hcguersoy:
In addition to the 'commit / push on succes' it would be nice to have a functionality to commit and / or push a image of a failed build to inspect it.

In my case, the successful build images are not very interesting but it would be nice to have the possibility to inspect why a build fails.

hcguersoy:
IMO it would be sufficient to commit / tag and optionally push.
Pushing large amounts of build containers into a private docker registry brings the problem of house keeping (in docker registry it's still not supported to delete images) so this should be optional.

rskang:
@hcguersoy Yes we just commit / tag on failure, no pushing. For the exact reasons you mention.
Will hopefully get some time to rebase our code onto upstream and get it up here later this week.

rskang:
Pull request here: https://github.com/jenkinsci/docker-plugin/pull/132 (only does tag / commit, no pushing related code at all)

pjdarton:
Closing due to lack of response on PR

PR #98, opened 2014-10-10
add options to push container on unstable/failed builds

michaelheyvaert:
Add two options to push the container at the end of a build in case the build is unstable or the build fails.
This changes the original behavior, the original push if successful check box triggered a push even if the build failed.

I use these new flags to store the builds for inspection on our private docker registry in case of test or build failures.

This merge request implements #91.


KostyaSha:
My guess https://github.com/jenkinsci/docker-plugin/pull/132#issuecomment-70603885 was right :)

shipperizer:
:+1:

samrocketman:
This feature would be useful for debugging.  e.g. temporarily enable pushing failed images to debug and then disable it when done.  I remember a time when I needed this.  I got around it in other non-intuitive ways.

PR #128, opened 2014-12-22
pass environment additions to docker container

### LINKED ISSUE #107: Allow to specify environment variables of container
Provide an input box that would specify the environment options just like docker -e option: 
  -e, --env=[]               Set environment variables

It can work the same way as a volumes input.

alari:
fixes #107


tunix:
Hi, despite it has been merged, I can't see this field in my Jenkins setup. I've docker-plugin version 0.8 -- do you need to publish a new package incl. this pr?

tunix:
Hi, I downloaded source, created the package and installed the hpi file. Environment input started to show. When I save it, it updates config.xml in the background but the input remains empty when I go back to Manage Jenkins -> Configure System.

I try to echo a variable (namely DOCKER_HOST) in the running slave container but the variable is empty. Is this a bug or am I doing something wrong?

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
```diff
@@ -40,6 +40,7 @@
     public final String dockerCommand;
     public final String volumesString;
     public final String volumesFrom;
+    public final String environmentsString;
```
KostyaSha: :-1: long nam environmentsString

tunix:
:+1: up

KostyaSha:
Also there is no help files.

KostyaSha:
For what this PR at all? Just name variable where host address will be stored?
This shouldn't be an option at all, either append variable to slave either add token macro.

PR #131, opened 2015-01-06
Adding HostConfig to ContainerConfig to fix Bind Volumes

evilbwil:
as to the docker API documentation (1.16) the HostConfig part of the create container call  is used to setup a container with bind volumes (binding a host dir to a container dir). As of now the HostConfig is only passed into the start container call (in which it get discarded in docker 1.4).

this needs https://github.com/nirima/jDocker/pull/5


magnayn:
Ok - I'm trying to merge in the move to docker-java/docker-java, I'll try this patch when that's done + merge.

magnayn:
Ok - I've merged back to docker-java, I'm not sure if it's configuration also requires items from createHostConfig into  createContainerConfig -- could you take a look if you have a 1.4 host?

evilbwil:
i tested it against:
Docker version 1.4.0, build 4595d4f

PR #134, opened 2015-01-09
Split lxc-conf field on comma instead of space

rskang:
Splitting on space causes issues with config options such as: lxc.cgroup.devices.allow=c 10:229 rwm
which is needed to allow FUSE mounts inside the containers

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -209,7 +209,7 @@ public abstract class DockerTemplateBase {
         List<HostConfig.LxcConf> temp = new ArrayList<HostConfig.LxcConf>();
         if( lxcConfString == null )
             return temp;
-        for (String item : lxcConfString.split(" ")) {
+        for (String item : lxcConfString.split(",")) {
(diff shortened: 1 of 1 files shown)


KostyaSha:
:+1: never use space as delimiter in future, already fixed such issues in other plugin

KostyaSha:
Will your change be backward compatible with old configurations?

magnayn:
What happens if a user actually does want to pass a comma?

Suspect it probably needs a multiple value UI control, or we need to allow some kinds of escaping (e.g if you wanted a space, you have to put "" round the term). Not sure what other plugins do..

KostyaSha:
@magnayn it pretty simple:
1) CSV works in 99% cases.
2) just change form control to textarea and during parse .split() with lineseparator
@magnayn could you appear in IRC?

rskang:
@magnayn I checked the lxc docs, there didn't appear to be any params that would be likely to have a comma in them. @KostyaSha's suggestion of a text field and splitting on newlines may be the best option though.

magnayn:
Lets do this for now - sounds like needing to specify a , is much rarer than needing to specify a " "; we can always update the UI later.

KostyaSha:
@magnayn please don't merge PR that will broke old configs. I don't see code that must update old configuration to new.

PR #139, opened 2015-01-15
Extension to start Slaves in Exclusive mode

imasen:
This is a extension to run labeled DockerSlaves in "Exclusive"-Mode. It makes it possible to have labeled jobs to run on specific slaves and general, unlabeled jobs to run on more general slaves.

E.g.

Job-Label: "android" --> runs on DockerSlave-Label in "Exclusive"-Mode: "android"
Job-Label: "" --> runs on DockerSlave-Label in "Normal"-Mode: ""


src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -214,11 +214,14 @@ public DockerTemplate getTemplate(String template) {
...
-     * Gets {@link DockerTemplate} that has the matching {@link Label}.
+     * Gets {@link DockerTemplate} that has the matching {@link Label}; if {@link Label} is null and togglemode false,
+     * then get this template.
      */
     public DockerTemplate getTemplate(Label label) {
         for (DockerTemplate t : templates) {
-            if(label == null || label.matches(t.getLabelSet())) {
+            if(label == null && !t.getToggleMode()) {
+                return t;
+            } else if (label.matches(t.getLabelSet())) {
```
KostyaSha: Merge to one if?

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
```diff
@@ -57,6 +57,7 @@
     public final String bindPorts;
     public final boolean bindAllPorts;
     public final boolean privileged;
+    public final boolean togglemode;
```
KostyaSha: Why not call var 'inclusive' or 'exclusive' (depends on what better fit to code)?

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
imasen: Looks good for me.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
imasen: Good idea, I think exclusive should be good for this. So if exclusive = false then it's Normal-Mode, if true then its Exclusive-Mode.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
thomassuckow: The elseif will give a Null pointer exception if the slave is set to exclusive

+1 to combining
```
if( (label == null && !t.getExclusiveMode()) || (label != null && label.matches(t.getLabelSet())) ) {
```

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
thomassuckow: I agree with exclusiveMode. Then getToggleMode becomes getExclusiveMode

KostyaSha: Why mode? exclusiveLabel will be more accurate then.

KostyaSha:
Also add help page for this parameter. You need place i.e. "exclusiveLabel-help.html" near jelly file.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
imasen: exclusiveMode is better, because it's called mode in Node.Mode JavaDoc: http://javadoc.jenkins-ci.org/hudson/model/Node.Mode.html

magnayn:
Looks good, but my memory of the Jenkins API has gotten a bit fuzzy (so apologies if this is obvious!).

Isn't the current behaviour the same? I.E: if you specify no label for a job, every template matches, but if you specify a named label, only templates with that name match?

I'd thought that exclusive mode didn't make any difference, because we forced slaves to only ever accept one job before terminating?

KostyaSha:
@magnayn when jenkins asks cloud provider whether it can provision it passes label. In this plugin default behavior for case when label is null (= not set in job) is to accept this job for build.
PR author wants not to run docker containers when he has other slaves. So this change stop accepting jobs with null label. 
I still insist not to use word "mode" because any boolean checkbox is "mode" for something. Better replace it with word "Label" to show on what it really influence.

KostyaSha:
Btw, i'm testing locally another patch for labels - ability to use the same label for multiple templates.

imasen:
@magnayn 

You are right. If you specify no label for a job, every template will match. I our environment this is a no-go. We have over 3000 Jobs configured on our CI. Around 80% are JEE Jobs with no label, the other 20% are android, iOS and C# Jobs with a specific label. For every label we have specific Docker-Images to build them, also for the non-Labled JEE jobs. So we need to run the non-labled jobs on a kind of "default"-Slave, the must not for example on a android slave.

magnayn:
Ah - OK, I think I understand - you need to be able to specify nodes that only are valid candidates if they have been *explicitly* specified. 

Cool, LGTM. Could you squash the history (I don't think we need to know it was once togglemode) and I'll merge.

imasen:
@magnayn 

I squashed the commits, I hope I didn't messed it up.

imasen:
I merged the master-branch changes of yesterday to solve the conflicts. CI-Job is hanging, "Waiting to hear about 3d40e14" and I can't start a new one.

KostyaSha:
Better rebase.

src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/template.jelly
```diff
@@ -84,6 +84,10 @@
             <f:checkbox/>
         </f:entry>
 
+        <f:entry title="${%Run slave in exclusive mode}" field="exclusiveMode">
+            <f:checkbox/>
+        </f:entry>
```
KostyaSha: Will check later, but probably we should reuse existed dropdown control that already has clever description.

KostyaSha: Help page is here src/main/webapp/help/system-config/master-slave/usage.html
jelly: hudson/slaves/DumbSlave/configure-entries.jelly
existed form: lib/form/slave-mode.jelly
Just reuse it and change variable to real Mode object.

KostyaSha:
:-1: until it will be reworked to use https://jenkins-ci.org/maven-site/jenkins-core/jelly-taglib-ref.html#form:slave-mode

imasen:
I changed the type and configuration of this. Please  have a look.

KostyaSha:
1) Don't add forms, just use them. slave-mode belong to core /lib/forms you shouldn't copy-paste form files.
2) Sonatype url shouldn't be a part of change.
3) Rebase commits and remove mentions about boolean.

imasen:
I have two problems getting this PR done. The plugin creates two containers if I start a build WITH label. One container has the correct label (and the job runs on it), but the other one created has no label (so it's don't needed for building the running job). I don't understand why it creates two slaves, if one build is running and why it creates one slave with a label (null) which does not match the job-label. I found out, that Jenkins triggers the provision method in DockerCloud twice. Once with label unset (null) and once with the correct label, I think therefore it creates two slave. This behavior does not seem to occur because of my changes, do we have a general problem?

The second problem is the jelly-form for slave-mode. It shows the wrong configuration, which does not match the configuration in config.xml. When I press save on the configuration screen, it saves the configuration correctly, but on a refresh it shows the default-entries. Do you have a quick fix?

KostyaSha:
1) Debug your issues, check what number of existed capacity is and what is happening. 
2) Default entries will be shown if you forgot to create getter method.

KostyaSha:
2) Or doesn't bind right form to variable. 
Just debug your issues in debug mode.

thomassuckow:
@imasen Is it by chance a matrix build? A matrix build starts a label-less job (unless you set the restriction on the job) and then starts its configurations with the labels.

KostyaSha:
Fix for lightweight jobs was added to master. I will try rebase/fix form/test/merge this PR

KostyaSha:
@magnayn FYI: i bit fixed your perms, now you can assign labels/issues/admin this repo

KostyaSha:
Repository layout was changed, i will pick changes to my 0.9 branch for testing.

PR #143, opened 2015-01-22
Use released docker-java version

guilhem:
/!\ error in build process

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,9 +24,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.nirima.docker-java</groupId>
+            <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java</artifactId>
-            <version>0.10.5-SNAPSHOT</version>
+            <version>0.10.5</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>


guilhem:
https://gist.github.com/guilhem/59fe9decf673ae1e8a4f

thomassuckow:
Should probably also think about adding the docker-java snapshot repository to the pom.

pom.xml
```diff
@@ -24,9 +24,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.nirima.docker-java</groupId>
+            <groupId>com.github.docker-java</groupId>
```
KostyaSha: This fix seems right

pom.xml
```diff
@@ -24,9 +24,9 @@
...
             <artifactId>docker-java</artifactId>
-            <version>0.10.5-SNAPSHOT</version>
+            <version>0.10.5</version>
```
KostyaSha: This wrong, SNAPSHOT version for not released plugin version is ok, because it needs fixes for functionality.

guilhem: `0.10.5` version is released... if we want another one, we may use `0.10.6-SNAPSHOT`

KostyaSha: Right, then ok

imasen:
Could someone upload the jar for the version of  @magnayn ? It's hard to develop with red lines in IDE.

thomassuckow:
His is in OSS
 <repository>
            <id>oss</id>
            <url>http://oss.sonatype.org/content/groups/public/</url>
            <releases>
               <enabled>false</enabled>
            </releases>
            <snapshots>
               <enabled>true</enabled>
            </snapshots>
        </repository>

magnayn:
My changes are also in github - I'm slowly sending PRs to docker-java to
try to get back to mainline.

KostyaSha:
@magnayn please fix master or branch latest released version so people may continue fixing their problems. You can release your any future not tested features as beta release from master then.

CedricGatay:
Adding the repository field worked for me to be able to build properly locally.

magnayn:
N/A.

We are nearly back to mainline. When that happens we can flip back.

PR #142, opened 2015-01-21
Added help files with documentation from [1] to make it easier to config...

imasen:
...ure the plugin.

[1] https://wiki.jenkins-ci.org/display/JENKINS/Docker+Plugin

PR #145, opened 2015-01-25
Provisioning improvements

thomassuckow:
Fix not downloading image if not present on server

Log the label and image name when provisioning

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -195,7 +195,7 @@ public class DockerCloud extends Cloud {
     }
 
     private DockerClient buildConnection() {
-        LOGGER.log(Level.FINE, "Building connection to docker host " + name + " URL " + serverUrl);
+        LOGGER.log(Level.FINE, "Building connection to docker host \"{0}\" at: {1}", new Object[]{name,serverUrl});
 
         return DockerClientBuilder.getInstance(getDockerClientConfig()).build();
     }
@@ -219,11 +219,13 @@ public class DockerCloud extends Cloud {
     public synchronized Collection<NodeProvisioner.PlannedNode> provision(Label label, int excessWorkload) {
         try {
 
-            LOGGER.log(Level.INFO, "Excess workload after pending Spot instances: " + excessWorkload);
+            LOGGER.log(Level.INFO, "Asked to provision {0} slave(s) for: {1}", new Object[]{excessWorkload,label});
 
             List<NodeProvisioner.PlannedNode> r = new ArrayList<NodeProvisioner.PlannedNode>();
 
             final DockerTemplate t = getTemplate(label);
+
+            LOGGER.log(Level.INFO, "Will provision \"{0}\" for: {1}", new Object[]{t.image,label});
 
             while (excessWorkload>0) {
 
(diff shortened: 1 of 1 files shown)


KostyaSha:
If you are already changing logging, please exclude slow logging.

thomassuckow:
Slow logging? I don't understand.

KostyaSha:
@thomassuckow https://www.google.com/?#q=java+slow+logging&pws=0

thomassuckow:
Do just mean adding isLoggable guards? Or do you have other optimizations in mind?

KostyaSha:
@thomassuckow  https://github.com/jenkinsci/cucumber-testresult-plugin/pull/2#discussion_r19616785

thomassuckow:
Bah. I also just noticed my first commit is broken. It always things the image is not present.

That raises the question, should it try to pull every time? Then you would get updates... If there is nothing to pull it just moves on.

magnayn:
Would be nice to be optional - people have been asking for this kind of
feature..

thomassuckow:
Updated commits.
* Fixed pull always occurring if tag was not explicitly specified
* Change most logging messages to use formatter
* Fix bug noticed in reworking where if amiCap was 0, the containerCap was ignored

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -366,9 +373,6 @@ public boolean apply(@Nullable Container container) {
      *
      */
     private synchronized boolean addProvisionedSlave(String ami, int amiCap) throws Exception {
-        if( amiCap == 0 )
-            return true;
-
```
KostyaSha: :-1: it looks like a good performance optimization when capacity is zero.

thomassuckow: When amiCap is zero, this if statement causes the containerCap (server maximum) to not be checked. This means an infinite number of containers could be spawned even when the containerCap is set.

Additionaly this return bypasses the provisoningAmi's counting. which means the count of running slaves will probably be wrong (I believe, I don't fully understand the provisioningAmi's object).

If it returned false this would be a totally different story because then nothing would be provisioned.

KostyaSha: https://github.com/jenkinsci/docker-plugin/pull/142#discussion_r23509336

thomassuckow: In which case it should be `return false;`

This leads to another issue:
In the event an ami has too many instances, the plugin will not look for other images that also satisfy the conditions. When matching labels, it should really be getting a list of templates that are compatible.

thomassuckow: I will wait for @magnayn to weigh in on #142. If this change turns out to be holding up this pull request, I'll move it to another PR (or someone could pull only the first two commits).

KostyaSha: For templates i have simple patch that allows set the same label for multiple amis.
Also, you can't define two instances with the same ami - either fetch of the number running instances wouldn't work.

magnayn:
I've got a bit lost - this LGTM, should we merge?

thomassuckow:
I "believe" we are in line with the wording in #142. I think we should merge.

The disagreement was about when the limit is 0, is that infinite or none. I think that should be fixed in separate pull request (if it is an issue). This PR includes a fix preserving existing behaviour (0 = infinite).

KostyaSha:
Let's merge and later fix all 0 capacity behaviours

PR #146, opened 2015-01-27
Use public methods in DockerTemplate to allow reuse

carlossg:
I'm reusing the DockerTemplate for the [Kubernetes plugin](https://github.com/carlossg/jenkins-kubernetes-plugin) and it would be nice to have access to the methods that are now set as protected


MarkEWaite (approved):
@dependabot squash and merge

Issue #89, opened 2014-08-25
Enhancement request: Support CPU and memory constraints


hcguersoy:
It would be helpful if we could limit the slave containers on memory usage and CPU priority.
This could be done by supporting the docker run options *-m* and *-c* .

PR #152, opened 2015-01-31
Issue #89 Enhancement request: Support CPU and memory constraints

imasen:
As wished in [1] support to configure cpu and memory constraints.

[1] https://github.com/jenkinsci/docker-plugin/issues/89

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
@@ -34,7 +34,7 @@ public class DockerSlave extends AbstractCloudSlave {
     private transient Run theRun;
 
     @DataBoundConstructor
-    public DockerSlave(DockerTemplate dockerTemplate, String containerId, String name, String nodeDescription, String remoteFS, int numExecutors, Mode mode, int memoryLimit, int cpuShares, String labelString, ComputerLauncher launcher, RetentionStrategy retentionStrategy, List<? extends NodeProperty<?>> nodeProperties) throws Descriptor.FormException, IOException {
+    public DockerSlave(DockerTemplate dockerTemplate, String containerId, String name, String nodeDescription, String remoteFS, int numExecutors, Mode mode, Integer memoryLimit, Integer cpuShares, String labelString, ComputerLauncher launcher, RetentionStrategy retentionStrategy, List<? extends NodeProperty<?>> nodeProperties) throws Descriptor.FormException, IOException {
         super(name, nodeDescription, remoteFS, numExecutors, mode, labelString, launcher, retentionStrategy, nodeProperties);
         Preconditions.checkNotNull(dockerTemplate);
         Preconditions.checkNotNull(containerId);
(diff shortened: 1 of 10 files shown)


KostyaSha:
Can't this be done using lxcConfString or passing run args with existed UI entries?

imasen:
@KostyaSha It should work but only until Docker version 0.9, since 0.9 Docker is using libcontainer instead of lxc. You can still use lxc, but you need to specify it explicit.

> --lxc-conf=[]: (lxc exec-driver only) Add custom lxc options --lxc-conf="lxc.cgroup.cpuset.cpus = 0,1"

> If the Docker daemon was started using the lxc exec-driver (docker -d --exec-driver=lxc) then the operator can also specify LXC options using one or more --lxc-conf parameters. These can be new parameters or override existing parameters from the lxc-template.go. Note that in the future, a given host's docker daemon may not use LXC, so this is an implementation-specific configuration meant for operators already familiar with using LXC directly.

magnayn:
Ah, then let's fix that bug. I much prefer using guava, it's there 'for free' with jenkins and it's ugly to replace 2 well-described lines with 7. It's also worse, because guava correctly strips out nasties that users do, like tabs.

I suspect it's a version mismatch as that method is since 15.0. I'll check what version is in the pom, but my guess is it's 11.x, which would explain it.

imasen:
@magnayn ah okay, I unterstand.

I will wait for PR157 and the I will rollback my changes on this method.

imasen:
@magnayn I rebased all commits and rollback my changes on split-method. Please merge this PR.

src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
```diff
@@ -34,7 +34,7 @@
     private transient Run theRun;
 
     @DataBoundConstructor
-    public DockerSlave(DockerTemplate dockerTemplate, String containerId, String name, String nodeDescription, String remoteFS, int numExecutors, Mode mode, String labelString, ComputerLauncher launcher, RetentionStrategy retentionStrategy, List<? extends NodeProperty<?>> nodeProperties) throws Descriptor.FormException, IOException {
+    public DockerSlave(DockerTemplate dockerTemplate, String containerId, String name, String nodeDescription, String remoteFS, int numExecutors, Mode mode, int memoryLimit, int cpuShares, String labelString, ComputerLauncher launcher, RetentionStrategy retentionStrategy, List<? extends NodeProperty<?>> nodeProperties) throws Descriptor.FormException, IOException {
```
menski: You added `int memoryLimit, int cpuShares` as constructor arguments even though your aren't using them. Also they are `int` instead of `Integer` which will lead to NPE in `DockerTemplate` line 264 if the values are `null`.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -169,19 +181,27 @@ public CreateContainerCmd createContainerConfig(CreateContainerCmd containerConf
...
+            containerConfig.withMemoryLimit(memoryInByte);
+        }
         if( dnsHosts.length > 0 )
             containerConfig.withDns(dnsHosts);
         if( volumesFrom != null && !volumesFrom.isEmpty() )
             containerConfig.withVolumesFrom(volumesFrom);
-	if(environment != null && environment.length > 0)
-            containerConfig.withEnv(environment);
+        if(environment != null && environment.length > 0)
+                containerConfig.withEnv(environment);
```
thomassuckow: Erroneous whitespace

PR #153, opened 2015-02-02
Don't run FlyweightTask on Docker container

### LINKED ISSUE #148: Docker containers are killed even if there is Flyweight tasks on it
Sometimes, jobs are launched in same docker container (like if there is >1 executor on slave).

This lead to failure when 1 finish, it stops container with other job in it.

More information:
* example: http://i.imgur.com/LfarPRw.png
* error: https://gist.github.com/guilhem/6cb6d0049eb42dce7653
* context:
  * Jenkins ver. 1.580.2
  * plugin 0.8 && a2222d954e6e1cb770b2775340d52a71b9110e85
  * Master without any executor (all is done in docker)
  * Use [debian-package-builder](https://github.com/jenkinsci/debian-package-builder-plugin) and [matrix project](https://github.com/jenkinsci/matrix-project-plugin) (building package for ubuntu 12.04/14.04)
  * Docker image look like this one: https://registry.hub.docker.com/u/guilhem/jenkins-slave/

guilhem:
Fix #148

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
@@ -10,6 +10,7 @@ import com.nirima.jenkins.plugins.docker.action.DockerBuildAction;
 
 import hudson.Extension;
 import hudson.model.*;
+import hudson.model.queue.CauseOfBlockage;
 import hudson.slaves.AbstractCloudSlave;
 import hudson.slaves.ComputerLauncher;
 import hudson.slaves.NodeProperty;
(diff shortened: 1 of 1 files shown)


KostyaSha:
Interesting why they are at all allowed to run not on master.

jglick:
This fix seems dubious to me. From a casual look at #148 I suspect that the immediate issue is that `OneShotRetentionStrategy` should not kill a slave instantly after completion of a heavyweight task if there is also a `FlyweightTask` still running on it.

But it is not clear to me whether it is valuable/necessary to have such tasks run on the cloud slave to begin with. (Matrix parent jobs can legitimately run on a slave, but a Docker slave is a poor place to keep a workspace since it will be thrown out.) If the flyweight task should be blocked generally, then it would be better to do this as a `QueueTaskDispatcher` in `durable-task-plugin`.

Overall I do not have a solid picture of the design of flyweight tasks and what use cases are considered legitimate.

PR #156, opened 2015-02-05
Lock guava version to the version used in Jenkins

carlossg:
To avoid runtime NoSuchMethodError
If newer guava is needed then would need to use the pluginFirstClassLoader

Fix compilation error in older guava

This doesn't fix the bigger issue, com.nirima.docker-java:docker-java depends on that newer version of guava. If they use some new methods then the only option is to use `pluginFirstClassLoader`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,17 @@
         </dependency>
     </dependencies>
 
+    <!-- lock dependencies to versions included in Jenkins -->
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>11.0.1</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <build>
         <plugins>
             <plugin>
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -87,8 +87,11 @@ public abstract class DockerTemplateBase {
     }
 
     public String[] splitAndFilterEmpty(String s) {
-        List<String> list = Splitter.on(' ').omitEmptyStrings().splitToList(s);
-        return list.toArray(new String[0]);
+        List<String> temp = new ArrayList<String>();
+        for (String item : Splitter.on(' ').omitEmptyStrings().split(s)) {
+            temp.add(item);
+        }
+        return temp.toArray(new String[temp.size()]);
     }
 
     public String getDnsString() {


src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -87,8 +87,11 @@ protected Object readResolve() {
     }
 
     public String[] splitAndFilterEmpty(String s) {
-        List<String> list = Splitter.on(' ').omitEmptyStrings().splitToList(s);
-        return list.toArray(new String[0]);
+        List<String> temp = new ArrayList<String>();
```
KostyaSha: How this change relates to guava?

KostyaSha:
Have you tried to run on 1.565.3 with this changes?

carlossg:
no, but 1.565.3 also depends in guava 11.0.1

The build is failing with the missing docker-java snapshot

KostyaSha:
Yes, it will fail until @magnayn align his version and patches with upstream. 
If i right understand, you shouldn't pin guava versions because they will start conflicting with different jenkins versions, because there is no right classloading isolations.

magnayn:
Yes - there is a problem with Jenkins it seems that its version of guava 'leaks'.

The 2 options are either to go backwards to 11.x, which is what I'll probably do, or to include a shaded version of the dependency, which might be better as there's nice features in newer versions.

magnayn:
I wasn't aware of pluginFirstClassLoader - maybe that's an option.

carlossg:
If you don't pin the guava version in the pom you'll just get whatever the dependencies require, and you are just pushing a compilation error to a runtime error much later on, like it happened in this case.

The other option is to use the pluginFirstClassLoader where your dependency is loaded before the jenkins dependency. I wouldn't do it for this small method.

KostyaSha:
Loading core dependencies from plugins with not tested on core versions will lead to various bugs.

KostyaSha:
@carlossg what jenkins version do you use?

carlossg:
1.580.3

wdym with "Loading core dependencies from plugins with not tested on core versions will lead to various bugs"

magnayn:
It *ought* not to cause problems, but if we return something that's implementing an API that's specified in terms of guava11 and we return it from our own JAR, there's potentially class casting issues. I don't know if you'd get this in practice however.

carlossg:
this patch doesn't add nor remove guava from the dependency list
guava version 1.18 is introduced through docker-java, so this only locks it to a lower version, the same one used in jenkins to compile against the same version it will run against.
There are issues between the plugin and system classloaders (I've seen that in jenkins-kubernetes) but just when doing dynamic lookup of classes

carlossg:
another option would be to exclude guava in the docker-java dependency, that way the one pulled transitively from jenkins dependencies would be used, with the provided scope.

magnayn:
Ok, so where-we're-at :

I have PRs open for the last 2 parts of docker-java changes.

However, I suspect I'll still need to be on a fork of it, because
a) I had to work around a classloader issue in DockerClientBuilder
b) docker-java is currently using the Jersey2, shaded copy of guava
(repackaged to jersey.repackaged.com.google.common.. ). It seems to be a
terribly old version of guava. I think this is a bad idea.

Once (if) those PRs are merged, I will re-spin a fork which
- re-instates guava
- declares the dependency as [11.0,) -- I.E: any version >= 11.0 is fine.

This should mean then that the jenkins plugin is happy, as it should pick
up whatever version (11.0.1 IIRC) in jenkins-core.

Longer term, It'd be good to figure out the correct way of using higher
versions of guava in jenkins plugins, but this isn't just a problem for
this plugin. I don't know if plugin-first classloading will work (ought to,
but untested. I have seen oddities here before). Best alternative is to
probably utilise maven-shade-plugin on the hpi itself. I wonder what other
projects do (11.x is rather old now).

magnayn:
Actually, I knew other plugins must have had this issue.

jclouds is guava 18.0, and fixes it in the pom and by shading:

https://github.com/jenkinsci/jclouds-plugin/blob/master/pom.xml

carlossg:
> - declares the dependency as [11.0,) -- I.E: any version >= 11.0 is fine.

That's not a good idea, it will pick the latest version. If you want to ensure 11.0 compatibility you need to have a guava 11.0 dependency, so maven compiles and tests against it

pom.xml
```diff
@@ -46,6 +46,17 @@
...
+    <!-- lock dependencies to versions included in Jenkins -->
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>11.0.1</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
jglick: Bad idea. If you want to pick up the version from Jenkins, add an `<exclude>` to the dependency that would otherwise suck in a newer Guava.

jglick:
Shading is the safest approach.

Do not use range dependencies, period.

pom.xml
carlossg: Yeah, that's the other option I mentioned

kohsuke:
FWIW, +1 on sharding.

thomassuckow:
@kohsuke Does #157 look like what you mean by shading?

carlossg:
closing in favor of shading PR #157

PR #162, opened 2015-02-23
Add dockerhost to /etc/hosts

kalmanb:
Hey, this PR was intended to add dockerhost to /etc/hosts via the 'extraHosts' api. It adds the IP address of the host system that is running the container. We require the IP for system testing.

I needed to add a feature in https://github.com/docker-java/docker-java so branched from your 'docker-plugin' branch. The feature has been merged into docker-java 1.0.1-SNAPSHOT so does depend on an least that version.

I also merged all the changes from your master branch and sorted a bunch of issues around the way docker-java configures volumes and saving Jenkins config. We've tested this version on our system and it works well. We don't use all the features of your plugin so haven't tested them all. We do use volume binding to the host filesystem, that is working.

If this merge is too crazy, let me know and I'll port the the 'dockerhost' feature to your master branch :)


magnayn:
I can't grok what this change actually is, it'd need to be rebased onto this docker-plugin..

KostyaSha:
@kalmanb rebase PR please

kalmanb:
Probably best to kill this PR for now. 
There were two changes, the first was to pull master into java-docker, which you've already done. The second was to add the IP address of the host machine to /etc/hosts in the docker container. It's probably easier for everyone if I redo that work against master. It's only a small change.

PR #168, opened 2015-03-04
Add the possibility to choose the image repository and tag when commited

jcsirot:
This patch adds 2 fields to customize the repository and tag when the docker image is commited. When these fields are left empty the default values are used (i.e. the job name for the repository and the build name for the tag).

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
@@ -22,6 +22,8 @@ public class DockerJobProperty extends hudson.model.JobProperty<AbstractProject<
      * Tag on completion (commit).
      */
     public final boolean tagOnCompletion;
+    public final String imageRepository;
+    public final String commitTag;
     public final String additionalTag;
     public final boolean pushOnSuccess;
     public final boolean cleanImages;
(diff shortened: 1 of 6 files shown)


magnayn:
If you want to tag the image, can't you just use the included Post-build step  for "Build/Publish docker containers" and specify it there?

jcsirot:
No, because we do use any dockerfile. In my use case we are building images for development environment and the provisioning is done with ansible.

The build steps are:

1. run a job on a container using a base image
2. clone the ansible playbook from the code repo
3. run the ansible playbook to provision the container
4. commit the container with a given repo/tag (for instance `foo:4` for the version 4 of the project foo build environment)

magnayn:
Ah, right - I understand.

I suspect me putting tagging / pushing into the job property was probably a mistake, and it ought to have been a builder step. That way there wouldn't be the "tag" and "additionalTag" stuff, we could just specify as many / few as you like in post-build steps (so if someone else comes along wanting extra tags - such as foo:latest, that doesn't need a additionalAdditionalTag..)

jcsirot:
I see one benefit of the current implementation (i.e. using a job property): Since the commit is executed when the slave is shutdown, it is possible to commit the container which was used to run the build. If the action is implemented as a build step or a post-build action, the commit command has to be sent from the container itself and I'm not sure it is recommended or even possible.

KostyaSha:
AFAIK plugin can't commit itself, but jenkins may have access to check and process what is configured. Moving from JobProperty to Publisher is not easy because of backward compatibility and configuration migration. I think JP can be kept for now

pjdarton:
This looks abandoned; closing.

Issue #81, opened 2014-08-13
Configurable number of executors


javierlopezfernandez:
Our test jobs normally chain other jobs. For that reason we require more than one executor per slave. Looking at the code I see numExecutors is hardcoded to 1 and I wonder if this is temporal or permanent feature

maafy6:
I'm going to throw in my lot by saying this would be pretty useful, although I don't claim the same use case that the OP has, I just want to be able to run multiple builds on the same slave.

It's not entirely true that spinning up multiple containers is the same as (or even some epsilon less than delta than) multiple executors in the same container. There is a decent overhead in launching a container, having the jar copied over, doing the handshake with the server, yada yada yada. If you have the same issue as several others have reported of slave containers being stuck in suspended mode, you can have a fair amount of churn waiting for a container that works, and when that happens, you want to milk it for all it's worth.

Even if you do have mostly successful containers, I've found it pretty easy to DOS myself with a job that has about 20 sub-jobs, each of which were spinning up containers to do their build. Even with caps (instance and total), it was about a 50-50 shot as to whether or not I'd have to give up and restart Jenkins because it couldn't handle the churn.

I hacked the changes in the above commit together to test on our system here and it's working much, much smoother. I'm not 100% interested in opening a pull request right now because the version that that's based off of is fairly old, and I haven't been bothered to figure out what I need to do to build with the new docker-java jar the pom mentions because it can't seem to find it, but that's the base. It mostly merges straight into the current master (one files had some imports rearranged, but that was the only real conflict).

KostyaSha:
experimental feature in master, try follow references too many duplicated issues..

mikemol:
> Probably permanent, because I can't see the use of >1 executor per docker slave.

I realize that this may or may not be particularly relevant to this specific issue, but in Openshift, when namespace quotas are enabled, you can only get so much CPU or memory available for your pods and containers. And, for some setups, you _must_ set up quotas and CPU/memory commits for each pod, which means you can't rely on an idle container leaving cycles or resources for another container to use; a container effectively has 100% utilization of its quota at all times.

I realize this is a terrible setup, and systems probably shouldn't be set up that way, but it happens, and it's not always under our control. As such, in order to be able to do effective resource sharing, the only way to do it in such as case is _inside_ the container, and the only way to do that here is with multiple executors to an agent.

pjdarton:
[...]

PR #174, opened 2015-03-19
Configurable number of executors #81

maafy6:
Adds `numExecutors` to the DockerTemplate class; any slave created with this template will have that many executors. Switched the retention strategy from `OnceRetentionStrategy` to `CloudRetentionStrategy` so that, if multiple jobs are spawned on a slave, it doesn't go down when the first job finishes.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -37,6 +37,7 @@ import hudson.model.Node;
 import hudson.model.labels.LabelAtom;
 import hudson.plugins.sshslaves.SSHLauncher;
 import hudson.security.ACL;
+import hudson.slaves.CloudRetentionStrategy;
 import hudson.slaves.ComputerLauncher;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.RetentionStrategy;
@@ -50,6 +51,8 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
 
 
     public final String labelString;
+    
+    public final int numExecutors;
 
     // SSH settings
     /**
@@ -102,6 +105,7 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
 
     @DataBoundConstructor
     public DockerTemplate(String image, String labelString,
+                          int numExecutors,
                           String remoteFs,
                           String remoteFsMapping,
                           String credentialsId, String idleTerminationMinutes,
@@ -127,6 +131,7 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
 
 
         this.labelString = Util.fixNull(labelString);
+        this.numExecutors = numExecutors;
         this.credentialsId = credentialsId;
         this.idleTerminationMinutes = idleTerminationMinutes;
         this.sshLaunchTimeoutMinutes = sshLaunchTimeoutMinutes;
@@ -227,10 +232,14 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
 
         logger.println("Launching " + image );
 
-        int numExecutors = 1;
         Node.Mode mode = Node.Mode.NORMAL;
 
-        RetentionStrategy retentionStrategy = new OnceRetentionStrategy(idleTerminationMinutes());
+        RetentionStrategy retentionStrategy = null;
+        if (numExecutors == 1) {
+            retentionStrategy = new OnceRetentionStrategy(idleTerminationMinutes());
+        } else {
+            retentionStrategy = new CloudRetentionStrategy(idleTerminationMinutes());
+        }
 
         List<? extends NodeProperty<?>> nodeProperties = new ArrayList();
 
@@ -273,7 +282,7 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
     }
 
     public int getNumExecutors() {
-        return 1;
+        return numExecutors;
     }
 
     @Override
--- a/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
+++ b/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
@@ -9,6 +9,7 @@ public class DockerTemplateTest {
 
     String image = "image";
     String labelString;
+    int numExecutors = 1;
     String remoteFs = "remoteFs";
     String remoteFsMapping = "remoteFsMapping";
     String credentialsId = "credentialsId";
@@ -36,6 +37,7 @@ public class DockerTemplateTest {
 
     private DockerTemplate getDockerTemplateInstanceWithDNSHost(String dnsString) {
         DockerTemplate instance = new DockerTemplate(image, labelString,
+                numExecutors,
                 remoteFs,
                 remoteFsMapping,
                 credentialsId, idleTerminationMinutes,
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
@@ -38,6 +38,7 @@ public class DockerBuilderNewTemplate extends Builder implements Serializable {
 
     public final String image;
     public final String labelString;
+    public final int numExecutors;
     public final String remoteFsMapping;
     public final String remoteFs;
     public final String credentialsId;
@@ -63,7 +64,7 @@ public class DockerBuilderNewTemplate extends Builder implements Serializable {
     public final String hostname;
 
     @DataBoundConstructor
-    public DockerBuilderNewTemplate(String image, String labelString, String remoteFs, String remoteFsMapping,
+    public DockerBuilderNewTemplate(String image, String labelString, int numExecutors, String remoteFs, String remoteFsMapping,
                                               String credentialsId, String idleTerminationMinutes,
                                               String sshLaunchTimeoutMinutes,
                                               String jvmOptions, String javaPath,
@@ -82,6 +83,7 @@ public class DockerBuilderNewTemplate extends Builder implements Serializable {
 
         this.image = image;
         this.labelString = labelString;
+        this.numExecutors = numExecutors;
         this.remoteFs = remoteFs;
         this.remoteFsMapping = remoteFsMapping;
         this.credentialsId = credentialsId;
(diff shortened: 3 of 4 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -102,6 +105,7 @@
 
     @DataBoundConstructor
     public DockerTemplate(String image, String labelString,
+    		              Integer numExecutors,
```
KostyaSha: Why not integer? Please check indentation

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -272,8 +276,8 @@ public InspectContainerResponse provisionNew() throws DockerException {
         return provisionNew(dockerClient);
     }
 
-    public int getNumExecutors() {
-        return 1;
+    public Integer getNumExecutors() {
```
KostyaSha: I think null is not acceptable here, so better use primitives

KostyaSha:
I think time for `@DataBoundConstructor`! Let's check what core version do we use first...

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
maafy6: I actually had `int` to start, saw there were some other recent fields that used `Integer` so I switched it at the last moment. Argument about not null makes sense to me, I'll change and fix indent tonight.

thomassuckow:
I would prefer a configurable option for once vs cloud. The once strategy is very useful to me because I want a pristine environment for each build.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
KostyaSha: Please remove this from constructor and add DataBoundSetter with right default value

KostyaSha:
Yes, strategy must be choosable...

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -227,10 +232,9 @@ public DockerSlave provision(StreamTaskListener listener) throws IOException, De
 
         logger.println("Launching " + image );
 
-        int numExecutors = 1;
         Node.Mode mode = Node.Mode.NORMAL;
 
-        RetentionStrategy retentionStrategy = new OnceRetentionStrategy(idleTerminationMinutes());
+        RetentionStrategy retentionStrategy = new CloudRetentionStrategy(idleTerminationMinutes());
```
hawky-4s-: Please make the use of CloudRetentionStrategy optional, eg. only use it when multiple executors are configured.

maafy6:
Sounds good. I will put that in (disregard the current bit about pulling out durable-task, tested on my system and pushed before I read the new comments over here).

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
```diff
@@ -9,6 +9,7 @@
 
     String image = "image";
     String labelString;
+    Integer numExecutors = 1;
```
KostyaSha: again, use privimite

KostyaSha:
Picking all changes to my 0.9 branch. I tried to use existed code that allows choose strategy, but it lists: 
`hudson.slaves.RetentionStrategy.Always` `hudson.slaves.RetentionStrategy.Demand` and `SimpleScheduledRetentionStrategy`...

KostyaSha:
`Jenkins.getInstance().getDescriptorList(RetentionStrategy.class)`
```
Result

Result: [hudson.slaves.RetentionStrategy$Always$DescriptorImpl@16c18976, hudson.slaves.SimpleScheduledRetentionStrategy$DescriptorImpl@72057e17, hudson.slaves.RetentionStrategy$Demand$DescriptorImpl@48858368]
```

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -227,10 +232,14 @@ public DockerSlave provision(StreamTaskListener listener) throws IOException, De
...
-        int numExecutors = 1;
         Node.Mode mode = Node.Mode.NORMAL;
 
-        RetentionStrategy retentionStrategy = new OnceRetentionStrategy(idleTerminationMinutes());
+        RetentionStrategy retentionStrategy = null;
+        if (numExecutors == 1) {
+            retentionStrategy = new OnceRetentionStrategy(idleTerminationMinutes());
+        } else {
+            retentionStrategy = new CloudRetentionStrategy(idleTerminationMinutes());
+        }
```
KostyaSha: Is it make any sense? What will happen if OnceRetentionStrategy will be used with more executors?

KostyaSha: Sent this question to maillist hope to get some comments https://groups.google.com/d/msg/jenkinsci-dev/6nGESEvJ_S0/M-jLIrHJS6AJ

KostyaSha:
@maafy6 could you point on code line that enforces OnceRetentionStrategy to be used only for one executor? OnceRetentionStrategy has one significant implementation check vs Cloud, it allows to keep slave when it executes durable-task.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
maafy6: I'm not 100% sure if this answers your question, but hopefully this will make sense. I have not tested with OnceRetentionStrategy and multiple executors. From the [source](https://github.com/jenkinsci/durable-task-plugin/blob/master/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java) and docs, it looked like it very much expected to be running on a slave with one executor. As far as I could tell from the source, as soon as *any* task was completed, it would try and terminate the slave. What I didn't want to happen was the following scenario:

* DockerTemplate configured with two executors and OnceRetentionStrategy
* Jenkins queues two jobs, QuickJob and LongJob
* Jenkins spins up a new slave with that template and assigns QuickJob to one executor and LongJob to the other on that slave.
* QuickJob completes quickly and triggers a terminate long before LongJob completes.

I wasn't sure what would happen at this point, whether or not it would let LongJob finish up before tearing down or if it would just get killed. If it's the latter, then we certainly can't allow Once with multiple executors, if it's the former, then it'd be fine (but it doesn't seem significantly different than running as Cloud with a low idle termination time).

As for the case where we're running a ContinuableExecutable and a Cloud strategy, I don't have that in my system and am not familiar enough with the paradigm to speak to it, so it is untested. (Do such tasks assume/require that they be built on the same node?)

magnayn:
|'m not sure I follow why you'd *ever* want more than one executor per slave?

Docker is good because it's extremely cheap to spin up environments that give you isolation. 

I do forsee wanting to be able to start downstream builds in the same container as the previous step. What you actually want to be able to do in that scenario is to commit the container, and pass it down as the named image for subsequent step (or steps - this might be a fan-out). 

I submitted a PR to jenkins-core to try to get this functionality, but PRs die a horrible death there...

maafy6:
@magnayn See my comment on #81 as to why I think this is useful.

KostyaSha:
@magnayn what PR? I'm testing locally code that allows to choose strategies and executor numbers. I see no problem to provide it because they work in the way they implemented.

KostyaSha:
I run docker with ALWAYS strategy and it keeps this slave always online, i can set multiple executors and the main idea that i just switching from static environment creation to in-docker. Docker can be used in different ways and it should be good to provide ability choosing configurations.

magnayn:
So would it be fair to say that it isn't the case that you *need* multiple executors in order to fulfill a requirement of your build, it's purely an optimisation - likely that because of launching a slave causes the spinning up of an entire JVM to host slave.jar ?

I.E: if launching slave.jar were zero cost, this would not be a problem?

I can't remember if we pass through the CPU limiting options to docker; we probably should.

Multiple executors make me hesitant, because it feels like breaking the entire point of running inside a container. If you're not using it to throw away the environment at the end of the build - what are you even gaining by running in docker at all?

magnayn:
PR for slave parameters was https://github.com/jenkinsci/jenkins/pull/1448 from October 2014...

maafy6:
> So would it be fair to say that it isn't the case that you need multiple executors in order to fulfill a requirement of your build, it's purely an optimisation - likely that because of launching a slave causes the spinning up of an entire JVM to host slave.jar ?
> 
> I.E: if launching slave.jar were zero cost, this would not be a problem?

I could agree with that, but with qualifications. It's not so much that it takes too much time for the JVM to spin up and et cetera, when I kicked off a build that had a lot of sub-jobs, it hosed the network on the box (even with limits on the number of containers) and it was a 50-50 shot as to whether or not Jenkins would just crash and have to be restarted. If I did not see those issues, this would not be as big of a concern for me.

As things are though, I don't want to pretend my use case is the only one that matters. The stuff @KostyaSha is doing with adding more strategies sounds fantastic to me because it will open up more ways of using it.

> I can't remember if we pass through the CPU limiting options to docker; we probably should.

We can pass through CpuShares, but not CpuSet (restrict which particular CPUs).

>Multiple executors make me hesitant, because it feels like breaking the entire point of running inside a container. If you're not using it to throw away the environment at the end of the build - what are you even gaining by running in docker at all?

As to what I'm gaining, I have a ton of jobs that have very similar environments. 40 some jobs that use the same label and docker image. It doesn't matter to them that the other jobs are built on that machine, but they do need that environment, and that's what docker provides. In my case, I just need sufficient, not clean.

I am sympathetic to the concern in general. Not because I think it's a misuse of docker, but because things like your PR to Jenkins get tricky in the future. In my mind, an excellent use of that would be to be able to dynmically specify volumes for a job. This is obviously incompatible with multiple executors, but I think both are still worth having and finding a way for them to co-exist.

magnayn:
One thing that might be affecting you is exhaustion of the /dev/random entropy pool. I'm a bit surprised that the network gets hosed though. Might be worth digging.

Thanks for explaining - I wanted to make sure I understood your challenges! I think as long as we give enough help-text around the options (and perhaps hide under 'advanced' if it isn't already it should be fine). In fact, your usecase would make a great example for 'choose this option if...'.

You'd be surprised at the number of 'I twiddled this setting I don't understand and now everything's bust' emails I get! :-)

KostyaSha:
I lost my hard-drive, will need some time to create choosable strategies, but i will want to fix labels handling firstly.

KostyaSha:
@maafy6 feel free to try hpi from #209 Configuration hidden under advanced button, some help pages + form validations to provide better description for end-users.

KostyaSha:
closed with #209 Please test new features and provide feedback!

PR #179, opened 2015-03-20
move resource files back to correct folder

menski:
- it seems that this files where missed as the docker-plugin source
  was moved into a separate subfolder


KostyaSha:
Seems repository was trashed with 'shade' "fixes"....

KostyaSha:
Looks valid because repository was messed after merge from separate branch

magnayn:
Yes, it's that they've never been in the correct folder, and it of course didn't generate a merge conflict because of their previous non-existence.

PR #181, opened 2015-03-22
Fix image repository/tag when committing.

jgriffiths1993:
Sanitizes build name and build number when tagging image.

Fixes issue where additional tag doesn't work.


KostyaSha:
Another one duplicated PR

PR #183, opened 2015-03-22
fix volume support

tjamet:
Setting the 'Volumes' advanced option inside a cloud with a value of "/tmp/docker-jenkins:/tmp/docker-test-volume" does not bind container /tmp/docker-test-volume to host /tmp/docker-jenkins any longer.

It can be reproduced cloning "https://github.com/tjamet/jenkins-docker-master.git" branch " test/docker-volume" and launching a build on docker-volume after having updating the docker plugin

### DIFF
--- a/docker-plugin/pom.xml
+++ b/docker-plugin/pom.xml
@@ -42,6 +42,12 @@
             <artifactId>durable-task</artifactId>
             <version>0.5</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.9.5</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -6,6 +6,7 @@ import shaded.com.google.common.base.Objects;
 import shaded.com.google.common.base.Splitter;
 import shaded.com.google.common.base.Strings;
 import shaded.com.google.common.collect.Iterables;
+import com.github.dockerjava.api.model.*;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.DockerException;
 import com.github.dockerjava.api.command.CreateContainerCmd;
@@ -19,6 +20,7 @@ import com.github.dockerjava.api.model.VolumesFrom;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.annotation.Nullable;
@@ -28,7 +30,6 @@ import javax.annotation.Nullable;
  */
 public abstract class DockerTemplateBase {
     private static final Logger LOGGER = Logger.getLogger(DockerTemplateBase.class.getName());
-
 
     public final String image;
 
@@ -45,7 +46,7 @@ public abstract class DockerTemplateBase {
     public final String hostname;
 
     public final String[] dnsHosts;
-    public final String[] volumes;
+    public final String volumesString;
     public final String volumesFrom;
     public final String[] environment;
 
@@ -89,7 +90,7 @@ public abstract class DockerTemplateBase {
         this.cpuShares = cpuShares;
 
         this.dnsHosts = splitAndFilterEmpty(dnsString);
-        this.volumes = splitAndFilterEmpty(volumesString);
+        this.volumesString = volumesString;
         this.volumesFrom = volumesFrom;
 
         this.environment = splitAndFilterEmpty(environmentsString);
@@ -109,7 +110,7 @@ public abstract class DockerTemplateBase {
     }
 
     public String getVolumesString() {
-        return Joiner.on(" ").join(volumes);
+        return volumesString;
     }
 
     public String getVolumesFrom() {
@@ -201,8 +202,21 @@ public abstract class DockerTemplateBase {
         if (dnsHosts.length > 0)
             containerConfig.withDns(dnsHosts);
 
-        if (volumesFrom != null && !volumesFrom.isEmpty())
-            containerConfig.withVolumesFrom(VolumesFrom.parse(volumesFrom));
+        if(volumesString!=null) {
+            for (String volume : splitAndFilterEmpty(volumesString)) {
+                try {
+                    containerConfig.withBinds(Bind.parse(volume));
+                } catch (IllegalArgumentException e) {
+                    containerConfig.withVolumes(Volume.parse(volume));
+                }
+            }
+        }
+
+        if(volumesFrom!=null) {
+            for (String volumeFrom : splitAndFilterEmpty(volumesFrom)) {
+                containerConfig.withVolumesFrom(VolumesFrom.parse(volumeFrom));
+            }
+        }
 
         containerConfig.withTty(this.tty);
 
@@ -223,9 +237,16 @@ public abstract class DockerTemplateBase {
         if( dnsHosts.length > 0 )
             hostConfig.withDns(dnsHosts);
 
-        // ?
-        //if (volumes.length > 0)
-        //    hostConfig.with VolumesFrom (volumes);
+        if(volumesString!=null) {
+            for (String volume : splitAndFilterEmpty(volumesString)) {
+                try {
+                    hostConfig.withBinds(Bind.parse(volume));
+                } catch (IllegalArgumentException e) {
+                    LOGGER.log(Level.WARNING,"Cannot create volume "+volume+" when starting host");
+                    //hostConfig.withVolumes(Volume.parse(volume));
+                }
+            }
+        }
 
         List<LxcConf> lxcConfs = getLxcConf();
 
--- a/docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-volumesString.html
+++ b/docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-volumesString.html
@@ -1,3 +1,10 @@
 <div>
-    A space separated list of host volume mounts, e.g. /host/path:/container/path:ro
+    A space separated list of host volume mounts : &lt;host/path&gt;[&lt;container/path&gt;[&lt;:mode&gt;]]
+    <ul>
+        <li>container/path create empty  volume container/path</li>
+        <li>host/path:container/path will mount host/path from host to container's container/path, read/write</li>
+        <li>host/path:container/path:rw will mount host/path from host to container's container/path, read/write</li>
+        <li>host/path:container/path:ro will mount host/path from host to container's container/path, read-only</li>
+    </ul>
+    Note: when provided, an invalid mode will mount the volume in read-only
 </div>
--- a/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
+++ b/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
@@ -1,12 +1,25 @@
 package com.nirima.jenkins.plugins.docker;
 
+import com.github.dockerjava.api.command.CreateContainerCmd;
+import com.github.dockerjava.api.model.*;
 import com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy;
 import hudson.slaves.RetentionStrategy;
+import com.github.dockerjava.api.command.StartContainerCmd;
+
 import static org.junit.Assert.*;
+
+import org.junit.Before;
 import org.junit.Test;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+
+import org.mockito.Matchers;
+import org.mockito.Mockito;
 
 public class DockerTemplateTest {
 
+    CreateContainerCmd createContainerCmd;
+    StartContainerCmd startContainerCmd;
 
     String image = "image";
     String labelString;
@@ -22,6 +35,8 @@ public class DockerTemplateTest {
     String prefixStartSlaveCmd = "prefixStartSlaveCmd";
     String suffixStartSlaveCmd = " suffixStartSlaveCmd";
     String instanceCapStr = "";
+
+    String dnsString = "0.0.0.0";
 
     String dockerCommand = "dockerCommand";
     String volumesString = "volumes";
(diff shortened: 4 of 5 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -216,9 +214,13 @@ public StartContainerCmd createHostConfig(StartContainerCmd hostConfig) {
         if( dnsHosts.length > 0 )
             hostConfig.withDns(dnsHosts);
 
-        // ?
-        //if (volumes.length > 0)
-        //    hostConfig.with VolumesFrom (volumes);
+        for( String _volume : volumes){
+            String __volume[] = _volume.split(":",2);
```
KostyaSha: Please name variables friendly, it's not C

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -216,9 +214,13 @@ public StartContainerCmd createHostConfig(StartContainerCmd hostConfig) {
         if( dnsHosts.length > 0 )
             hostConfig.withDns(dnsHosts);
 
-        // ?
-        //if (volumes.length > 0)
-        //    hostConfig.with VolumesFrom (volumes);
+        for( String currentVolume : volumes){
```
KostyaSha: remote space before String

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -216,9 +214,13 @@ public StartContainerCmd createHostConfig(StartContainerCmd hostConfig) {
...
+            String bindAndVolume[] = currentVolume.split(":",2);
```
KostyaSha: put space after comma

KostyaSha:
So if there is no ":" symbol, it binds host directory to the same container directory?

tjamet:
Yes, it seems to me to be quite an intuitive implementation

KostyaSha:
@tjamet could you also check that help file describes right it?

tjamet:
having a look at the help, there is another option not handled right now, i'll handle it as well in the code

maafy6:
The behaviour when only one entry is specified (no `:` symbol) is entirely counter-intuitive to me. I would expect it to act like the `docker create --volumes=...` option. In that case, it would create a volume inside the container, but it wouldn't be mounted to the matching host directory. (It would just live somewhere in /var/lib.)

To be honest, I never tried that on the Jenkins docker image config, but that's what I've always assumed it would do. It saves a couple of keystrokes to not have to type out `/home/jenkins/.ssh:/home/jenkins/.ssh` I guess, but it doesn't seem particularly compelling to me, especially when it doesn't match docker and it would be the obvious way to implement creating a non-mounted volume.

KostyaSha:
I think we should read docker docks an implement the same.... Multiple entries should be supported with expandableTextbox and new line separator. We should just decide what should be in one line :)

maafy6:
After some digging around, it looks like we should probably move the logic to the `createContainerConfig` method. The `CreateContainerCmd` object it's working on has two methods that should allow everything we need:

`withVolumes` -- This is what we would need to handle the `docker create --volume=<container_only>` case.

tjamet:
according to this documentation https://docs.docker.com/reference/run/#volume-shared-filesystems docker implements volumes the way @maafy6 suggested it.
```
-v=[]: Create a bind mount with: [host-dir]:[container-dir]:[rw|ro].
       If "container-dir" is missing, then docker creates a new volume.
```

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
```diff
@@ -83,4 +116,77 @@ public void testDnsHosts() {
...
+
+        instance = getDockerTemplateInstanceWithVolumesString("");
+        assertEquals(0,instance.volumes.length);
+
+        instance = getDockerTemplateInstanceWithVolumesString("volume");
+        assertEquals(1,instance.volumes.length);
+        assertArrayEquals("volume".getBytes(),instance.volumes[0].getPath().getBytes());
+        assertArrayEquals("volume".getBytes(), instance.volumes[0].getVolume().toString().getBytes());
+
+        instance = getDockerTemplateInstanceWithVolumesString("volume0 volume1");
```
lanwen: can you split it to separated test methods? It a little bit hard to read :)

KostyaSha:
https://issues.jenkins-ci.org/browse/JENKINS-28143 seems related

docker-plugin/pom.xml
```diff
@@ -43,6 +43,12 @@
             <artifactId>httpclient</artifactId>
             <version>4.3.6</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.9.5</version>
```
KostyaSha: must be test scope

tjamet: i'll update it

KostyaSha:
Could you rebase?

KostyaSha:
Also will this code work correctly with old configurations?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -98,6 +97,43 @@ protected Object readResolve() {
         return this;
     }
 
+    protected Bind[] volumesFromString(String volumesString){
+        List<Bind> bindList = new ArrayList<Bind>();
+        for( String volume : splitAndFilterEmpty(volumesString)){
+            List<String> volumeSpecs = Splitter.on(":").omitEmptyStrings().splitToList(volume);
```
KostyaSha: I feel that you should better do it with regexp and optional group matching. 
And it will be very good to move such functionality to docker-java....

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -98,6 +97,43 @@ protected Object readResolve() {
         return this;
     }
 
+    protected Bind[] volumesFromString(String volumesString){
```
KostyaSha: I think better move single entry parsing to separate static method. Then it will be easier test parsing.

KostyaSha:
If you want i can rework this PR and open parallel PR for review.

tjamet:
I'm updating the PR

KostyaSha:
Test with `@LocalData` with old CloudTemplate format and check that it was loaded right also required.

RyanGannon:
When will we see a fix for this reaching a release?

KostyaSha:
@RyanGannon please don't spam, when we decide that it ready and works.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -201,8 +202,21 @@ public CreateContainerCmd createContainerConfig(CreateContainerCmd containerConf
...
             containerConfig.withDns(dnsHosts);
 
-        if (volumesFrom != null && !volumesFrom.isEmpty())
-            containerConfig.withVolumesFrom(VolumesFrom.parse(volumesFrom));
+        if(volumesString!=null) {
+            for (String volume : splitAndFilterEmpty(volumesString)) {
+                try {
+                    containerConfig.withBinds(Bind.parse(volume));
+                } catch (IllegalArgumentException e) {
+                    containerConfig.withVolumes(Volume.parse(volume));
```
KostyaSha: Why this trick needed?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -45,7 +46,7 @@
     public final String hostname;
 
     public final String[] dnsHosts;
-    public final String[] volumes;
+    public final String volumesString;
```
KostyaSha: Better store in real format and return String only in expected for UI getter. This must simplify many code.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -223,9 +237,16 @@ public StartContainerCmd createHostConfig(StartContainerCmd hostConfig) {
...
-        // ?
-        //if (volumes.length > 0)
-        //    hostConfig.with VolumesFrom (volumes);
+        if(volumesString!=null) {
+            for (String volume : splitAndFilterEmpty(volumesString)) {
+                try {
+                    hostConfig.withBinds(Bind.parse(volume));
+                } catch (IllegalArgumentException e) {
+                    LOGGER.log(Level.WARNING,"Cannot create volume "+volume+" when starting host");
+                    //hostConfig.withVolumes(Volume.parse(volume));
```
KostyaSha: Looks inconsistent with previous block.

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-volumesString.html
```diff
@@ -1,3 +1,10 @@
 <div>
-    A space separated list of host volume mounts, e.g. /host/path:/container/path:ro
+    A space separated list of host volume mounts : &lt;host/path&gt;[&lt;container/path&gt;[&lt;:mode&gt;]]
+    <ul>
+        <li>container/path create empty  volume container/path</li>
+        <li>host/path:container/path will mount host/path from host to container's container/path, read/write</li>
+        <li>host/path:container/path:rw will mount host/path from host to container's container/path, read/write</li>
+        <li>host/path:container/path:ro will mount host/path from host to container's container/path, read-only</li>
+    </ul>
+    Note: when provided, an invalid mode will mount the volume in read-only
```
KostyaSha: Better don't mount at all imho. Either it will be not obvious. Also all input data must be checked with FormValidation.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
tjamet: In fact there is no withVolumes for `StartContainerCmd` and almost all methods of it have been made deprecated:
'TODO: Almost all methods are deprecated as they have corresponding siblings in {@link CreateContainerCmd} now.'
from https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/command/StartContainerCmd.java
I've tried to be as consistent as possible from the existing code

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
tjamet: Bind 'mounts' a volume from the host, 'Volume' creates a volume
Both adds parsers but didn't find common parser to detect wether we were talking of a bind `hostPath:ContainerPath[:Mode]` or a volume `ContainerPath`

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-volumesString.html
tjamet: didn't find the form validation, where about is it?

KostyaSha: https://wiki.jenkins-ci.org/display/JENKINS/Form+Validation

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: Is it two separate features?

KostyaSha:
```
-v=[]: Create a bind mount with: [host-dir]:[container-dir]:[rw|ro].
       If "container-dir" is missing, then docker creates a new volume.
```
Isn't this handled somewhere in docker-java?

tjamet: https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/model/Bind.java#L51

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: but with Volumes vs Binds that probably wouldn't work...

KostyaSha:
No responce, will pick and fix comments myself

KostyaSha:
Breaks backward compatibility, reworking. Will pick help page and tests from this PR.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: After analysing library changes it seems that better store input data as `String[]`. This should save us from changes in docker-java.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: You can't create volume during start.
Create container support bind + volume creation calls.
Start only support additional binding. 
In this plugin case it should be enough to configure container during creation.

KostyaSha:
Also seems i found when volumes where broken https://github.com/jenkinsci/docker-plugin/commit/b1377e7bca80c6be49668e4d114469736d6b9262
As i see it supports only binding and not volume creation.

After investigation i have no strict opinion what configuration should be required: combined like in docker client or separated (that is more close to API structure).
Will PR my WIP changes for further discussions.

KostyaSha:
UPD: better use combined configuration, because it will keep backward compatibility with old user configurations that used "volume" variable for bind configuration.

KostyaSha:
Merged alternative fix to master, thanks for your PR. I think integration test will be more useful rather then mocking.

KostyaSha:
@lanwen if you want, you can pick @tjamet tests and propose PR (but see my comments).

lanwen:
@KostyaSha mocking is for unit tests and integration tests is for integration tests. Its different layers of tests. I'll send PR with tests soon

KostyaSha:
@lanwen when people has issues with integration and no with plain classes, then mocking is useless. I'm glad to have more tests and of course will accept them, but i will prefer concentrate on integration tests.

lanwen:
@KostyaSha did you know about test piramid? If lot of unit tests, no need lot of integration tests. When people have problems with integration, this means not enough testing of component which connects to another component. Integration tests should only check that this component correctly can send info to another.

http://www.martinfowler.com/bliki/TestPyramid.html

KostyaSha:
@lanwen i know what is jenkins and what is production usage. You may have 100% successful unit tests, but nobody will use this plugin if it wouldn't work in their jenkins.

KostyaSha:
@lanwen i will suggest firstly fix bugs and refactor some bad places and only then wrote tests. This will be good candidate for 1.0 version.

Issue #188, opened 2015-03-27
Exception while provisioning leaks container


maafy6:
From time to time, the docker plugin will attempt to provision a slave and, after creating and starting the container, will hit an exception trying to inspect the container. When this happens, no slave is created in Jenkins and the container is not cleaned up. A few of my images use an instance count of 1, and so this effectively blocks that image from being used until I manually clean up the containers.

KostyaSha:
Please provide logs, in what state container exist on docker host?
Was ssh slave connection attempt performed?

maafy6:
Container is up and running, exception was thrown at the end of `DockerTemplateBase.provisionNew()` inside `DockerSlave.provision()`, before it even tried the ssh connection. (Sorry, logs are a pain for an air-gapped network.)

I'm trying out a fix right now, if everything looks good after lunch I'll submit a PR with what I'm running. Short version of the plan for the PR is to have `DockerTemplateBase.provisionNew()` just return a string and do the inspect container later, then if that hits an exception, try and clean up the container.

KostyaSha:
`docker ps -a` after issue happen please

KostyaSha:
Moving discussion to #189

PR #189, opened 2015-03-27
Handle exception inspecting newly created container.

maafy6:
PR to address #188 

It's entirely possible that the container can be salvaged at this point; the root cause of this is usually a SocketTimeoutException. There's usually nothing wrong with the container itself at this point; it will be up and running and I can ssh into it and poke around, but there's no slave.jar because Jenkins already gave up on it.

This adds essentially one more layer of protection in terms of managing the containers. In testing, I've also seen the same SocketTimeoutException trigger on the `RemoveContainerCmd.exec()` call here, but after checking docker it seemed that it did at least manage to do the job and remove the container in that instance.


maafy6:
It's definitely more likely to happen when the host is overloaded, though I've also seen it when there's not much else going on. When I look at the logs, it's also not uncommon to see the ListImages/ListContainers time out in `DockerCloud.countCurrentDockerSlaves`, though that is relatively harmless since it will eventually try that again.

KostyaSha:
I will pick this change

Issue #190, opened 2015-03-31
DockerComputerLauncher always connects to docker host


srwalter:
DockerComputerLauncher gets the port for SSH from the binding as it should, but then it always assumes that the IP to connect to is the same as the docker host itself.  This is incorrect behavior in several scenarios, such as when the port is only bound to one interface, or the docker host is actually a Docker Swarm.

KostyaSha:
PR itself is Issue, closing this.

PR #191, opened 2015-03-31
DockerComputerLauncher: get host IP from binding

srwalter:
The IP address of the binding may not be the same as the IP of the
docker host.  The port may be bound to only one interface, or we could
be connecting through a Docker Swarm, in which case the container is
actually running on a different machine altogether.

Fixes issue #190

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerLauncher.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerLauncher.java
@@ -46,15 +46,19 @@ public class DockerComputerLauncher extends DelegatingComputerLauncher {
 
             ExposedPort sshPort = new ExposedPort(22);
             int port = 22;
+            String host = null;
 
             Ports.Binding[] bindings = detail.getNetworkSettings().getPorts().getBindings().get(sshPort);
 
             for(Ports.Binding b : bindings) {
                 port = b.getHostPort();
+                host = b.getHostIp();
             }
 
-            URL hostUrl = new URL(template.getParent().serverUrl);
-            String host = hostUrl.getHost();
+            if (host == null) {
+                URL hostUrl = new URL(template.getParent().serverUrl);
+                host = hostUrl.getHost();
+            }
 
             LOGGER.log(Level.INFO, "Creating slave SSH launcher for " + host + ":" + port);
             


thomassuckow:
LGTM

KostyaSha:
Why not get it always?

KostyaSha:
Btw, it need to be checked what should be used, hostname or ip.

srwalter:
We do get it always, in that we will always call getHostIp().  However, according to the docs for that function in docker-java, getHostIp() will return null for the "bind to all interfaces (0.0.0.0)" case.  In that case, we *should* use the IP of the docker host itself to connect.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerLauncher.java
```diff
@@ -46,15 +46,19 @@ private static SSHLauncher getSSHLauncher(InspectContainerResponse detail, Docke
...
             for(Ports.Binding b : bindings) {
                 port = b.getHostPort();
+                host = b.getHostIp();
             }
 
-            URL hostUrl = new URL(template.getParent().serverUrl);
-            String host = hostUrl.getHost();
+            if (host == null) {
+                URL hostUrl = new URL(template.getParent().serverUrl);
+                host = hostUrl.getHost();
```
KostyaSha: Why not just template.getParent().serverUrl.getHost() ?

srwalter: I just moved the existing code inside a conditional.  I can change it if you like

KostyaSha: Have no code under, hand afair some addresses already URL, so if it URL, then no need to create new objects. And usually URI is preferred instead URL.

KostyaSha: Checked, change looks right.

jgriffiths1993:
See comment on https://github.com/jenkinsci/docker-plugin/commit/152b35c2fe40eb07f77d9f5f0b8209ebb0776a9a 

`host` should not be `null` as docker will assign all interfaces to the binding (0.0.0.0). If this happens, the SSH launcher will attempt to connect to localhost.

Using `if(host == null || host.equals("0.0.0.0"))` will fix this direct issue.

However, in cases where the Cloud URL is proxying to a docker server which is binding ports to all interfaces, this will still not work.

srwalter:
I agree with also testing for host.equals("0.0.0.0").  The case of proxying to a docker server binding to 0.0.0.0 should be handled by the proxy, however.  Swarm, at least, is smart enough to report the IP of the real docker host, not 0.0.0.0.

hawky-4s-:
:+1:

KostyaSha:
PR to this branch additional commit with fix please

KostyaSha:
Please wait some time and pick new hpi for testing from PR builder.

ryrydundun:
Hi, When should this be available as downloadable to Jenkins?
 I don't see a new version yet.

PR #202, opened 2015-04-11
[JENKINS-27900] Image Name class(es) and custom repository name

jgriffiths1993:
* Custom repository name within optional block for jobProperty
* jobProperty using DataBoundSetter for this field, as opposed to filling up DataBoundConstructor
* DockerImageName class to provide an API for constructing, sanitizing and validation image names (and their respective components)
* DockerImageNameException for identifying specific exceptions for this
* Majority of logic and regular expressions for validation from Docker v1.5 source code
* If repository is not specified, uses job name and sanitizes it properly
* Sanitizes build number and additional tag
* Fairly exhaustive tests for DockerImageName - see these for API

Please review; I'm sure that there will be some changes left to do!

### DIFF
--- a/docker-java-shaded/pom.xml
+++ b/docker-java-shaded/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.nirima</groupId>
     <artifactId>docker-java-shaded</artifactId>
     <packaging>jar</packaging>
-    <version>0.10.2</version>
+    <version>0.10.3-SNAPSHOT</version>
     <name>Docker-java shaded jar for jenkins plugin</name>
 
 
--- a/docker-plugin/pom.xml
+++ b/docker-plugin/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>com.nirima</groupId>
         <artifactId>docker-plugin-parent</artifactId>
-        <version>0.10.2</version>
+        <version>0.10.3-SNAPSHOT</version>
     </parent>
 
     <groupId>com.nirima</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 
     <groupId>com.nirima</groupId>
     <artifactId>docker-plugin-parent</artifactId>
-    <version>0.10.2</version>
+    <version>0.10.3-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>Docker plugin parent pom</name>
 
@@ -18,7 +18,7 @@
         <connection>scm:git:ssh://github.com/jenkinsci/docker-plugin.git</connection>
         <developerConnection>scm:git:ssh://git@github.com/jenkinsci/docker-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/docker-plugin</url>
-        <tag>docker-plugin-0.10.2</tag>
+        <tag>HEAD</tag>
     </scm>
 
     <properties>


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
```diff
@@ -135,16 +137,23 @@ private void slaveShutdown(TaskListener listener) throws DockerException, IOExce
 
         DockerClient client = getClient();
 
-
+        // Get image name or use job/project name
+        String repositoryName = getJobProperty().imageRepository;
+        if (Strings.isNullOrEmpty(repositoryName)) {
```
lanwen: maybe use defaultIfEmpty from commons-lang?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+    /**
+     * De-constructs given image name into relevant objects
+     * @param imageName Full image name for docker image, or null
+     */
+    public DockerImageName(String imageName) throws DockerImageNameException {
+        tag = null;
+        repository = null;
+        namespace = null;
+        registry = null;
+        if (imageName == null || imageName.equals("")) {
```
lanwen: `Strings.isNullOrEmpty(tagToken)` ?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+    private RepositoryName repository;
+    private NamespaceName namespace;
+    private RegistryName registry;
+    
+    /**
+     * De-constructs given image name into relevant objects
+     * @param imageName Full image name for docker image, or null
+     */
+    public DockerImageName(String imageName) throws DockerImageNameException {
+        tag = null;
```
lanwen: why you reinit nulls with nulls?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+            // Must be a namespace
+            } else {
+                nameMap.put("namespace", nameParts[0]);
+                nameMap.put("registry", null);
+            }
+        // Full 3-part repository name
+        } else if (nameParts.length == 3) {
+            nameMap.put("registry", nameParts[0]);
+            nameMap.put("namespace", nameParts[1]);
+            nameMap.put("repository", nameParts[2]);
```
lanwen: maybe use conts to save name of each part? Or too many duplications instead

jgriffiths1993:
Thanks @lanwen - I've cleaned some of this up now

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
```diff
@@ -14,6 +15,12 @@
 public class DockerJobProperty extends JobProperty<AbstractProject<?, ?>> {
 
     /**
+     * Image repository name for committed containers (image names)
+     */
+    @DataBoundSetter
+    public String imageRepository;
```
KostyaSha: Never do fields public, DataBoundSetter is for setters

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+/**
+ * For constructing and validating docker image names, and their respective parts:
+ * [REGISTRY/][NAMESPACE/]REPOSITORY[:TAG]
+ * 
+ * Based on Docker v1.5 source
+ * @author Joshua Griffiths 08/04/2015
+ */
+
+
+public class DockerImageName {
```
KostyaSha: I think almost all code cab be replaced with one regexp. Do you know regexps?

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerJobProperty/config.jelly
```diff
@@ -2,21 +2,22 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
 
     <f:optionalBlock title="Docker Container" name="hasDockerContainer" inline="true" checked="${instance!=null}" >
-        <f:entry title="${%Commit on successful build}" field="tagOnCompletion">
-            <f:checkbox default="false"/>
-        </f:entry>
+        <f:optionalBlock title="${%Commit on successful build}" field="tagOnCompletion" inline="true">
```
KostyaSha: Optional block in optional block just for some number of values looks like a wrong design

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
jgriffiths1993: Sure, I can move it to DataBoundConstructor in that case

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerJobProperty/config.jelly
jgriffiths1993: I thought it best as the hidden values are only applicable to committed containers; I do see your point as the optional block has no adjacent options.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
KostyaSha: Damn, not constructor, create setter and put DataBoundSetter

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerJobProperty/config.jelly
KostyaSha: JobProperty was put into optionalBlock with inline=true, because jenkins core sucks and doesn't wrap it in existed UI (i have idea how fix it, but it need some time). In your case you will be forced to wrap all this options in if statements in java code - that usually sucks by design.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+        public RepositoryName(String namePart) {
+            super(namePart);
+            partRegexp = "^([a-z0-9_.-]*)$";
+            partInvertedRegexp = "[^a-z0-9_.-]";
+        }
+        
+        @Override
+        public void validate() {
+            if (!partName.matches(partRegexp)) {
+                throw new DockerImageNameException("Invalid registry: Registry must match [a-z0-9_.-]");
```
KostyaSha: Why not just return boolean?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+            }
+        }
+        
+        @Override
+        public boolean isValid() {
+            return partName.matches(partRegexp);
```
KostyaSha: And you have this method, looks really duplicated

KostyaSha:
For this change looks very complex for just tag name validation.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
jgriffiths1993: isValid() returns boolean, validate() throws exceptions with the reason for the name being invalid - e.g. for form validation when users enter the repository name

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
jgriffiths1993: I was going with something like:
"(?:(?:.*/(?:(?:[a-z0-9.]+[a-z0-9.-](?=[^-]))/)/)|(?:.+[:.].+)|(?:(?:[a-z0-9.]+[a-z0-9.-](?=[^-]))/))?[a-z0-9_.-]+(?::[a-zA-Z0-9][a-zA-Z0-9_.-]*)?"

It just couldn't be used to specifically clean the names up

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
KostyaSha: It looks absolutely redundant

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
```diff
@@ -0,0 +1,491 @@
...
+    
+    public static class RepositoryName extends AbstractDockerImageName {
+        
+        /**
+         * @param namePart Part of image name
+         */
+        public RepositoryName(String namePart) {
+            super(namePart);
+            partRegexp = "^([a-z0-9_.-]*)$";
+            partInvertedRegexp = "[^a-z0-9_.-]";
```
KostyaSha: Usually constants are put as static final in class fields

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
jgriffiths1993: Haha I know how much you hate constructor :wink:

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
KostyaSha: You matching by one regexp and printing as error other regexp.

jgriffiths1993:
It definitely is; however I'd scoped to use this class to add validation to forms and to be used in the step when containers are pushed (identifying the registry from the name, if not index.docker.io) which is where all the extra crap comes from

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
KostyaSha: Using public fields is bad solution, because you will unable to change it's behaviour later.
Constructor makes sense when there are not many parameters. Template for example has so many fields, so it better to use setters. Also constructor with critical fields + setter is more groovy friendly in case you will create some configuration programmatically.

lanwen: [Javadoc](http://stapler.kohsuke.org/apidocs/org/kohsuke/stapler/DataBoundSetter.html) says that it can be used on field 
> The field is discovered through simple reflection, so its name must match the property name, but its access modifier can be anything.

Will it work without setter in case of private field with annotation?

KostyaSha: You should follow docs, jelly checks for public field and then for getter/setter AFAIR. But again, using public fields is bad practise.

jgriffiths1993: It does indeed, I just think @KostyaSha is pointing it out as bad practice

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerSlave.java
```diff
@@ -152,8 +161,9 @@ private void slaveShutdown(TaskListener listener) throws DockerException, IOExce
             String tagToken = getAdditionalTag(listener);
 
             if( !Strings.isNullOrEmpty(tagToken) ) {
-                // ?? client.image(tag_image).tag(tagToken, false);
-                client.tagImageCmd(tag_image,null,tagToken).exec();
+                client.tagImageCmd(imageId, imageName.toString(), tagToken)
```
KostyaSha: If you created a real wrapper object, then using .toString() is bad idea, wrapper class should provide some clear getter that wouldn't depend on toString() implementation that usually has format "Key = value" for printing all object internals.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
KostyaSha: You can define multiple patterns with group matches, example https://github.com/jenkinsci/github-plugin/blob/master/src/main/java/com/cloudbees/jenkins/GitHubRepositoryName.java#L26-L48

KostyaSha: And if this regexp is really right, then it's ok to use, but document/comment somehow. Not sure that it right java regexp, on first look it perl/PCRE.

jgriffiths1993:
Would a helper/utility class be more attractive than a full naming class?

KostyaSha:
IMO just one class DockerImageName that will have parsed data maybe enough. GitHubRepositoryName looks for me as good example. Do you have any arguments for having divided/sub-classed storage? Any pluses for having them so OOP'ed?

jgriffiths1993:
Definitely; my only argument for the excessive OOP was for situations where you may just want to validate a tag, or a registry

KostyaSha:
1) Can it be validated with generic parsing? Not specified part of entered value can be just empty or null.
2) If tag needs really to be reused in all docker places, then it may sense to implement them as Describable with UI and use it as f.property, f.optionalProperty, f.repeatableList, if not then it can just exist in DockerImageName.

Can you summarise/review all places in plugin where used every part of DockerImageName format to understand what is really need to be shared/reused?

jgriffiths1993:
Sorry, had to run some Saturday errands.
I'll have a look into this and update it with the changes.

KostyaSha:
While searching the best way i committed small fix for tag to master and released 0.9.0-rc1 because we should go forward and align users with current code. Better tagging will go to some next release.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/DockerImageName.java
KostyaSha: btw, is it the same as NameParser from docker-java do?

KostyaSha:
@jgriffiths1993 hi, could you renew this topic and provide list of things that you want enhance?

jgriffiths1993:
I'll do this separately once I've taken a look at how things are going now

KostyaSha:
@jgriffiths1993 could you just wrote what do you want achieve?

Issue #204, opened 2015-04-15
Can't add Docker as cloud provider in Jenkins 1.609


parshimers:
Installing the plugin (release or beta, both do the same) on Jenkins 1.609 works, however when one tries to add Docker as a cloud provider in "Manage Jenkins", nothing appears to happen. In the log this is thrown:

```
Apr 15, 2015 3:50:58 PM hudson.ExpressionFactory2$JexlExpression evaluate
WARNING: Caught exception evaluating: descriptor.getPropertyType(instance,field).itemTypeDescriptorOrDie in /$stapler/bound/ed39b8b6-53ff-4b89-87f7-711ddb07a59f/render. Reason: java.lang.reflect.InvocationTargetException
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor409.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.apache.commons.jexl.util.PropertyExecutor.execute(PropertyExecutor.java:125)
	at org.apache.commons.jexl.util.introspection.UberspectImpl$VelGetterImpl.invoke(UberspectImpl.java:314)
	at org.apache.commons.jexl.parser.ASTArrayAccess.evaluateExpr(ASTArrayAccess.java:185)
	at org.apache.commons.jexl.parser.ASTIdentifier.execute(ASTIdentifier.java:75)
	at org.apache.commons.jexl.parser.ASTReference.execute(ASTReference.java:83)
	at org.apache.commons.jexl.parser.ASTReference.value(ASTReference.java:57)
	at org.apache.commons.jexl.parser.ASTReferenceExpression.value(ASTReferenceExpression.java:51)
	at org.apache.commons.jexl.ExpressionImpl.evaluate(ExpressionImpl.java:80)
	at hudson.ExpressionFactory2$JexlExpression.evaluate(ExpressionFactory2.java:74)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$3.run(CoreTagLibrary.java:134)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.WhenTag.doTag(WhenTag.java:46)
[...]

PR #205, opened 2015-04-15
Upstream master

KostyaSha:
For #204


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java
```diff
@@ -17,6 +19,7 @@
      * Tag on completion (commit).
      */
     public final boolean tagOnCompletion;
+    @CheckForNull
```
lanwen: Just for interest - what this annotation does?

oleg-nenashev: This annotation is a part of JSR-305 standard, which is being supported by almost all static analysis tools and IDEs.

PR #207, opened 2015-04-19
Small code cleanups

KostyaSha:
Wow, my strategy and labels change was pushed to GH before FS dead. It has many coding style changes, staging hunks/lines to minimise them, but keeping the most critical cleanups... Splitting to single commits them


KostyaSha:
Shouldn't break anything, merging.

PR #208, opened 2015-04-20
fix #139, #164 temp tag fix

KostyaSha:
i also have strategy changes and some jelly fixes, but they will be with portion of choosable strategy.

PR #209, opened 2015-04-21
Strategy full

KostyaSha:
Current version with all commits that i'm working on. #208 contains only tested portion


KostyaSha:
Still testing but will be glad to get review/comments @magnayn @thomassuckow .
I will move strategy to DataBoundSetter, default should work right in most cases, so no need in having it in big constructor.
Jelly changes between two changes and it will be extremely difficult to reorder commits for me. heteroProperty now allows reordering templates that can influence for template picking for the same label that exists in multiple templates.

KostyaSha:
It is the last big work that i have and then i will move to PR/issues fixing.

magnayn:
Looks reasonable to me..

KostyaSha:
Tested manually, even 2 executors with cloud strategy works fine and picks the second job from queue. 
Of course this features experimental. Will be interested in feedback later :). Merging!

maafy6:
Will try and give this a shot next week. On vacation through the weekend.

PR #224, opened 2015-05-22
Hide null commit hash in action view

lanwen:
Hide empty hash on not commited container

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/224)


KostyaSha:
IMHO "docker" word redundant form image

lanwen:
>Usage guidelines
>Do not modify or alter the Docker Marks; ... Use our official logo as seen above.

>Do not:
>Modify the logo in any other way than scaling it
>Use the logo without the letters 'docker'

KostyaSha:
LOL, more docker stupideness...

lanwen:
IMO Logo with letters a little bit prettier than without. Think it not difficult to do in way docker restrict

KostyaSha:
According to their rules word "docker" also can't be used. Plugin has "docker" letters near logo everywhere. This absolutely redundant to have letters in two places. It prettier only because it not scaled. And btw scaling was fixed in some commits before, what plugin version do you use?

KostyaSha:
Also please separate commits

lanwen:
Screens with not scaled was from 0.8 (but i've seen them on 0.9 too). Ok, agree with redundant with two places "docker" word, so only commit hash hide in this pr

PR #225, opened 2015-05-22
Add tests for mapped ws browser

lanwen:
Add tests for simple component. This is a part of lot unit tests, started from simple components. 
Also, simplify component code.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/225)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowser.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowser.java
@@ -1,23 +1,27 @@
 package com.nirima.jenkins.plugins.docker.ws;
 
-import shaded.com.google.common.base.Strings;
 import com.nirima.jenkins.plugins.docker.action.DockerBuildAction;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Job;
 import hudson.model.Run;
 import hudson.model.WorkspaceBrowser;
+
 import java.io.File;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import static org.apache.commons.io.FileUtils.getFile;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 
 @Extension
-public class MappedFsWorkspaceBrowser extends WorkspaceBrowser  {
+public class MappedFsWorkspaceBrowser extends WorkspaceBrowser {
 
     private static final Logger LOGGER = Logger.getLogger(MappedFsWorkspaceBrowser.class.getName());
 
     public MappedFsWorkspaceBrowser() {
-        LOGGER.info(this.getClass().getName() + " initializing.");
+        LOGGER.log(Level.INFO, "{0} initializing...", this.getClass().getName());
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowser.java
```diff
@@ -25,17 +29,13 @@ public FilePath getWorkspace(Job job) {
         Run lastBuild = job.getLastBuild();
 
         if (lastBuild != null) {
-            DockerBuildAction a = lastBuild.getAction(DockerBuildAction.class);
+            DockerBuildAction action = lastBuild.getAction(DockerBuildAction.class);
 
-            if (a != null && !Strings.isNullOrEmpty(a.remoteFsMapping)) {
-                File mappedRemoteWorkspace = new File(a.remoteFsMapping);
-                mappedRemoteWorkspace = new File(mappedRemoteWorkspace, "workspace");
-                mappedRemoteWorkspace = new File(mappedRemoteWorkspace, job.getName());
```
KostyaSha: What was this 2 lines about?

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowserTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.hamcrest.Matchers.both;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.endsWith;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.startsWith;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.when;
+
+/**
+ * User: lanwen
```
KostyaSha: Please put right javadoc `@author lanwen` or better your first and last names. Date and time imho redundant.

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowserTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Mock
+    private Run run;
+
+    @Test
+    public void shouldReturnWsMappingOnFieldInAction() throws Exception {
+        MappedFsWorkspaceBrowser mappedFsWorkspaceBrowser = new MappedFsWorkspaceBrowser();
+
+        when(job.getLastBuild()).thenReturn(run);
+        when(job.getName()).thenReturn(JOB_NAME);
+        when(run.getAction(DockerBuildAction.class))
```
KostyaSha: Probably MappedFsWorkspaceBrowser class can be just refactored to have static method that will extract path from action, then it will be easier to test it without mocking.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowser.java
lanwen: Construction of full path - `fsMapping/workspace/jobname` with help of `File(File parent, String child)`. Same as inside of `getFile` method in commons-io

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/ws/MappedFsWorkspaceBrowserTest.java
lanwen: don't think it should be done - because of it needs to check how it gets the action from jenkins (it needs mocks) and because of content of such static method will be equal the content of inner `if` (its overhead). Mocks here is simple and easy to use

KostyaSha:
LGTM

PR #229, opened 2015-05-23
Add macAddress configuration for template

### LINKED ISSUE #226: Add the ability to set the MAC address when starting a container
We have a docker image that requires a static MAC address due to a node lock license requirement. I have tried to use the LXC Conf Options to set the MAC address with no luck. 

I changed the daemon to use the lxc driver (--exec-driver=lxc) and set the LXC option to
lxc.network.hwaddr=02:d0:c6:0a:29:33

It seems to just be ignored, so we went back to using the libcontainer driver (default)

So, to try to fix this,  I upgraded my fork to use version 1.3 of the docker-java dependency which introduced a withMacAddress method to the dockerjava/api/command/CreateContainerCmd.java class.

When I modify the DockerTemplateBase class of the plugin and set containerConfig.withMacAddress("02:d0:c6:0a:29:33"), the container starts with that MAC Address. 

So, here are my questions:

    1) Is there a way to set the container before starting to a static MAC address with the current plugin
         version? 
    2) If the answer to 1 is no, are there any plans to upgrade the plugin to use version 1.3 of docker-java?
    3) If the answer to 2 is yes, would adding a MAC Address field to the plugin config panel make sense?

Basically, we really need a way to set a static MAC address when starting a container from the plugin configuration panel that works.

KostyaSha:
Closes #226

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -1,6 +1,7 @@
 package com.nirima.jenkins.plugins.docker;
 
 import com.github.dockerjava.api.model.*;
+import org.apache.commons.lang.StringUtils;
 import shaded.com.google.common.base.Function;
 import shaded.com.google.common.base.Joiner;
 import shaded.com.google.common.base.Objects;
@@ -20,6 +21,8 @@ import java.util.logging.Logger;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
+
+import static org.apache.commons.lang.StringUtils.*;
 
 /**
  * Base for docker templates - does not include Jenkins items like labels.
@@ -71,6 +74,8 @@ public abstract class DockerTemplateBase {
     public final boolean privileged;
     public final boolean tty;
 
+    @CheckForNull private String macAddress;
+
     public DockerTemplateBase(String image,
                               String dnsString,
                               String dockerCommand,
@@ -84,7 +89,8 @@ public abstract class DockerTemplateBase {
                               String bindPorts,
                               boolean bindAllPorts,
                               boolean privileged,
-                              boolean tty
+                              boolean tty,
+                              String macAddress
 
     ) {
         this.image = image;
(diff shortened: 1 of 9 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -279,6 +295,10 @@ public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig
         if (environment != null && environment.length > 0)
             containerConfig.withEnv(environment);
 
+        if (getMacAddress() != null && !getMacAddress().trim().isEmpty()) {
```
lanwen: `StringUtils.isNotBlank()`

KostyaSha: I don't like guava functions internals. Jenkins has Utils.* And all input variables handling need to be reviewed later. So this not critical atm.

lanwen: its apache commons-lang, not guava. Jenkins's Utils.* mostly copy-paste from there

KostyaSha: changed

dcheese:
Looks good. The only suggestion would be to change the name to "MAC Address" within the config panel. Other than this,  +1 to merge

KostyaSha:
@dcheese changed, thanks

dcheese:
So, when will the next release be available that includes this (docker-plugin-0.9.0-rc2)?

KostyaSha:
@dcheese i think it will be 0.9.0 because volumes was the main blocker. There are many different not blocking issues so i think it makes sense to do periodic releases RORE! 
So will do new release soon.

dcheese:
great, thanks!

dcheese:
soon, like a week or a month?

KostyaSha:
@dcheese today, just doing release and fixing for other plugin.

KostyaSha:
```
[INFO] Uploaded: http://maven.jenkins-ci.org:8081/content/repositories/releases/com/nirima/docker-plugin/0.9.0/docker-plugin-0.9.0.hpi (11573 KB at 1079.6 KB/sec)
```

dcheese:
just one last thing, I noticed that there was a default MAC address set instead of it defaulting an empty string/blank. Is that what you intended? If it not set, then it should generate a MAC address which I think should be the default behavior. Thoughts?

PR #231, opened 2015-05-24
move all logic for getting docker client to separate class

lanwen:
Cloud class is too heavy - code for setup docker client should be placed in other class
Also it can help to test integration in future

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/231)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -69,6 +69,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.HashMap;
 
+import static com.nirima.jenkins.plugins.docker.client.ClientBuilderForPlugin.dockerClient;
+
 /**
  * Docker Cloud configuration. Contains connection configuration,
  * {@link DockerTemplate} contains configuration for running docker image.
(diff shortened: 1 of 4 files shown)


KostyaSha:
Sorry, but I wouldn't merge your refactorings because they doesn't fix any issues. I'm busy with fixing real bugs and feature requests and your changes will only add load to my work.

lanwen:
(Error will be fixed)

Why you think it adds load to your work? This PR adds tests and simplifies one of the hardest to understanding class. In future it can help to avoid some bugs and issues. Also it do nothing itself, just move methods from one class to another.

lanwen:
@KostyaSha This plugin has zero tests for such complexity. Its hard to write tests when it have such big classes with all logic in hundred of lines on one method. More tests - more stability and less issues. Writing tests is my business. If we ever just fix issues and implement new features without tests and without any refactoring, this plugin will be unmaintainable soon

lanwen:
@KostyaSha Can such refactorings be merged after 1.0 release?

KostyaSha:
When most bugs/long waiting features will be done - of course yes.

lanwen:
@magnayn thanks!
I'll update PR after #235 and stable version

KostyaSha:
I added to least, actions can be changed but i firstly want unbundle launcher as it heavy work that i already started.

KostyaSha:
Reviewed again, shouldn't conflict with my wip.

PR #227, opened 2015-05-23
WIP fix Volumes

KostyaSha:
superseed #183

PR #232, opened 2015-05-25
Test for empty volumesFrom in config.xml

lanwen:
https://github.com/jenkinsci/docker-plugin/pull/227#issuecomment-105224233

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/232)


KostyaSha:
Configuration data handling will be also changed. It inconsistent and bad in many places.

PR #233, opened 2015-05-25
for Volumes

KostyaSha:
Filters dirty inputs more carefully

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -1,6 +1,7 @@
 package com.nirima.jenkins.plugins.docker;
 
 import com.github.dockerjava.api.model.*;
+import org.apache.commons.lang.StringUtils;
 import shaded.com.google.common.base.Function;
 import shaded.com.google.common.base.Joiner;
 import shaded.com.google.common.base.Objects;
@@ -110,16 +111,32 @@ public abstract class DockerTemplateBase {
 
     protected Object readResolve() {
         if (volumesFrom != null) {
-            setVolumesFrom2(new String[]{volumesFrom});
+            if (StringUtils.isNotBlank(volumesFrom)) {
+                setVolumesFrom2(new String[]{volumesFrom});
+            }
             volumesFrom = null;
         }
 
         return this;
     }
 
+    //TODO move/replace
     public static String[] splitAndFilterEmpty(String s, String separator) {
         List<String> list = Splitter.on(separator).omitEmptyStrings().splitToList(s);
-        return list.toArray(new String[0]);
+        return list.toArray(new String[list.size()]);
+    }
+
+    //TODO move/replace
+    public static String[] filterStringArray(String[] arr) {
+        final ArrayList<String> strings = new ArrayList<>();
+        if (arr != null) {
+            for (String s : arr) {
+                if (StringUtils.isNotEmpty(s)) {
+                    strings.add(s.trim());
+                }
+            }
+        }
+        return strings.toArray(new String[strings.size()]);
     }
 
     public String getDnsString() {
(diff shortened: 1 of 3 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -110,16 +111,32 @@ public DockerTemplateBase(String image,
...
+        return list.toArray(new String[list.size()]);
+    }
+
+    //TODO move/replace
+    public static String[] filterStringArray(String[] arr) {
+        final ArrayList<String> strings = new ArrayList<>();
+        if (arr != null) {
+            for (String s : arr) {
+                if (StringUtils.isNotEmpty(s)) {
+                    strings.add(s.trim());
```
lanwen: here can be npe

KostyaSha:
@lanwen are you ok with this changes?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -110,16 +111,32 @@ public DockerTemplateBase(String image,
 
     protected Object readResolve() {
         if (volumesFrom != null) {
```
lanwen: just check once
```
if (StringUtils.isNotBlank(volumesFrom)) {
                setVolumesFrom2(new String[]{volumesFrom});
}
```

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -110,16 +111,32 @@ public DockerTemplateBase(String image,
...
+    //TODO move/replace
     public static String[] splitAndFilterEmpty(String s, String separator) {
         List<String> list = Splitter.on(separator).omitEmptyStrings().splitToList(s);
-        return list.toArray(new String[0]);
+        return list.toArray(new String[list.size()]);
+    }
+
+    //TODO move/replace
+    public static String[] filterStringArray(String[] arr) {
+        final ArrayList<String> strings = new ArrayList<>();
```
lanwen: don't use concrete class names if it has interface - in this case it should be List

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: You are wrong, this code lines not about beauty, it data migration.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: It filters array of strings.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
lanwen: Are you kidding me?

What for you need two ifs if second if also checks first?
And why you asking me ok, or not, if you don't listen me?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: As you see i mentioned that it need to be removed or replaced in future. I have no time for beatifying code that will be removed soon.

lanwen:
ok to me

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
KostyaSha: Your proposed replacement absolutely wrong from data migration view point. Even if current code duplicated, then it not an issue. And as i repeated 100 times personally to you "configuration handling will be changed" no sense in wasting time on code beauty. Instead of continuing unbundling launcher i spent the whole day on unrelated to fixes comments.

lanwen: So you can just write `will be fixed next time` and merge without any doubt. And next time don't write duplicated code.

PR #234, opened 2015-05-25
Launchers

KostyaSha:
Just to start some discussion. And exclude potential conflicts. 
I'm working on adding other ways on connecting slave to master. JNLP is one of ideas and should allow connecting slaves back through firewalls while only connection to docker API will be required. Other variants of launching maybe experimented then. (For local connection like @jglick suggested it even can  be possible to do connection through plain stdin/stdout without ssh/security overhead).
Also with this changes slave/template configuration becomes close to default dumbslaves, that are user-friendly.

So current plan:
- unbundle existed launcher to separate class. Sshconnector is very useful because it doesn't have Host address like sshlauncher. 
- move DockerTemplateBase related methods, fields, help files to DockerTemplateBase. Corresponding config.jelly should help cleanup-ing and keeping code up-to-date in right place.
- unhardcode 22 port for DockerSSH ComputerLauncher
- DockerComputerLauncher become a marker class and it's implementations should handle host names and docker related part during connection/launching
- After this changes DockerTemplate can be simplified and contain in itself DockerTemplateBase without extending classes. For UI this will be simply `<f.property>`
- Add data migration and tests for it

@magnayn will you have any objections for this work? 

/cc @thomassuckow

Current code is WIP and you can skip hard commenting.
PS for container create/start i found inconsistency in docker-java and docker docs, so this will be clean-uped later.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/234)


KostyaSha:
Seems cloud api sucks. 
- for having container id as Slave name container must be run before Slave object creation
- if i want connect jnlp with secret, then container with this cmd args must be created after Computer object in launch method, but then slave.toComputer call fails that fails provisioner.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerSSHLauncher.java
```diff
@@ -0,0 +1,185 @@
...
+            Ports.Binding[] bindings = detail.getNetworkSettings().getPorts().getBindings().get(sshPort);
+
+            for (Ports.Binding b : bindings) {
+                port = b.getHostPort();
+                host = b.getHostIp();
+            }
+
+            //get address, if docker on localhost, then use local?
+            if (host == null || host.equals("0.0.0.0")) {
+                URL hostUrl = new URL(template.getParent().serverUrl);
```
lanwen: `URI.create(..).getHost()` can help avoid catching of MailformedUrlExc

KostyaSha:
@stephenc could you suggest something? Is it possible to achieve what is wrote in last comment here?

stephenc:
@KostyaSha 

> if i want connect jnlp with secret, then container with this cmd args must be created after Computer object in launch method, but then slave.toComputer call fails that fails provisioner.

OK, so the trick there is to use a different `AgentProtocol` that does redirect mapping based on your pre-computed secret for the container. That would let you generate the secret up front.... though somewhat of a scary thing to implement correctly as you need to do a lot of copy and paste and you need to keep in sync with the core version

stephenc:
Let me clarify, so implementing `AgentProtocol` on its own is not too bad... provided you are implementing a new protocol name... what you'd actually want to do is *take over* the `JNLP2-connect` protocol, intercept if the `Secret-Key` is one of yours and pass through to the other implementations if not yours... if `Secret-Key` is one of yours then you need to pass through the redirect...

Here is where stuff gets tricky...

To inspect the `Secret-Key` you need to read the input stream... so you need to wrap the input stream in a self-unreferencing buffer that captures the initial read and lets you rewrite the initial part!

because that input stream is a DataInputStream who's first "object" is a UTF-8 string that is a `Properties` of the request...

You will need to read those `Properties` and if one of yours you will rewrite the `Node-Name` and `Secret-Key` values and provide those as the alternative first object... otherwise you just reset the stream to the beginning

KostyaSha:
@stephenc what is your opinion on what step is better create and run container? I just want remove hardcodes and provide extensibility because docker can be used in different ways and having everything hardcoded is not very good.

stephenc:
So all the stuff needed to get the container up and running and ready to connect should take place in the `Callable<Node>` that is submitted to `Computer.threadPoolForRemoting` in the call to `Cloud.provision(...)` with the resulting `Future<Node>` being returned in the `PlannedNode`

Do not forget that the `Callable<Node>` should **not** add the `Node` to Jenkins, instead that needs to be left up to `NodeProvisioner`... and with the current Cloud API you need to persist somehow or other your activities so that you know what to tidy up or continue with in the event of a Jenkins restart.

As such, if you are doing this "right" (or at least as right as can be done in the current API) the Callable will start up the slave and wait for it to be ready. At that point it will know the container id (so can create the Node with that node name) and it can even run the command for the JNLP connector against the Node as it's last hurrah!

This business I have seen of people spin-waiting in the connector/launcher while waiting for the Node to be ready is one of the more wrong ways of implementing the cloud API

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -254,7 +249,8 @@ public Node call() throws Exception {
                                 // TODO: record the output somewhere
                                 DockerSlave slave = null;
                                 try {
-                                    slave = t.provision(new StreamTaskListener(System.out));
+//                                    slave = t.provision(new StreamTaskListener(System.out));
+                                    slave = provision(t);
```
stephenc: Look at the evilness below here:

```
jenkins.addNode(slave);
```

NOOOOOOOOOOOOOOOOOO! Leave that up to NodeProvisioner as per [javadoc](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/slaves/Cloud.java#L145) and the baseline version of Jenkins is well well at least 1.503

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -269,10 +265,11 @@ public Node call() throws Exception {
                                     //
                                     // deferring the completion of provisioning until the launch
```
stephenc: This is the wrong way to wait. What you want to do is check the docker container until it is **ready** to connect and then **and only then** return. This way sadly needs to do the evil adding to Jenkins so that `toComputer()` does not return `null`...

wrong wrong and more wrong!

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
KostyaSha: This part of code is not mine and of course it weird for me, my idea was to implement all objects and pass to Cloud API .

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
KostyaSha: This also not my code, but without it i had spawned many containers. 
What will be the right way of provisioning/calling some methods for this block of code?
As i see Computer object created after slave added and created by update() calls after adding slave.

KostyaSha:
@stephenc well note, this commits are different ways where i tried to move container creation/launcher to get ability to have jnlp/ssh launchers...

stephenc:
Just so that people understand my previous comments in context...

I rate this implementation of the Cloud API as maybe the 3rd/4th best implementation of the Cloud API... coming after

1. operations-center-cloud (which has problems that are conveniently masked by the provisioning logic in operations-center-server)... sadly this is a closed source implementation so I cannot show you it
2. nectar-vmware (which only wins over mansion because of [this evilness](https://github.com/jenkinsci/mansion-cloud-plugin/blob/master/src/main/java/com/cloudbees/jenkins/plugins/mtslavescloud/PlannedMansionSlave.java#L247) ) 
3. [mansion](https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Cloud+Connector+Plugin) / nectar-vmware Note that mansion has problems that are conveniently masked by the provisioning logic in our backing cloud and if it didn't add the node to jenkins in the Callable<Node> it would win out over the closed source nectar-vmware

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
stephenc: @KostyaSha the back story is that I have been telling @kohsuke that the Cloud API is "a pile of shite" and he asks for "why" and "where is the evidence". Well this is one of the pieces of evidence: that implementations end up doing this...

The Javadocs clearly state that implementations should not add the node to Jenkins and it is the responsibility of NodeProvisioner to do the adding. Even [@kohsuke has done this](https://github.com/jenkinsci/mansion-cloud-plugin/blob/master/src/main/java/com/cloudbees/jenkins/plugins/mtslavescloud/PlannedMansionSlave.java#L247) in that which I previously considered to be the second best cloud api implementation...

A good cloud API would give you the abstractions such that you can get the sequencing of all these events correct and return the node instance at the correct point in time.

KostyaSha:
Thanks for comments, will review mentioned plugins and try rework this.

stephenc:
@KostyaSha sadly there is no decent OSS cloud implementation I can point you to...

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
thomassuckow: I am not super familiar with the cloud api, but @stephenc please tell me if I am understanding.

NodeProvisioner.PlannedNode takes a Future. Currently, we provision a slave (including doing some bad things) and return a node immediately despite it not being connected yet. Instead what we should be doing is creating a slave and get connected to the docker instance and finally resolve the future.

I'm guessing if that is what were are supposed to be doing there is some gotcha that prevents us from creating a slave that way...

stephenc: That's the general overview of it.

The gotcha - if there is one - is that in some ways the easiest way to test if the node is ready is to try to connect with a launcher... which needs the node to be added to Jenkins so you can get the other half of the Node/Computer duality created.

The correct way is to use a non-launcher based way to determine that the node is ready to connect with

KostyaSha: Sorry, but when launcher will be used then? And it seems that in case of jnlp this may look crazy, though i can have some timeout and do connection retries from docker side. 
So backing to my initial idea, i can move current launcher configuration to separate launcher class (unbundle hardcodes), then i can have some (static) methods that i can call from Provisioner for determining that connection is ok, like i did method for configuring dockerCreate parameters.  And then Cloud API will use "standard" launcher methods?

KostyaSha:
Ah, i see 1) and 2) are closed source. 3) is difference only in a way of waiting, so if i right understand i should remove slave adding and implement some other way of waiting slave (for ssh i can just use ssh connection but without using Slave/Computer?).

steve-jansen:
@KostyaSha 

Can you use a 2nd set of hands on this?  I was going to add support for JNLP launchers (and possibly jenkins-swarm) when I stumbled on this.

#243 is a regular problem for me, and figure ssh slaves for docker is overkill when we can just use JNLP as PID 1...

KostyaSha:
I hope i will end it soon, just doing another one round and testing slow running instances.

KostyaSha:
```
Jun 09, 2015 10:59:16 PM INFO com.nirima.jenkins.plugins.docker.DockerCloud provision
Asked to provision 1 slave(s) for: null
Jun 09, 2015 10:59:16 PM INFO com.nirima.jenkins.plugins.docker.DockerCloud provision
Will provision 'evarga/jenkins-slave', for label: 'null'
Jun 09, 2015 10:59:16 PM FINE com.nirima.jenkins.plugins.docker.client.ClientBuilderForPlugin
Building connection to docker host "docker" at: http://192.168.1.4:2375
Jun 09, 2015 10:59:17 PM INFO com.nirima.jenkins.plugins.docker.DockerCloud addProvisionedSlave
Provisioning 'evarga/jenkins-slave' number '1' on 'docker'; Total containers: '1'
Jun 09, 2015 10:59:17 PM INFO com.nirima.jenkins.plugins.docker.DockerCloud provisionWithWait
Trying to run container for evarga/jenkins-slave
Jun 09, 2015 10:59:17 PM FINEST com.nirima.jenkins.plugins.docker.utils.PortUtils sleepFor
Sleeping for 10 SECONDS
Jun 09, 2015 10:59:27 PM FINEST com.nirima.jenkins.plugins.docker.utils.PortUtils sleepFor
Sleeping for 10 SECONDS
Jun 09, 2015 10:59:37 PM INFO com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
SSH port is open on 192.168.1.4:49160
Jun 09, 2015 10:59:37 PM INFO com.nirima.jenkins.plugins.docker.DockerComputerSSHLauncher getSSHLauncher
Creating slave SSH launcher for 192.168.1.4:49160
Jun 09, 2015 10:59:52 PM FINE com.nirima.jenkins.plugins.docker.DockerComputer
 Computer DockerComputer{name=a29032dc8657@docker, slave=DockerSlave{name=a29032dc8657@docker, containerId=a29032dc86578538f51046c9fe7aaeabcd7dd8eaf63d93c7c86e21c3e68066bb, template=DockerTemplate{image=evarga/jenkins-slave, parent=DockerCloud{name=docker, serverUrl=http://192.168.1.4:2375}}}} taskAccepted
```
with          `   createCmd.withCmd("bash", "-c", "sleep 20 && /usr/sbin/sshd -D");` works like a charm.

magnayn:
Heh - much of the initial commits were cribbed from the jenkins-jclouds and
jenkins-EC2 cloud implementations, which I (perhaps naively) assumed were
the 'right way' to do things.

Nothing to do with this commit, but the things I notice (certainly on my
install), which we have struggled with (and may point to deficiencies in
the cloud API) are

- inadvertant deadlocking inside Jenkins (surprising need to synchronize()
in the cloud provider - which may no longer be true
- I get lots of SSH connection that never seem to complete. Various people
have tinkered with retrying, timeouts, waits and suchlike. I've got one
machine that never seems to actually get used (containers spin up and spin
down, but they never get used)
- Ones like the above get stuck in a "(offline) (suspended)" state, that
I'm not sure what it actually means from a 'user' perspective

KostyaSha:
One more thing, cloud retention strategy that has 0 timeout will kill launching node if it launchers > 1 minute because clock timing is 1 minute. But it ok because connecting >1 minute to ready slave is bad.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerJNLPLauncher.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public ComputerLauncher getPreparedLauncher(DockerTemplate template, InspectContainerResponse containerInspectResponse) {
+        return jnlpLauncher;
+    }
+
+    @Override
+    void appendContainerConfig(DockerTemplateBase dockerTemplate, CreateContainerCmd createContainerCmd) {
+        // jnlp command for connection with secret?
+        createContainerCmd.withCmd("wget http://10.6.60.60:8090/jenkins/jnlpJars/slave.jar",
```
stephenc: Should the IP address be hard-coded? (I don't mind if this is a sensible default and the address is hidden in an Advanced box)

Or is this an implicit configuration that all slave images can safely assume the IP address mapping?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -48,43 +48,53 @@
...
      */
-    public final String sshLaunchTimeoutMinutes;
+    @Deprecated
+    private String sshLaunchTimeoutMinutes;
 
     /**
      * Field jvmOptions.
      */
-    public final String jvmOptions;
+    @Deprecated
```
stephenc: probably want to make all these `@Deprecated` fields as `transient` so that they do not get saved out on resave

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -225,12 +235,17 @@ public int getSSHLaunchTimeoutMinutes() {
...
         }
     }
+
     /**
      * Initializes data structure that we don't persist.
      */
     protected Object readResolve() {
         super.readResolve();
 
+//        if (launcher != null) {
```
stephenc: Should migrate the `@Deprecated` fields if non-null

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
KostyaSha: This is WIP

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerJNLPLauncher.java
KostyaSha: this is WIP i will call you when something will be ready for review

stephenc:
From what I can see this is headed in a better direction. I think there is some refactoring in the launchers that would be nice to have, but I'd rather see this set of changes - once finished - adopted first before I try to add additional decoration and improvements on top.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerJNLPLauncher.java
stephenc: Well I was just providing some comments on what you've done so far. This was by no means a "final" review, rather just some things I spotted... I rather note down things I spot when I spot them than occupy my head trying to remember them for when the review actually is ready...

Studies show that most people's working memory is about 6-8 "slots" and when you fill them up, something falls out... keeping 2 of these slots occupied for my comments on this WIP is too expensive a cost to pay. You are free to ignore my comments if you want

KostyaSha:
Changes are very dirty of course, i need get working variant before doing clean-ups and backward compatibility. Atm i can't send jnlp connection command with docker-java. Manually 'docker exec' works fine. And i think no need in trying to run jnlp as PID1, because we can't know jnlp mac and jnlpUrl before starting container.

KostyaSha:
@magnayn what was the idea of overriding `isAcceptingTasks()` ?

stephenc:
isAcceptingTasks() is used to flag that there should be no more work assigned to the node. The super method uses a "boolean" flag to control the result, but that flag is accessible to all of Jenkins and thus a rogue plugin could force it back to true. By overriding you can ensure that the policy of the cloud (i.e. one build per provisioning) is respected irrespective of what rogue plugins decide to do...

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
KostyaSha: I know what to do with backward compatibility.

stephenc:
Of course that is why *one might want to override it*. 

In this case it seems to be a bad plan.

That method should not block **ever** so having it sleep for 5 seconds will kill provisioning from the Queue.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
stephenc: I know you do, but we all can benefit with a reminder comment in a pull request... I've been known to put such reminders in myself **on my own pull requests** so that I don't forget. This is a big change... a good change, but still big. There is lots that anyone could forget, reminders are a good thing... but if you're going to get all tetchy maybe I will stop reviewing this change altogether.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerComputerJNLPLauncher.java
KostyaSha: I just want save your time. You can ignore generic things like backward compatibility. ATM i'm playing with better design (in general docker based methods for launcher: wait, contribute), then i will need review what is broken because hardcoded launcher in template was used in other (non cloud functionality) parts of plugin.

KostyaSha:
Docker API is really crazy :(. 
<lazy-mode>
@stephenc 1) shouldn't such activities belong to retention strategies instead of Computer? Imho retention should decide when accept/delete/etc..
2) I think retention strategy lives in it's own life and doesn't aligned with cloud provisioning... I have successfully launched container but strategy decide terminate it, but build works and ends fine...
</lazy-mode>
Of course i will investigate this problems, but maybe you already know.

stephenc:
I have said the whole cloud API is a pile of shite where no-one can actually completely implement it correctly, but anyways I'll see if I can take a peek tomorrow in between fixing some deadlock issue and cutting our RC off 1.609.1 (IOW I may be busy and forget to look at this tomorrow... Friday worst case... It's important that there is a better OSS implementation of the crappy cloud API so we can point people at that rather than have even more clones of the really bad impls)

KostyaSha:
@stephenc ok, don't worry will try to review retention mechanism again, maybe i lost some boolean flags and it "conflicts".

KostyaSha:
Oh, lol i forgot that i'm on 1.565.3. It can be also 1.565.3 issues :D
```
INFO: Started provisioning Image of evarga/jenkins-slave from docker with 1 executors. Remaining excess workload:0.0
Jun 11, 2015 12:14:19 AM com.nirima.jenkins.plugins.docker.DockerCloud provisionWithWait
INFO: Trying to run container for evarga/jenkins-slave
Jun 11, 2015 12:14:29 AM com.nirima.jenkins.plugins.docker.DockerComputerJNLPLauncher launch
INFO: Executing: [bash, -c, cd /home/jenkins/ && wget http://192.168.1.3:8090/jenkins/jnlpJars/slave.jar -O slave.jar && java -jar slave.jar -jnlpUrl http://192.168.1.3:8090/jenkins/computer/3419d1431e5b@docker/slave-agent.jnlp]
Jun 11, 2015 12:14:29 AM hudson.slaves.NodeProvisioner update
INFO: Image of evarga/jenkins-slave provisioningE successfully completed. We have now 3 computer(s)
Jun 11, 2015 12:14:30 AM hudson.TcpSlaveAgentListener$ConnectionHandler run
INFO: Accepted connection #2 from /192.168.1.4:35250
Jun 11, 2015 12:14:31 AM hudson.slaves.CloudRetentionStrategy check
INFO: Disconnecting 3419d1431e5b@docker
Jun 11, 2015 12:14:31 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
INFO: Disconnected computer
Jun 11, 2015 12:15:35 AM hudson.model.Run execute
INFO: test #51 main build action completed: SUCCESS
Jun 11, 2015 12:15:35 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
INFO: Disconnected computer
```
Btw i have idea how to get PID 1, so i hope soon start clean-uping this "implementation"

steve-jansen:
Crazy idea - would support for launching a slave that uses containerized Jenkins swarm slaves simplify any of these headaches?

My shop would be indifferent to launching containerized slaves on demand via SSH or JNLP or swarm - what ever is the most reliable and scalable.

The UDP broadcast and security part of swarm doesn't bother me since our Jenkins farm is running on a dedicated VLAN.

stephenc:
swarm is another pile of crazyness... don't get me started on that!

On 11 June 2015 at 00:12, Steve Jansen <notifications@github.com> wrote:

> Crazy idea - would support for launching a slave that uses containerized
> Jenkins swarm slaves simplify any of these headaches?
>
> My shop would be indifferent to launching containerized slaves on demand
> via SSH or JNLP or swarm - what ever is the most reliable and scalable.
>
> The UDP broadcast and security part of swarm doesn't bother me since our
> Jenkins farm is running on a dedicated VLAN.
>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/docker-plugin/pull/234#issuecomment-110943748>
> .
>



-- 
Stephen Connolly ∙ Software Developer & Architect ∙ CloudBees, Inc.
<http://www.cloudbees.com/>

KostyaSha:
Review status: 0 of 55 files reviewed, 3 unresolved discussions, some commit checks failed.

---

<sup>**[docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java, line 165 \[r4\]](https://reviewable.io:443/reviews/jenkinsci/docker-plugin/234#-Jr_-qccSCE4v7Wcjf5d-r4-245)** ([raw file](https://github.com/jenkinsci/docker-plugin/blob/7f117d34eabd5f3ff7387b258b3fa52f581fd5c5/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java#L165)):</sup>
No problems. Added reviewable.io

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/jenkinsci/docker-plugin/234)

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
KostyaSha: No problems. Added reviewable.io

KostyaSha:
atm it looks:
screenshot 2015-06-12 03 38 10

steve-jansen:
:+1: on the screenshot

KostyaSha:
@stephenc i think you can review general design of how Cloud API is be used now: 
- I think it makes sense to move launchers into new package
- Termination need to be tested (parent object is not saved now and it should help avoid NPEs, but on other hand cloud remove/rename may end with lost container. Some analog of Computer/Slave will be needed for Cloud/Cloud. Or allow remove Cloud only when associated computers/slaves will be deleted).
- launcher internals is still WIP (testing jnlp and it's configuration)
- everything in `readResolve` is try-wrapped either configuration page will loose configured Cloud.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -108,51 +78,48 @@ public DockerCloud(String name,
...
         this.version = version;
         this.credentialsId = credentialsId;
         this.serverUrl = serverUrl;
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
-        if (templates != null)
-            this.templates = new ArrayList<DockerTemplate>(templates);
-        else
+
+        if (templates != null) {
```
KostyaSha: TODO migrate to setter

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerLauncher.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Return valid configured launcher that will be used for launching slave
+     */
+    public abstract ComputerLauncher getPreparedLauncher(String cloudId, DockerTemplate dockerTemplate, InspectContainerResponse ir);
+
+    /**
+     * Contribute container parameters needed for launcher.
+     * i.e. port for exposing, command to run, etc.
+     */
+    public abstract void appendContainerConfig(DockerTemplateBase dockerTemplate, CreateContainerCmd createContainerCmd);
```
KostyaSha: Probably better require everywhere `DockerCloud, DockerTemplate, InspectContainerResponse/CreateContainerCmd` ?

KostyaSha: Not a problem, can be changed later. I don't expect that somebody will implement this outside this plugin.

KostyaSha:
@steve-jansen if you want help, then testing is welcome.

steve-jansen:
@KostyaSha awesome, will build latest from this branch and take it for a test drive.

KostyaSha:
@steve-jansen Thanks, also please check "manage old data page" i expect that something can be forgot. My laptop is half broken again so i may periodically be silent.

KostyaSha:
@steve-jansen you can pick hpi https://jenkins.ci.cloudbees.com/job/plugins/job/docker-plugin/343/com.nirima$docker-plugin/

steve-jansen:
Hi @KostyaSha !

I ran a few builds on my local machine using docker-compose and a job designed to smoke test our dockerized Jenkins slaves.  Initial results look promising.  I don't see the NPE issue anymore, and the first attempt at creating a slave was successful across 10 builds; previously, a build would typically need to cycle through 3-5 slave creation cycles to land a working slave.

A few notes:

1.  I assume the `SEVERE: Failed to connect to 172.17.42.1:32774` log lines are just noise in the log, waiting for the SSH port to be ready.

2.  The stack trace for `WARNING: Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5 termination trace` is also a bit confusing to me.  Is this expected behavior?

3. I did see a slave creation failure due to `java.io.IOException: Unexpected termination of the channel`.  I seem to recall this exception as related to a longstanding issue with SSH slaves in Jenkins, so not sure it has anything to do with this plugin.

See below for logs.

Cheers,
Steve

Example log of successful creations:

```
jenkins_1 | Jun 23, 2015 11:45:38 AM com.nirima.jenkins.plugins.docker.DockerCloud provision
jenkins_1 | INFO: Asked to provision 1 slave(s) for: docker
jenkins_1 | Jun 23, 2015 11:45:38 AM com.nirima.jenkins.plugins.docker.DockerCloud provision
jenkins_1 | INFO: Will provision 'myregistry/jenkins-slave:1.609.1', for label: 'docker'
jenkins_1 | Jun 23, 2015 11:45:38 AM com.nirima.jenkins.plugins.docker.DockerCloud addProvisionedSlave
jenkins_1 | INFO: Provisioning 'myregistry/jenkins-slave:1.609.1' number '0' on 'docker-slave'; Total containers: '1'
jenkins_1 | Jun 23, 2015 11:45:38 AM hudson.slaves.NodeProvisioner$StandardStrategyImpl apply
jenkins_1 | INFO: Started provisioning Image of myregistry/jenkins-slave:1.609.1 from docker-slave with 1 executors. Remaining excess workload: 0
jenkins_1 | Jun 23, 2015 11:45:38 AM com.nirima.jenkins.plugins.docker.DockerCloud provisionWithWait
jenkins_1 | INFO: Trying to run container for myregistry/jenkins-slave:1.609.1
jenkins_1 | Jun 23, 2015 11:45:38 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | SEVERE: Failed to connect to 172.17.42.1:32774 (try 1/10) - There was a problem while connecting to 172.17.42.1:32774
jenkins_1 | Jun 23, 2015 11:45:48 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | INFO: SSH port is open on 172.17.42.1:32774
jenkins_1 | Jun 23, 2015 11:45:48 AM com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher getSSHLauncher
jenkins_1 | INFO: Creating slave SSH launcher for 172.17.42.1:32774
jenkins_1 | Jun 23, 2015 11:45:58 AM hudson.slaves.NodeProvisioner$2 run
jenkins_1 | INFO: Image of myregistry/jenkins-slave:1.609.1 provisioning successfully completed. We have now 2 computer(s)
jenkins_1 | [06/23/15 11:46:01] SSH Launch of e17a7bd029d1@docker-slave on 172.17.42.1 completed in 3,705 ms
jenkins_1 | Jun 23, 2015 11:46:20 AM hudson.model.Run execute
jenkins_1 | INFO: docker-smoke-test #5 main build action completed: SUCCESS
jenkins_1 | Jun 23, 2015 11:46:20 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Disconnected computer
jenkins_1 | Jun 23, 2015 11:46:21 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Stopped container e17a7bd029d1649c42f5e1470317ce4cadbf8807b6d557ba3bdb8c2e141d27de
jenkins_1 | Jun 23, 2015 11:46:21 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Shutdowned slave for e17a7bd029d1649c42f5e1470317ce4cadbf8807b6d557ba3bdb8c2e141d27de
jenkins_1 | Jun 23, 2015 11:46:21 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Removed container e17a7bd029d1649c42f5e1470317ce4cadbf8807b6d557ba3bdb8c2e141d27de
jenkins_1 | Jun 23, 2015 11:46:21 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:46:20 UTC 2015 by Thread[Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5,5,main] [id=425]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:63)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:46:21 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:46:20 UTC 2015 by Thread[Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5,5,main] [id=425]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.model.Computer.disconnect(Computer.java:465)
jenkins_1 | 	at hudson.slaves.SlaveComputer.disconnect(SlaveComputer.java:601)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerSlave._terminate(DockerSlave.java:147)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:67)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:46:21 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:46:21 UTC 2015 by Thread[Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5,5,main] [id=425]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at jenkins.model.Nodes$3.run(Nodes.java:165)
jenkins_1 | 	at hudson.model.Queue._withLock(Queue.java:1207)
jenkins_1 | 	at hudson.model.Queue.withLock(Queue.java:1143)
jenkins_1 | 	at jenkins.model.Nodes.removeNode(Nodes.java:160)
jenkins_1 | 	at jenkins.model.Jenkins.removeNode(Jenkins.java:1700)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:70)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:46:21 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:46:21 UTC 2015 by Thread[Executor #0 for e17a7bd029d1@docker-slave : executing docker-smoke-test #5,5,main] [id=425]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.model.Computer.disconnect(Computer.java:465)
jenkins_1 | 	at hudson.slaves.SlaveComputer.disconnect(SlaveComputer.java:601)
jenkins_1 | 	at jenkins.model.Nodes$3.run(Nodes.java:166)
jenkins_1 | 	at hudson.model.Queue._withLock(Queue.java:1207)
jenkins_1 | 	at hudson.model.Queue.withLock(Queue.java:1143)
jenkins_1 | 	at jenkins.model.Nodes.removeNode(Nodes.java:160)
jenkins_1 | 	at jenkins.model.Jenkins.removeNode(Jenkins.java:1700)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:70)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
```

Log containing the `java.io.IOException: Unexpected termination of the channel`

```
jenkins_1 | INFO: Asked to provision 1 slave(s) for: docker
jenkins_1 | Jun 23, 2015 11:54:18 AM com.nirima.jenkins.plugins.docker.DockerCloud provision
jenkins_1 | INFO: Will provision 'myregistryjenkins-slave:1.609.1', for label: 'docker'
jenkins_1 | Jun 23, 2015 11:54:18 AM com.nirima.jenkins.plugins.docker.DockerCloud addProvisionedSlave
jenkins_1 | INFO: Provisioning 'myregistryjenkins-slave:1.609.1' number '0' on 'docker-slave'; Total containers: '1'
jenkins_1 | Jun 23, 2015 11:54:18 AM hudson.slaves.NodeProvisioner$StandardStrategyImpl apply
jenkins_1 | INFO: Started provisioning Image of myregistryjenkins-slave:1.609.1 from docker-slave with 1 executors. Remaining excess workload: 0
jenkins_1 | Jun 23, 2015 11:54:18 AM com.nirima.jenkins.plugins.docker.DockerCloud provisionWithWait
jenkins_1 | INFO: Trying to run container for myregistryjenkins-slave:1.609.1
jenkins_1 | Jun 23, 2015 11:54:18 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | SEVERE: Failed to connect to 172.17.42.1:32776 (try 1/10) - There was a problem while connecting to 172.17.42.1:32776
jenkins_1 | Jun 23, 2015 11:54:28 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | INFO: SSH port is open on 172.17.42.1:32776
jenkins_1 | Jun 23, 2015 11:54:28 AM com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher getSSHLauncher
jenkins_1 | INFO: Creating slave SSH launcher for 172.17.42.1:32776
jenkins_1 | Jun 23, 2015 11:54:38 AM hudson.slaves.NodeProvisioner$2 run
jenkins_1 | INFO: Image of myregistryjenkins-slave:1.609.1 provisioning successfully completed. We have now 2 computer(s)
jenkins_1 | Jun 23, 2015 11:54:39 AM hudson.slaves.CloudRetentionStrategy check
jenkins_1 | INFO: Disconnecting f37c15089b3a@docker-slave
jenkins_1 | Jun 23, 2015 11:54:39 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Disconnected computer
jenkins_1 | Jun 23, 2015 11:54:39 AM hudson.remoting.SynchronousCommandTransport$ReaderThread run
jenkins_1 | SEVERE: I/O error in channel f37c15089b3a@docker-slave
jenkins_1 | java.io.IOException: Unexpected termination of the channel
jenkins_1 | 	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:50)
jenkins_1 | Caused by: java.io.EOFException
jenkins_1 | 	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2328)
jenkins_1 | 	at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:2797)
jenkins_1 | 	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:802)
jenkins_1 | 	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:299)
jenkins_1 | 	at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:40)
jenkins_1 | 	at hudson.remoting.AbstractSynchronousByteArrayCommandTransport.read(AbstractSynchronousByteArrayCommandTransport.java:34)
jenkins_1 | 	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:48)
jenkins_1 |
jenkins_1 | [06/23/15 11:54:39] SSH Launch of f37c15089b3a@docker-slave on 172.17.42.1 failed in 1,276 ms
jenkins_1 | Jun 23, 2015 11:54:39 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Stopped container f37c15089b3acdceb99fbf4e2ec96248e301633ffe5ea918769cefbc93b3ea04
jenkins_1 | Jun 23, 2015 11:54:40 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Removed container f37c15089b3acdceb99fbf4e2ec96248e301633ffe5ea918769cefbc93b3ea04
jenkins_1 | Jun 23, 2015 11:54:48 AM com.nirima.jenkins.plugins.docker.DockerCloud provision
jenkins_1 | INFO: Asked to provision 1 slave(s) for: docker
jenkins_1 | Jun 23, 2015 11:54:48 AM com.nirima.jenkins.plugins.docker.DockerCloud provision
jenkins_1 | INFO: Will provision 'myregistryjenkins-slave:1.609.1', for label: 'docker'
jenkins_1 | Jun 23, 2015 11:54:48 AM com.nirima.jenkins.plugins.docker.DockerCloud addProvisionedSlave
jenkins_1 | INFO: Provisioning 'myregistryjenkins-slave:1.609.1' number '0' on 'docker-slave'; Total containers: '1'
jenkins_1 | Jun 23, 2015 11:54:48 AM hudson.slaves.NodeProvisioner$StandardStrategyImpl apply
jenkins_1 | INFO: Started provisioning Image of myregistryjenkins-slave:1.609.1 from docker-slave with 1 executors. Remaining excess workload: 0
jenkins_1 | Jun 23, 2015 11:54:48 AM com.nirima.jenkins.plugins.docker.DockerCloud provisionWithWait
jenkins_1 | INFO: Trying to run container for myregistryjenkins-slave:1.609.1
jenkins_1 | Jun 23, 2015 11:54:48 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | SEVERE: Failed to connect to 172.17.42.1:32777 (try 1/10) - There was a problem while connecting to 172.17.42.1:32777
jenkins_1 | Jun 23, 2015 11:54:58 AM com.nirima.jenkins.plugins.docker.utils.PortUtils bySshWithEveryRetryWaitFor
jenkins_1 | INFO: SSH port is open on 172.17.42.1:32777
jenkins_1 | Jun 23, 2015 11:54:58 AM com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher getSSHLauncher
jenkins_1 | INFO: Creating slave SSH launcher for 172.17.42.1:32777
jenkins_1 | Jun 23, 2015 11:55:08 AM hudson.slaves.NodeProvisioner$2 run
jenkins_1 | INFO: Image of myregistryjenkins-slave:1.609.1 provisioning successfully completed. We have now 2 computer(s)
jenkins_1 | [06/23/15 11:55:11] SSH Launch of ac031b44467b@docker-slave on 172.17.42.1 completed in 3,350 ms
jenkins_1 | Jun 23, 2015 11:55:30 AM hudson.model.Run execute
jenkins_1 | INFO: docker-smoke-test #7 main build action completed: SUCCESS
jenkins_1 | Jun 23, 2015 11:55:30 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Disconnected computer
jenkins_1 | Jun 23, 2015 11:55:30 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Stopped container ac031b44467bc394c7662d5eb10a87b1d3e9fbe8820f5d90b94582f1c65db025
jenkins_1 | Jun 23, 2015 11:55:30 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Shutdowned slave for ac031b44467bc394c7662d5eb10a87b1d3e9fbe8820f5d90b94582f1c65db025
jenkins_1 | Jun 23, 2015 11:55:30 AM com.nirima.jenkins.plugins.docker.DockerSlave _terminate
jenkins_1 | INFO: Removed container ac031b44467bc394c7662d5eb10a87b1d3e9fbe8820f5d90b94582f1c65db025
jenkins_1 | Jun 23, 2015 11:55:30 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:55:30 UTC 2015 by Thread[Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7,5,main] [id=560]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:63)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:55:30 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:55:30 UTC 2015 by Thread[Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7,5,main] [id=560]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.model.Computer.disconnect(Computer.java:465)
jenkins_1 | 	at hudson.slaves.SlaveComputer.disconnect(SlaveComputer.java:601)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerSlave._terminate(DockerSlave.java:147)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:67)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:55:30 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:55:30 UTC 2015 by Thread[Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7,5,main] [id=560]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at jenkins.model.Nodes$3.run(Nodes.java:165)
jenkins_1 | 	at hudson.model.Queue._withLock(Queue.java:1207)
jenkins_1 | 	at hudson.model.Queue.withLock(Queue.java:1143)
jenkins_1 | 	at jenkins.model.Nodes.removeNode(Nodes.java:160)
jenkins_1 | 	at jenkins.model.Jenkins.removeNode(Jenkins.java:1700)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:70)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
jenkins_1 | Jun 23, 2015 11:55:30 AM hudson.model.Executor finish2
jenkins_1 | WARNING: Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7 termination trace
jenkins_1 | hudson.model.Computer$TerminationRequest: Termination requested at Tue Jun 23 11:55:30 UTC 2015 by Thread[Executor #0 for ac031b44467b@docker-slave : executing docker-smoke-test #7,5,main] [id=560]
jenkins_1 | 	at hudson.model.Computer.recordTermination(Computer.java:214)
jenkins_1 | 	at hudson.model.Computer.disconnect(Computer.java:465)
jenkins_1 | 	at hudson.slaves.SlaveComputer.disconnect(SlaveComputer.java:601)
jenkins_1 | 	at jenkins.model.Nodes$3.run(Nodes.java:166)
jenkins_1 | 	at hudson.model.Queue._withLock(Queue.java:1207)
jenkins_1 | 	at hudson.model.Queue.withLock(Queue.java:1143)
jenkins_1 | 	at jenkins.model.Nodes.removeNode(Nodes.java:160)
jenkins_1 | 	at jenkins.model.Jenkins.removeNode(Jenkins.java:1700)
jenkins_1 | 	at hudson.slaves.AbstractCloudSlave.terminate(AbstractCloudSlave.java:70)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.done(DockerOnceRetentionStrategy.java:79)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy.taskCompleted(DockerOnceRetentionStrategy.java:61)
jenkins_1 | 	at hudson.slaves.SlaveComputer.taskCompleted(SlaveComputer.java:301)
jenkins_1 | 	at com.nirima.jenkins.plugins.docker.DockerComputer.taskCompleted(DockerComputer.java:73)
jenkins_1 | 	at hudson.model.queue.WorkUnitContext.synchronizeEnd(WorkUnitContext.java:145)
jenkins_1 | 	at hudson.model.Executor.finish1(Executor.java:424)
jenkins_1 | 	at hudson.model.Executor.run(Executor.java:394)
jenkins_1 |
```

KostyaSha:
@steve-jansen thanks. Could you also describe what jenkins core version and ssh slaves plugins were used? Also, have you tried jnlp launcher?

steve-jansen:
@KostyaSha  I'm using the LTS 1.609.1 war file with below plugins as a dockerized master.

> Also, have you tried jnlp launcher?

D'oh!  Didn't realize JNLP is an option now! Sweet!  Will try this :smiley: 

```
ansicolor:0.4.1
build-blocker-plugin:1.6
build-pipeline-plugin:1.4.7
build-timeout:1.14.1
build-name-setter:1.3
credentials:1.22
credentials-binding:1.4
custom-tools-plugin:0.4.4
delivery-pipeline-plugin:0.9.3
# docker-plugin:0.9.4 installed as local hpi file until jenkinsci/docker-plugin#234 ships
durable-task:1.5
envinject:1.91.3
extended-choice-parameter:0.44
git-client:1.17.1
git:2.3.5
groovy:1.25
jquery:1.11.2-0
next-build-number:1.1
# job-dsl-plugin installed as local hpi file until JENKINS-21750 is implemented
# see https://github.com/steve-jansen/job-dsl-plugin/releases
parameterized-trigger:2.26
plain-credentials:1.1
# do not upgrade promoted builds until JENKINS-22679 is truly fixed
promoted-builds:2.16
rebuild:1.24
scm-api:0.2
ssh-agent:1.7
ssh-credentials:1.10
ssh-slaves:1.9
stashNotifier:1.8
token-macro:1.10
throttle-concurrents:1.8.4
workflow-step-api:1.8
```

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
```diff
@@ -0,0 +1,153 @@
...
+        final InputStream resourceAsStream = DockerComputerJNLPLauncher.class.getResourceAsStream("init.sh");
+
+        String cdCmd = "cd /home/jenkins/";
+        String wgetSlaveCmd = "wget " + rootUrl + "jnlpJars/slave.jar -O slave.jar";
+        String jnlpConnectCmd = "java -jar slave.jar "
+                + "-jnlpUrl " + rootUrl + dockerComputer.getUrl() + "slave-agent.jnlp ";
+//                + "-secret " + dockerComputer.getJnlpMac();
+
+        String[] connectCmd = {
+                "bash", "-c", cdCmd + " && " + wgetSlaveCmd + " && " + jnlpConnectCmd
```
steve-jansen: @KostyaSha my preference would be to declare an `ENTRYPOINT` in my dockerized slave and have the plugin send in the `-jnlpUrl` and `-secret` args.  This would let we do things like baking the slave.jar into my docker build if I want to.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
```diff
@@ -0,0 +1,153 @@
...
+        final String containerId = dockerComputer.getContainerId();
+        final String rootUrl = Jenkins.getInstance().getRootUrl();
+        final DockerClient connect = dockerComputer.getCloud().connect();
+
+        // exec jnlp connection in running container
+        // TODO implement PID 1 replacement
+
+        final InputStream resourceAsStream = DockerComputerJNLPLauncher.class.getResourceAsStream("init.sh");
+
+        String cdCmd = "cd /home/jenkins/";
```
steve-jansen: @KostyaSha should `remoteFS` apply here?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
KostyaSha: I have WIP other way of running and executing container, but this what i need to control from plugin. Allowing users defining connections will be not a good solution.You can't have secret before running container, so your request doesn't make sense.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
KostyaSha: yes, need split from local work.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
steve-jansen: Perhaps this pseudo example will help me explain:

```
cat > Dockerfile << EOF
FROM myregistry/java:8

RUN mkdir /home/jenkins
WORKDIR /home/jenkins
ADD http://myartifactory/repo/slave.jar /home/jenkins

ENTRYPOINT [ "java", "-jar", "slave.jar"]
EOF

docker build myslave .

docker run myslave -jnlpUrl http://jenkins.example.com:8080/some/jnlp/url -secret someSecret
```

Above avoids the need to execute the `cd /home/jenkins` and `wget` at runtime.  [ENTRYPOINT](https://docs.docker.com/reference/builder/#entrypoint) allows you to concatenate args from a `docker run` into a base command, rather than overwriting the command.  It can be a very powerful alternative to `CMD` in a Dockerfile.

KostyaSha: Repeating again: 

> You can't have secret before running container

"I have other way of executing and running slave processes"

I don't want block this PR because of nuances that can be changed later.

steve-jansen: :+1:

steve-jansen:
@KostyaSha :+1: on the JNLP launcher for me.

A few small lessons learned for me:

### Root URL
The URL to the master needs to be routable from the slave to the master.  While this should be super obvious to anyone experienced with JNLP, I'd some twists using a dockerized master with dockerized JNLP slaves.

As context, I use Groovy code in an init.groovy.d script to set the FQDN of the Jenkins config like this:

```
def config = jenkins.model.JenkinsLocationConfiguration.get()
config.setUrl('http://jenkins.example.com:8080')
```

In my case, the FQDN was reachable externally but not between containers due `/etc/hosts` in my docker host not propagating to the slave container.  Perhaps having an option to override the value for `rootUrl` would be helpful, with a sensible default to `Jenkins.getInstance().getRootUrl()` like you have.  I've experienced problems like this as well with elastic/floating IPs vs internal IPs in OpenStack clusters, where the global root URL is routable from external points but not internal points.

### Slave FS
The hardcoded `/home/jenkins` also causes some glitches in my setup, which I could workaround if needed.  Realize you already know this :)

KostyaSha:
> The URL to the master needs to be routable from the slave to the master.

It must be in a help file, i will check, maybe i didn't commit it.

I will think also about rootUrl. Most plugins allowing redefining it, but from jenkins design and configuration it looks strange that every plugin allows changing it.

lanwen:
:+1:  works fine for me without any config change

lanwen:
## JNLP Launcher
Works only if "everyone can do anything" security is checked. Otherwize - it can't handle jnlp url contains slave name 

e.g
```
cd /home/jenkins/ && wget http://jenkins/jnlpJars/slave.jar -O slave.jar && java -jar slave.jar -jnlpUrl http://jenkins/computer/234d5af0369c@local/slave-agent.jnlp
Failing to obtain http://jenkins/computer/234d5af0369c@local/slave-agent.jnlp
java.io.IOException: Failed to load http://jenkins/computer/234d5af0369c@local/slave-agent.jnlp: 404 Not Found
	at hudson.remoting.Launcher.parseJnlpArguments(Launcher.java:275)
	at hudson.remoting.Launcher.run(Launcher.java:219)
	at hudson.remoting.Launcher.main(Launcher.java:192)
Waiting 10 seconds before retry
```

KostyaSha:
> what null means?

it means that there is no label, because 'no label' can be a valid label :)



> Works only if "everyone can do anything" security is checked.

Don't remember what implementation was pushed here. I used jnlp with secret mac auth, so additional changes shouldn't be required. But i will check.

KostyaSha:
@steve-jansen please try new build

steve-jansen:
@KostyaSha will do :smile: 

Roger, Roger

KostyaSha:
It should also fix retention strategy termination errors. I found why they caused and will also do 0.9.X hotfix.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderNewTemplate.java
```diff
@@ -148,18 +147,20 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
...
-                        sshLaunchTimeoutMinutes,
-                        jvmOptions, javaPath,
-                        memoryLimit, cpuShares,
-                        prefixStartSlaveCmd,
-                        suffixStartSlaveCmd, instanceCapStr,
-                        dnsString, dockerCommand,
-                        volumesString, volumesFrom, environmentsString, lxcConfString, hostname,
-                        bindPorts, bindAllPorts, privileged, tty, macAddress);
-                ((DockerCloud) c).addTemplate(t);
+//TODO restore after review
```
KostyaSha: Does anyone know what is the use case of this builder?

steve-jansen:
A few things didn't go well for me:

1. The init.sh script seems to fail to phone home to the Jenkins master.  At first, I thought it was just my prior problem with the Jenkins rootUrl() being wrong for my docker-compose.yml setup.   The failure would cause new nodes to be created, yet old ones were not destroyed.  Here's a screenshot of a single job build causing may nodes to be created.



2. The `@` char in the node name seems to cause problems for the JNLP URL on 1.609.1 LTS.  



3. Creating a new JNLP slave with the same name less the `@` seems to fix this.

```
docker run -it --rm --workdir /home/jenkins --user jenkins -v ${PWD}/slave.jar:/home/jenkins/slave.jar myregistry/jenkins-slave:1.609.1 java -jar slave.jar -jnlpUrl http://172.17.0.14:8080/computer/02d42e1a48b8-docker-slave/slave-agent.jnlpJul 03, 2015 4:19:36 AM hudson.remoting.jnlp.Main createEngine
INFO: Setting up slave: 02d42e1a48b8-docker-slave
Jul 03, 2015 4:19:36 AM hudson.remoting.jnlp.Main$CuiListener <init>
INFO: Jenkins agent is running in headless mode.
Jul 03, 2015 4:19:36 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Locating server among [http://172.17.0.14:8080/]
Jul 03, 2015 4:19:37 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Handshaking
Jul 03, 2015 4:19:37 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connecting to 172.17.0.14:50000
Jul 03, 2015 4:19:37 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Trying protocol: JNLP2-connect
Jul 03, 2015 4:19:37 AM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connected
```

KostyaSha:
Never comment to commits

KostyaSha:
@steve-jansen provide `docker logs` for suspended container. Suspended happens when launch was not successful and i not sure why strategy doesn't kill it.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerJNLPLauncher.java
```diff
@@ -0,0 +1,214 @@
...
+
+    public static class DescriptorImpl extends Descriptor<ComputerLauncher> {
+
+        public Class getJNLPLauncher() {
+            return JNLPLauncher.class;
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "(Experimental) Docker JNLP launcher";
```
KostyaSha: @steve-jansen i'm marking this launcher as experimental. And if it doesn't work - will fix later. I think everybody need fix related to ssh. 
Or even will remove from available for some time...

KostyaSha:
> The @ char in the node name seems to cause problems for the JNLP URL on 1.609.1 LTS.

Thanks for report, i changed naming.

> The failure would cause new nodes to be created, yet old ones were not destroyed.

This happens when launch is called and slave not connected or when slave is not accepting tasks. Added cleanups.

KostyaSha:
This is really funny, but sometimes remotingThreadPool doesn't work for me...

KostyaSha:
Works fine and better then current master. Going forward.

jglick:
Did you ever try making a launcher based on `docker-exec`? IIUC this would have the predictability of an SSH launcher combined with the low overhead of the JNLP launcher. But I have no idea if it works on Swarm.

KostyaSha:
@jglick you can't know address of container before running container.

KostyaSha:
Or what launching steps do you mean?

jglick:
Forget `docker exec`. I think you could actually just use `docker run --attach` (typically also passing `--rm=true`), assuming `ENTRYPOINT` is already set up to run `java -jar slave.jar` (else `--entrypoint` might be a workaround). Basically like a preconfigured `CommandLauncher`. Simpler than using an SSH launcher (and does not require SSHD in the container); and more straightforward than a JNLP launcher since you are not waiting for a callback: you can directly attach stdin/stdout of the container to the master side of the channel.

steve-jansen:
+1 on `docker run`.  It's way simpler and will scale very nicely if you point the Docker engine at a Swarm cluster URL.

+1 on assuming `ENTRYPOINT` will run `java -jar slave.jar` and accept extra args via the `docker run` command args.

KostyaSha:
Could you firstly learn cloud api before suggesting solutions that will have the latest positions in @stephenc  rating?

steve-jansen:
@KostyaSha not sure I understand ^^^... is that comment for me or @jglick ?

jglick:
@KostyaSha seems uninterested in using `--attach` for a launcher. Probably would work fine for a traditional Docker server, but unclear if it would scale under Swarm: depends on whether Swarm can physically hand off I/O streams without multiplexing (which in turn probably depends on how the client protocol handles streams). Beyond my current level of Docker expertise.

KostyaSha:
JNLP connects directly from slave to host, that's enough for lightweight connection. @jglick or do you know any overhead in it?

jglick:
The JNLP connection is a direct TCP socket. It is just annoying that you have to wait for it to call back to you. Generally nicer to have launchers that make the connection _from_ master _to_ the slave computer. Whether it matters in this case, I do not know.

KostyaSha:
master to slave enforces you to have open ports, that is bad.

jglick:
> master to slave enforces you to have open ports

Not with `docker run --attach`, which connects to the stdio of the main process in the container, not relying on a container port.

PR #240, opened 2015-05-29
Temp Fix volumesFrom

KostyaSha:
Works with single value now


KostyaSha:
Manually verified, should help with mounting volumesFrom at least from first entry.

PR #247, opened 2015-06-12
[FIXED JENKINS-28886] Do not register specific strategies

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-28886

@KostyaSha, @jglick

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -11,8 +11,10 @@ import com.github.dockerjava.api.model.Bind;
 import com.github.dockerjava.api.model.PortBinding;
 import com.github.dockerjava.api.model.Volume;
 import com.github.dockerjava.api.model.VolumesFrom;
+import com.nirima.jenkins.plugins.docker.strategy.DockerCloudRetentionStrategy;
 import com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy;
 import com.trilead.ssh2.Connection;
+
 import hudson.Extension;
 import hudson.Util;
 import hudson.model.*;
@@ -26,15 +28,20 @@ import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import hudson.util.StreamTaskListener;
 import jenkins.model.Jenkins;
+
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
+
 import shaded.com.google.common.base.Objects;
 import shaded.com.google.common.base.Strings;
 import shaded.com.google.common.collect.ImmutableList;
 
 import javax.ws.rs.ProcessingException;
+
 import java.io.IOException;
 import java.io.PrintStream;
 import java.util.ArrayList;
@@ -380,6 +387,15 @@ public class DockerTemplate extends DockerTemplateBase implements Describable<Do
             }
 
             return FormValidation.ok();
+        }
+
+        @Restricted(DoNotUse.class)
+        public List<Descriptor<RetentionStrategy<?>>> getDockerRetentionStrategies() {
+            List<Descriptor<RetentionStrategy<?>>> strategies = new ArrayList<Descriptor<RetentionStrategy<?>>>();
+            strategies.addAll(RetentionStrategy.all());
+            strategies.add(DockerOnceRetentionStrategy.DESCRIPTOR);
+            strategies.add(DockerCloudRetentionStrategy.DESCRIPTOR);
+            return strategies;
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


KostyaSha:
Please, firstly provide plan for fixing core.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -382,6 +389,15 @@ public FormValidation doCheckVolumesFromString(@QueryParameter String volumesFro
             return FormValidation.ok();
         }
 
+        @Restricted(DoNotUse.class)
+        public List<Descriptor<RetentionStrategy<?>>> getDockerRetentionStrategies() {
+            List<Descriptor<RetentionStrategy<?>>> strategies = new ArrayList<Descriptor<RetentionStrategy<?>>>();
+            strategies.addAll(RetentionStrategy.all());
+            strategies.add(DockerOnceRetentionStrategy.DESCRIPTOR);
+            strategies.add(DockerCloudRetentionStrategy.DESCRIPTOR);
```
KostyaSha: Have you verified that order is preserved?

olivergondza:
I hope Jesse helps me out to decide if this is an abuse of descriptors or a valid use-case. Though, we both seem to agree that those implementation should not be registered as long as they are not general purpose.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
olivergondza: No, is there any desired order? It should be trivial to create any order you like this way.

KostyaSha: This is why ordinal that you deleted exists.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -382,6 +389,15 @@ public FormValidation doCheckVolumesFromString(@QueryParameter String volumesFro
             return FormValidation.ok();
         }
 
+        @Restricted(DoNotUse.class)
+        public List<Descriptor<RetentionStrategy<?>>> getDockerRetentionStrategies() {
+            List<Descriptor<RetentionStrategy<?>>> strategies = new ArrayList<Descriptor<RetentionStrategy<?>>>();
```
KostyaSha: java 1.7 available

jglick:
:+1:, this looks right to me.

> an abuse of descriptors or a valid use-case

I think there is nothing better to be done. The design of `Describable`/`Descriptor` is backwards and illogical, but do not get me started.

KostyaSha:
Next time please squash one logical changes into one commit.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -382,6 +389,15 @@ public FormValidation doCheckVolumesFromString(@QueryParameter String volumesFro
             return FormValidation.ok();
         }
 
+        @Restricted(DoNotUse.class)
```
KostyaSha: Why?

olivergondza: No one needs to refer to the symbol for now, it is exposed for jelly consumption only. It is not a part of plugin API so it can be changed freely in the future.

KostyaSha: Really?

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/config.jelly
```diff
@@ -47,11 +47,10 @@
         </f:entry>
 
         <f:advanced>
-
             <!-- reworked strategy choosed from core -->
             <f:dropdownList name="retentionStrategy" title="${%Availability}"
                             help="/help/system-config/master-slave/availability.html">
-                <j:forEach var="d" items="${h.getRetentionStrategyDescriptors()}">
+                <j:forEach var="d" items="${instance.descriptor.dockerRetentionStrategies}">
```
KostyaSha: Why not just `descriptor.` ?

olivergondza: IIRC it did not worked.

jglick: Sounds like something else is broken then. In normal form construction `descriptor` is always set.

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/config.jelly
```diff
@@ -47,11 +47,10 @@
...
                     <j:if test="${d != null}">
                         <!-- TODO find instanceof method in jelly-->
                         <j:set var="prefix" value="${d.displayName=='Docker Once Retention Strategy'? '': 'Experimental: '}" />
```
jglick: BTW (maybe since fixed) this should be checking `d.id` to work in any locale.

jglick: (Or just edit the `getDisplayName` of that strategy to begin with, and use `f:dropdownDescriptorSelector` rather than all this stuff.)

Issue #102, opened 2014-10-14
Support removing volumes after container stop/delete


fraz3alpha:
I have a number of docker images that have a data VOLUME configured which is only needed for the lifetime of the container. When starting the container manually, unless the run command has --rm included, these containers will not delete the volume after the container has exited - as a result, I have a lot of files building up on the remote docker host.

As far as I can tell I can't see a mechanism to provide the --rm argument. Is there one? If not, is there a workaround?

fraz3alpha:
Does anyone know if this was added when the code was migrated to use docker-java? Or if the new 0.9 version will include this change request?

m1no:
I wrote a patch to support an additional option to remove future dangling volumes when the docker slave gets shutdown. I will send it upstream in the next hours.

KostyaSha:
Should work since 0.10.1

maksonlee:
Slave deletion will hang until the volume has been deleted, in our case, the volume contains lots of data, so it takes minutes to remove it, at the meantime, if you try to request a new job run, it will not be scheduled into build queue.

jenkins: 1.651
docker plugin: 0.16.0

PR #253, opened 2015-06-20
Add remove volumes option at slave shutdown

m1no:
This should be a solution to #102:

Docker Java has support for removing volumes at shutdown already. Bare with me, I'm a complete JAVA noob. I orientated myself at config options like "privileged" to get my check box in. Maybe there is unnecessary code, but i didn't knew better and it passed the local build and internal test run :)


KostyaSha:
sorry but this wouldn't be merged until #234

PR #260, opened 2015-07-05
Don't cache terminating boolean in DockerOnceRetentionStrategy.

KostyaSha:
Leads to not removed slaves.
hotfix...


KostyaSha:
Decided to do 0.10.0 release, so closing this.

PR #275, opened 2015-07-17
Fix #263. Load 0.8.0 config data.

### LINKED ISSUE #263: Upgrading from 0.8 to 0.10 causes configuration data loss
I upgraded to 0.10 from 0.8 this evening and the container settings got wiped, in particular the image names, volume mounts, DNS settings were all blank in the new configuration format.

Reverting back to 0.8 restored them.  

You should put a comment in the plugin that indicates upgrading will cause data loss.  (It's a pre-release thing so this is expected, but it is a bit painful having to re-enter everything).

KostyaSha:
config.xml splitted to test methods to have clear separation for errors.


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
```diff
@@ -141,7 +141,7 @@ private String getVolumesString() {
 
     @Deprecated
     private String getEnvironmentsString() {
-        return Joiner.on("\n").join(environment);
+        return environment == null ? "" : Joiner.on("\n").join(environment);
```
lanwen: better `Joiner.on("\n").join(trimToEmpty(environment));`

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplate2Test.java
```diff
@@ -21,6 +19,33 @@
...
+        assertThat(tBase, notNullValue());
+
+        assertThat(tBase.getDockerCommandArray()[0], equalTo("/usr/bin/start-jenkins-slave.sh"));
+        assertThat(tBase.getDnsString(), equalTo("8.8.8.8"));
+
+        assertThat(tBase.getVolumesString(), equalTo("/dev/log:/dev/log"));
+        assertThat(asList(tBase.getVolumes()), hasSize(1));
+        assertThat(tBase.getVolumes()[0], equalTo("/dev/log:/dev/log"));
+
+        assertFalse(tBase.bindAllPorts);
```
lanwen: please add simple comment to assertTrue/False, `assertFalse("bindAllPorts", tBase.bindAllPorts);` for example

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
KostyaSha: +

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplate2Test.java
KostyaSha: they are all from static config and method is pretty simple - validate loaded data. I usually set debugger breakpoint and navigate objects/errors.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
KostyaSha: Wouldn't work, environment is `String[]`

lanwen: `defaultIfNull(environment, new String[] {})` ?

KostyaSha: overheaded :)

lanwen:
:+1: for fix

Issue #79, opened 2014-08-11
Feature request: A check box for specifying whether to pull the corresponding repository before launching the container


janitza-jamu:
Hi, i want my docker cloud instances to always use the latest image for a resulting container. Therefore it would be nice, to first pull the repository on the docker host.

So every time i build a new image and push it to the registry it will soon be used by all the cloud instances.

halvorgb:
Great idea.

jan-mueller-work:
Does the java docker client lib already offer the ability to pull a repo?

halvorgb:
I began to create a pull request branching from master, but I encountered problems building the project. (Some docker-java dependency problems.) I tried to bump the docker-java version from 0.10.5-SNAPSHOT to 0.10.5, but that didn't work either (some packages referenced no longer exist in that version it seems.)

My (untested) fix for this issue is here in a .diff format. https://gist.github.com/halvorgb/c19c5076ffe6c23f65c0

Is there some known commit i could check out that builds? If my understanding is correct, this fix would be a simple one and I've already done most of it.

thomassuckow:
You can use the @magnayn docker-java if you add a snapshot repo https://github.com/jenkinsci/docker-plugin/pull/143#issuecomment-71269606

pjdarton:
+1

FYI, when I did some work on a Jenkins enhancement, I found that there was a cloudbees server looking for pull requests on the source such that all one needed to do was to issue a pull request (via GitHub) and it'd automatically build and test your source for you, and let you know if it thought the pull request looked good or not.
I believe that updating the plugin version from a -SNAPSHOT build to an official build is something that the official plugin owner would do as part of the official release process (and result in a new version appearing in the Jenkins update page).

As for the code itself, it seemed to me like the proposed code changes would result in it fetching the image whether or not it needed to.  Personally, I'd prefer to see it doing some form of up-to-date check rather than a brute-force "just re-fetch it all anyway" approach.  Not sure how easy that would be to achieve though...

KostyaSha:
@pjdarton sorry i don't understand what do you mean and what about are you talking. 

If there is different use cases, then feature should have different modes/strategies.

pjdarton:
At present, the official plugin only fetches the image if it doesn't have a copy of that image.  It does not check that copy it has matches the one in the repository, so if it has previously fetched "foo:latest" then even if there's a new "foo:latest" it will continue using the old version of foo.

[...]

PR #289, opened 2015-07-25
Refactor for pull strategy + pull strategy

KostyaSha:
There is no AMI terminology in docker. It was copy-paste from other plugins.
Called new variables to be more generic in case some other plugin decided to copy-paster in future.

Implements #79

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -7,7 +7,10 @@ import com.cloudbees.plugins.credentials.common.AbstractIdCredentialsListBoxMode
 import com.cloudbees.plugins.credentials.common.StandardCertificateCredentials;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.DockerException;
-import com.github.dockerjava.api.command.*;
+import com.github.dockerjava.api.command.CreateContainerCmd;
+import com.github.dockerjava.api.command.CreateContainerResponse;
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import com.github.dockerjava.api.command.StartContainerCmd;
 import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.api.model.Image;
 import com.github.dockerjava.api.model.Version;
@@ -67,7 +70,7 @@ public class DockerCloud extends Cloud {
     /**
      * Total max allowed number of containers
      */
-    public final int containerCap;
+    private int containerCap = 100;
 
     /**
      * Track the count per image name for images currently being
@@ -75,7 +78,7 @@ public class DockerCloud extends Cloud {
      */
     private static final HashMap<String, Integer> provisionedImages = new HashMap<>();
 
-    @DataBoundConstructor
+    @Deprecated
     public DockerCloud(String name,
                        List<? extends DockerTemplate> templates,
                        String serverUrl,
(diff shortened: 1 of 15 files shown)


KostyaSha:
Closes #79

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategy.java
```diff
@@ -0,0 +1,56 @@
+package com.nirima.jenkins.plugins.docker;
+
+/**
+ * @author Kanstantsin Shautsou
+ */
+public enum DockerImagePullStrategy {
+
+    PULL_ALWAYS("Pull all images every time") {
```
lanwen: how can we localize it?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -252,16 +279,51 @@ public static String runContainer(DockerTemplateBase dockerTemplateBase,
...
+            LOGGER.info("Pulling image '{}' since one was not found.  This may take awhile...", imageName);
+            //Identifier amiId = Identifier.fromCompoundString(ami);
+            try (InputStream imageStream = getClient().pullImageCmd(imageName).exec()) {
+                int streamValue = 0;
+                while (streamValue != -1) {
+                    streamValue = imageStream.read();
+                }
+            }
+
+            LOGGER.info("Finished pulling image '{}'", imageName);
```
lanwen: if we don't pull image, why we write that we finished?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+        public boolean pullIfNotExists(String imageName) {
+            return true;
+        }
+
+        @Override
+        public boolean pullIfExists(String imageName) {
+            return imageName.endsWith(":latest");
+        }
+    },
+    PULL_NEVER("Never pull") {
```
lanwen: think use case in javadoc will be useful

KostyaSha: Use case it up to user, just implement ability to not pull any images, i.e. somebody on mobile on dial-up

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategy.java
KostyaSha: Probably, https://github.com/jenkinsci/jenkins/blob/9c443c8d5bafd63fce574f6d0cf400cd8fe1f124/core/src/main/java/hudson/model/BallColor.java#L57
But if don't want waste time on localisation :)

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategy.java
lanwen: who run slave via mobile internet? :)

KostyaSha: or just in high cost vm. Do you have any objections for it? ;)

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
KostyaSha: We are pulling, this block is only for pulling image

lanwen: miss end of "if" statement

lanwen:
for code :+1: , but you should think about localization and tests. Its not a waste of time

KostyaSha:
@lanwen i think functionality is more critical and i'm cycling between different usage blockers. I'm ok to do translations when code will be stable. 
I aslo thinking about uses ReposTag object instead of String, but i fear changed in docker-java and will keep as is for now.

KostyaSha:
@magnayn never comment single commits, use `changes` tab instead

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -96,16 +100,46 @@ public DockerCloud(String name,
...
     }
 
     public int getConnectTimeout() {
         return connectTimeout;
     }
 
+    /**
+     * @deprecated use {@link #getContainerCap()}
+     */
+    @Deprecated
```
KostyaSha: restored this method to have ABI/API compatibility

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -252,16 +293,51 @@ public static String runContainer(DockerTemplateBase dockerTemplateBase,
...
+                    streamValue = imageStream.read();
+                }
+            }
+
+            LOGGER.info("Finished pulling image '{}'", imageName);
+        }
+    }
 
     private DockerSlave provisionWithWait(DockerTemplate dockerTemplate) throws IOException, Descriptor.FormException {
+        pullImage(dockerTemplate);
```
KostyaSha: moved `pullImage` into this block

KostyaSha:
Verified manually plus @DanielHouston promised to test it on Monday.
Merging, but @lanwen feel free to review.

PR #295, opened 2015-08-14
Clarify field name according to docker api docs

KostyaSha:
https://docs.docker.com/reference/api/docker_remote_api_v1.18/#build-image-from-a-dockerfile
TODO expose all options
@lanwen


lanwen:
:+1:

magnayn:
Yeah.

Docker's terminology is hideous :-(

KostyaSha:
@magnayn their API also sucks
```
New!

The fields Debug, IPv4Forwarding, MemoryLimit, and SwapLimit are now returned as boolean instead of as an int.
```
wtf..

Issue #235, opened 2015-05-25
Roadmap to 1.0


KostyaSha:
Let's try summarise what is really required to be 1.0:
- [x]  unbundle launchers and add JNLP launcher (allows choosing connection variant with different credentials)  #234 
- [x] connection robustness (reconnections, etc) #150 #230 
- [x] pulling improvements #79 #289 (0.11.0)
- [ ] integrational tests and ath 
- [ ] release JNLP launcher as experimental (should bypass #206 #219)
- [ ] actions for tagging, pushing to registry  https://github.com/jenkinsci/docker-plugin/labels/tag
- [ ] Support for TLS #245 (review as it should work but doesn't?)
- [ ] rework docker builder ( #296 ...)
- [ ] swarm integration ( #299 )
- [x] unbundle docker connection credentials (?)
- [ ] review steps integrations into possible flows (#301)
- [x] load/performance #221 
- [ ] imho migrate to Lists instead arrays and keep data migration (or even wrote new DockerContainerConfiguration or DockerTemplateBase2 class that will have Lists, Describable, migrate to it and simply reuse from other places)
- [ ] share templates configuration between clouds #238
- [x] other refactoring #231
- [ ] Review security
- [ ] check API exposing
- [ ] logging (Docker cloud provisioning is good now, but time measurement logging TODO)
- [ ] ... 
- [ ] functional tests (@lanwen very want to help), after code base will be stable/simplified
- [ ] check/update documentation

I think we can slowly releasing 0.10.0, 0.11.0 and etc with small portions of changes and when everything will work do final 1.0

KostyaSha:
@magnayn @thomassuckow :pager: 
First item was picked suddenly, i found this very interesting and started PoC, then found that UI and Templates part can be simplified.

magnayn:
+1 for doing 0.10.0 etc.

[] It strikes me that the templates should be defined outside of the clouds so they can be reused. (I have a 'build cluster' of 5 or 6 machines, and I end up having to specify the jenkins image for each.

KostyaSha:
@magnayn i think you must use swarm that is compatible by API and can spread containers load.
Btw i will do 0.9.0 today, have good feedback for current master.

KostyaSha:
@magnayn seems i got your idea.

KostyaSha:
If launchers unbundling failure, i will skip it and merge @lanwen commit and continue doing other items.

KostyaSha:
[...]

Issue #117, opened 2014-11-14
Internal Server Error when trying to provision image


gdubya:
I'm getting started with Docker and the Jenkins plugin but having some trouble.

**Versions**
Jenkins: 1.589
Docker Plugin: 0.8
Docker: 1.3.1

In Jenkins: I've created a simple Jenkins job to execute "ping -n 10 www.google.com"
I created a simple Docker image as described on the Docker plugin wiki page and named it "jenkins-2". If I test the image from the command line it works fine (can run /bin/bash, su to the jenkins user, etc.) but when I try to build the job I see the error below repeated in the logs.

How can I find more details regarding what is causing the internal server error?

> Nov 14, 2014 11:24:35 PM com.nirima.jenkins.plugins.docker.DockerCloud provision
> INFO: Excess workload after pending Spot instances: 1
> Nov 14, 2014 11:24:35 PM com.nirima.jenkins.plugins.docker.DockerCloud countCurrentDockerSlaves
> INFO: Images found: [Image{repository=null, tag=null, id=b6e41c882026939401203ab667d06dee0d53a4b6e26dc541934952b658a6ce4d, created=1416001376, size=334164, virtualSize=762138779}]
> Nov 14, 2014 11:24:35 PM com.nirima.jenkins.plugins.docker.DockerCloud addProvisionedSlave
> INFO: Provisioning for AMI jenkins-2; Estimated number of total slaves: 0; Estimated number of slaves for ami jenkins-2: 0
> Launching jenkins-2
> Nov 14, 2014 11:24:35 PM hudson.slaves.NodeProvisioner$StandardStrategyImpl apply
> INFO: Started provisioning Image of jenkins-2 from MyDocker with 1 executors. Remaining excess workload: 0
> Nov 14, 2014 11:24:35 PM com.nirima.jenkins.plugins.docker.DockerTemplateBase getLxcConf
> WARNING: Specified option:  is not in the form X=Y, please correct.
> Nov 14, 2014 11:24:36 PM com.nirima.jenkins.plugins.docker.DockerCloud$1 call
> SEVERE: Error in provisioning; slave=null, template=DockerTemplate{image=jenkins-2, parent=DockerCloud{name=MyDocker, serverUrl=http://localhost:4243}}
> javax.ws.rs.InternalServerErrorException: HTTP 500 Internal Server Error
>     at org.glassfish.jersey.client.JerseyInvocation.convertToException(JerseyInvocation.java:929)
[...]

PR #299, opened 2015-08-22
feat(DockerCloud): option to use docker hostname as slave display name

menski:
Adds an option to the docker cloud to use the hostname of the docker
host as slave display name. If the hostname cannot be resolved the IP
address is used.

This is useful for setups with docker swarm. The display name of the
slave will then contain the information on which actual docker host
the job is running.

Limitation: The docker container must have a port mapping to
get the hostname of the slave.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/299)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -12,7 +12,9 @@ import com.github.dockerjava.api.command.CreateContainerResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.command.StartContainerCmd;
 import com.github.dockerjava.api.model.Container;
+import com.github.dockerjava.api.model.ExposedPort;
 import com.github.dockerjava.api.model.Image;
+import com.github.dockerjava.api.model.Ports;
 import com.github.dockerjava.api.model.Version;
 import com.github.dockerjava.core.NameParser;
 import com.nirima.jenkins.plugins.docker.launcher.DockerComputerLauncher;
@@ -41,6 +43,8 @@ import javax.servlet.ServletException;
 import javax.ws.rs.ProcessingException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.*;
 import java.util.concurrent.Callable;
 
@@ -62,6 +66,7 @@ public class DockerCloud extends Cloud {
     public final int readTimeout;
     public final String version;
     public final String credentialsId;
+    public final boolean displayHostname;
 
     private transient DockerClient connection;
 
@@ -84,7 +89,8 @@ public class DockerCloud extends Cloud {
                        int connectTimeout,
                        int readTimeout,
                        String credentialsId,
-                       String version) {
+                       String version,
+                       boolean displayHostname) {
         super(name);
         Preconditions.checkNotNull(serverUrl);
         this.version = version;
@@ -92,6 +98,7 @@ public class DockerCloud extends Cloud {
         this.serverUrl = serverUrl;
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
+        this.displayHostname = displayHostname;
 
         if (templates != null) {
             this.templates = new ArrayList<>(templates);
@@ -114,7 +121,8 @@ public class DockerCloud extends Cloud {
                        int connectTimeout,
                        int readTimeout,
                        String credentialsId,
-                       String version) {
+                       String version,
+                       boolean displayHostname) {
         super(name);
         Preconditions.checkNotNull(serverUrl);
         this.version = version;
@@ -122,6 +130,7 @@ public class DockerCloud extends Cloud {
         this.serverUrl = serverUrl;
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
+        this.displayHostname = displayHostname;
 
         if (templates != null) {
             this.templates = new ArrayList<>(templates);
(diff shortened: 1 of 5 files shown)


lanwen:
Think it should be added to page "Built on Docker" and not slave name. Docker swarm is abstraction layer to avoid such interaction with each concrete host. But the information should be available by request.

If this information will be on page for https://github.com/jenkinsci/docker-plugin/blob/master/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/action/DockerBuildAction.java - is it solves your problem?

menski:
We like to see the actual host of the build while it runs to easily spot problems. This way we can already see in the Jenkins executor overview when something goes wrong with the scheduling or a host isn't accepting any jobs.

As this is an option it is the decision of the user if he wants to see this information or not.

lanwen:
What kind of problems it helps to solve? Maybe just extend logging? Think hostname exposing is wrong way to research slave provisioning problems

menski:
We just want to quickly see where a job is actually running. This helped us to spot problems early and it adds some useful information to the slave name.

lanwen:
With this logic it will be useful to add image name, exposed ports, image hash, attached volumes etc :) IMHO If you get regular problems and spot them with help of hostname, probably you do something wrong.

KostyaSha:
Sorry, but i wouldn't merge it now, because i plan review swarm integration separately. I will add it to roadmap. Cloud name + container hash is really unique identifier, if you want see more info, then it should be displayed in slave view page (in TODO review this API provided by jenkins: afair you just need extended configure.jelly for slave object). Also plugin has docker cloud page, then is in bad state and may show more advanced info.

KostyaSha:
@menski if you want you can help me with #299 ;)

menski:
@KostyaSha do you mean #235? Or any specific issue? As #299 is the current one.

KostyaSha:
@menski sorry, wrong url in buffer https://github.com/docker-java/docker-java/pull/299 (seems we have two PRs with the same id?)

KostyaSha:
> (in TODO review this API provided by jenkins: afair you just need extended configure.jelly for slave object).

@menski ^^

nnordrum:
Because I love to derail things... wouldn't it be better to create a field with that is evaluated with the Token-Macro plugin and then also publish some additional tokens that could be used?

I just can't imagine we could possibly come up with a strategy that makes everybody happy, and this way you could create a "sensible default", and then let @menski use hostname if desired, and I could use the name of the committer (since for us, commit = container, and toss them after each job), which I've had at least 5 people independently ask for...

magnayn:
So anyone who has used docker in production will know precisely why knowing
the actual running host is important. That's the whole reason the slave
name is prepended with the cloud name in the first place. So I reject the
notion that 'if you want to know the name, you're doing it wrong' or that
it should be hidden from view. It's a useful feature.

That said, where there's 2 options, there's more than likely >2. I quite
like the @nnordrum idea of an expandable token macro name. Best solution
would probably be a dropdown list of 'naming strategy', with perhaps
options of
default - automatically choose (we may detect later that it's a swarm and
choose automatically)
cloud - as it is today
host - as the PR does it

then later: other - provides text box to provide a string (expandable
macro).

then if someone comes up with something un-thought-of, we don't need
another boolean flag to select it.

@menski / @nnordrum - would you be interested in re-spinning the feature in
this way?

magnayn:
@nnordrum BTW - I don't think you'll be able to name the slave based on the commit (or author) that it's going to build, as that isn't the order of things in Jenkins. Slaves that are provisioned have no knowledge of the build that they are going to execute.

nnordrum:
@magnayn I completely agree at the time of provision, but is there some reason why the slave couldn't rename itself? I'm far from an expert on the internals of jenkins slave naming... 

I'm basing this almost completely on the fact that the Build Executor Status box's entire innerHTML is replaced from ajaxExecutors, so we should be able to rename the computer/slave/whatever it's called and then it would show up "soon".

magnayn:
Sure, you could do that - the slave's name can be changed from within the
build; you'd need to add a build step within the project configuration that
does that.

Such a build step though wouldn't be tied neccesarily to docker-plugin
however, you could equally well apply 'expand tokens in this string and set
the slave name' to any project.

nnordrum:
I could just add a system groovy command then I bet.

any pointers on how to rename a slave? :)

nnordrum:
ok that didn't work as expected... created a Freestyle Job with an "Execute system Groovy script" step
```
out.println build.builtOn.name
build.builtOn.name = "${build.builtOn.name}-RENAMED"
out.println build.builtOn.name
```
and an Execute shell step
```
sleep 120
```

Fails almost immediately
```
[EnvInject] - Loading node environment variables.
Building remotely on BuildSwarm-f944ad2dfabd (maven docker groovy java) in workspace /home/jenkins/workspace/test-rename
BuildSwarm-f944ad2dfabd
BuildSwarm-f944ad2dfabd-RENAMED
ERROR: Build step failed with exception
java.lang.NullPointerException: no workspace from node DockerSlave{name=BuildSwarm-f944ad2dfabd-RENAMED, containerId=f944ad2dfabd, template=DockerTemplate{configVersion=2, labelString='maven java docker groovy', launcher=com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher@1286d72e, remoteFsMapping='', remoteFs='/home/jenkins', instanceCap=2147483647, mode=EXCLUSIVE, retentionStrategy=com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy@4094db2b, numExecutors=1, dockerTemplateBase=DockerTemplateBase{image=jenkins_slave}, removeVolumes=false, pullStrategy=PULL_NEVER}} which is computer null and has channel null
	at hudson.tasks.CommandInterpreter.perform(CommandInterpreter.java:76)
	at hudson.tasks.CommandInterpreter.perform(CommandInterpreter.java:66)
	at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
	at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:779)
	at hudson.model.Build$BuildExecution.build(Build.java:205)
	at hudson.model.Build$BuildExecution.doRun(Build.java:162)
	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:537)
	at hudson.model.Run.execute(Run.java:1744)
	at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:374)
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

Any ideas?

KostyaSha:
@nnordrum you can't do this.

nnordrum:
@KostyaSha I didn't think you could, but I wasn't really in a position to argue...

magnayn:
You might be able to get it to work, but it's likely to be a certain amount of pain.

The `build` stores a string, which is the name of the slave that it is being built on (`String builtOn`)

`build.builtOn` returns `Jenkins.getInstance().getNode(builtOn);`. In the jenkins instance, it has a hashmap of names of nodes to their instances.

By changing the slave's name, you're actually changing its whole identity (equals(), hashCode). You might be able to get around this by 
- changing the actual builtOn string in the build, as well as the name of the slave
- adding the renamed node into the jenkins node map

Cleanup at the end is likely to be a pig, and you'll either have to change everything that was using the old node id to the new one, or live with 2 slaves until the end (I.E: most objects store ID - String - references to objects that are looked up, rather than direct references)

Sounds like too much trouble to me. Things really don't expect the ID to change once it's set, and the slave is always instantiated without any knowledge of what it's about to build.

As an aside - you may then see the attractive 'description' field in Slave, only to discover that it's `final`. (that's right kids. The identity is mutable,and the description is immutable :facepunch: )

You might want to ask on the ML if it's possible to extend the UI items in the build executor status list (LHS) in the same way that you can add extra status columns for builds, I think that woiuld be generally useful (e.g: not only is this build #117 but it's of PR #32).

nnordrum:
what about `Node.getDisplayName()`? (i.e. `DockerSlave.getDisplayName()`)

```
    public String getDisplayName() {
        return getNodeName(); // default implementation
    }
```

That *shouldn't* have any of those side effects--right?

KostyaSha:
Display names will be the right direction.

magnayn:
It won't  have those side-effects, and sadly it won't work (in that I don't think it will give you what you want).

You can override both getDisplayName() and getNodeDescription().

The former isn't used in the UI (It'll simply continue to display cloudname-xxxxxxx). The latter appears if you click the node for it's details (appears in brackets), but after the machine is deleted of course even that disappears.

KostyaSha:
Description is different. 
Hint: queue should support display names in the same way as job/folder.

magnayn:
Yes, it should, but it doesn't appear to.

I think a wider Jenkins change is the way to go.

There are few use-cases here.

1) I use docker-swarm. When a machine is provisioned (I.E: *before the
build starts*), I would like to see the actual host it was provisioned on
(particularly for diagnosis features).
2) I would like to see, for the builds in progress, details about that
build such as
 - the SHA1
 - the author / owner
 - the PR number
 - etc etc etc

#1 can probably be done with either changing the naming policy for the ID,
or the display name if this can be added to the build executor status view.
#2 smells like a hack. Whilst I'm sure it could be done, it'd be much nicer
if the build executor status list could show additional properties of the
things that it's building other than just the build number, perhaps from
existing actions that have been added into the build objects. People using
clouds other than docker would also likely find this useful.

KostyaSha:
@magnayn for displayNames i filled https://issues.jenkins-ci.org/browse/JENKINS-31560 , that should be pretty simple change.

for 2) It should be done via additional page in jenkins because this small executors list widget will be too overloaded. Showing detailed info makes sense only for those who wants see it. Then such page can operate in the same way as 'Views'. At all it can be just one more view.

KostyaSha:
Lol, it already supported. Just Override getDisplayName and additional field in DockerComputer, then set it whenever you want, if you need attach something to build for history, then just append action onTaskCompleted :)

ndeloof:
doesn't seem to be relevant now docker swarm mode supersede docker-swarm

PR #300, opened 2015-08-24
chore(docker-java): update docker-java version to 2.0.2

menski:
- update docker-java dependency to 2.0.2
- uses async docker cmd callbacks for Pull/Push/BuildImage cmd
- read timeout was moved from `DockerClientConfig` to
  `DockerCmdExecFactoryImpl`

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -6,6 +6,7 @@ import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.AbstractIdCredentialsListBoxModel;
 import com.cloudbees.plugins.credentials.common.StandardCertificateCredentials;
 import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.DockerClientException;
 import com.github.dockerjava.api.DockerException;
 import com.github.dockerjava.api.command.CreateContainerCmd;
 import com.github.dockerjava.api.command.CreateContainerResponse;
@@ -15,6 +16,7 @@ import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.api.model.Image;
 import com.github.dockerjava.api.model.Version;
 import com.github.dockerjava.core.NameParser;
+import com.github.dockerjava.core.command.PullImageResultCallback;
 import com.nirima.jenkins.plugins.docker.launcher.DockerComputerLauncher;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
@@ -319,14 +321,14 @@ public class DockerCloud extends Cloud {
 
             long startTime = System.currentTimeMillis();
             //Identifier amiId = Identifier.fromCompoundString(ami);
-            try (InputStream imageStream = getClient().pullImageCmd(imageName).exec()) {
-                int streamValue = 0;
-                while (streamValue != -1) {
-                    streamValue = imageStream.read();
-                }
+            try {
+                getClient().pullImageCmd(imageName).exec(new PullImageResultCallback()).awaitSuccess();
+                long pullTime = System.currentTimeMillis() - startTime;
+                LOGGER.info("Finished pulling image '{}', took {} ms", imageName, pullTime);
             }
-            long pullTime = System.currentTimeMillis() - startTime;
-            LOGGER.info("Finished pulling image '{}', took {} ms", imageName, pullTime);
+            catch (DockerClientException e) {
+                LOGGER.error("Error while pulling image '{}'", imageName, e);
+            }
         }
     }
 
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -321,14 +321,9 @@ public class DockerCloud extends Cloud {
 
             long startTime = System.currentTimeMillis();
             //Identifier amiId = Identifier.fromCompoundString(ami);
-            try {
-                getClient().pullImageCmd(imageName).exec(new PullImageResultCallback()).awaitSuccess();
-                long pullTime = System.currentTimeMillis() - startTime;
-                LOGGER.info("Finished pulling image '{}', took {} ms", imageName, pullTime);
-            }
-            catch (DockerClientException e) {
-                LOGGER.error("Error while pulling image '{}'", imageName, e);
-            }
+            getClient().pullImageCmd(imageName).exec(new PullImageResultCallback()).awaitSuccess();
+            long pullTime = System.currentTimeMillis() - startTime;
+            LOGGER.info("Finished pulling image '{}', took {} ms", imageName, pullTime);
         }
     }
 
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
@@ -1,5 +1,7 @@
 package com.nirima.jenkins.plugins.docker.builder;
 
+import com.github.dockerjava.api.model.PullResponseItem;
+import com.github.dockerjava.core.command.PullImageResultCallback;
 import com.nirima.jenkins.plugins.docker.*;
 import shaded.com.google.common.base.Strings;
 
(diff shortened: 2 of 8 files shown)


menski:
This can't be merge without a docker-java release with this [fix](https://github.com/docker-java/docker-java/commit/74caed563ec466465a1eb39335519f5d0b2351a4) in docker-java. Otherwise there will be a classloader problem in jenkins.

KostyaSha:
@menski i wouldn't merge not released version, that why i'm working in upstream.

Also, why serviceLoader is required for docker-plugin at all?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -319,14 +321,14 @@ public boolean apply(Image image) {
...
-            try (InputStream imageStream = getClient().pullImageCmd(imageName).exec()) {
-                int streamValue = 0;
-                while (streamValue != -1) {
-                    streamValue = imageStream.read();
-                }
+            try {
+                getClient().pullImageCmd(imageName).exec(new PullImageResultCallback()).awaitSuccess();
+                long pullTime = System.currentTimeMillis() - startTime;
+                LOGGER.info("Finished pulling image '{}', took {} ms", imageName, pullTime);
+            }
```
KostyaSha: `} catch (...) {`

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -319,14 +321,14 @@ public boolean apply(Image image) {
...
+            catch (DockerClientException e) {
+                LOGGER.error("Error while pulling image '{}'", imageName, e);
```
KostyaSha: Don't catch, errors are catched and reported on root provisioning level.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
```diff
@@ -87,9 +89,20 @@ public void execute(AbstractBuild<?, ?> build) throws DockerException, IOExcepti
 
         LOGGER.info("Pulling image " + xImage);
 
-        InputStream result = client.pullImageCmd(xImage).exec();
+        final StringBuilder resultBuilder = new StringBuilder();
 
-        String strResult = IOUtils.toString(result);
+        PullImageResultCallback resultCallback = new PullImageResultCallback() {
+            public void onNext(PullResponseItem item) {
```
KostyaSha: don't fill stringbuilder, print into listener

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -271,14 +268,19 @@ private String pushImages() throws IOException {
 
                 Identifier identifier = Identifier.fromCompoundString(tagToUse);
 
-                PushImageCmd pushImageCmd = getClient().pushImageCmd(identifier);
+                PushImageResultCallback resultCallback = new PushImageResultCallback() {
+
+                    public void onNext(PushResponseItem item) {
+                        resultBuilder.append(item.toString());
```
KostyaSha: the same, print into listener!

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/client/ClientBuilderForPlugin.java
```diff
@@ -15,6 +15,7 @@
 public class ClientBuilderForPlugin  {
 
     private final DockerClientConfig config;
+    private Integer readTimeout;
```
KostyaSha: use primitive

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/client/ClientBuilderForPlugin.java
```diff
@@ -32,10 +33,17 @@ public static ClientBuilderForPlugin getInstance(ClientConfigBuilderForPlugin do
         return new ClientBuilderForPlugin(dockerClientConfig.build());
     }
 
+    public ClientBuilderForPlugin withReadTimeout(Integer readTimeout) {
```
KostyaSha: primitive

menski:
@KostyaSha I just did the update and wanted to have it tracked her until next docker-java is released. Service Loader is not required by docker-plugin but without the docker-java fix it will always require a DockerCmdExecFactory service in the classpath otherwise it fails during initializing of the DockerClientBuilder class.

KostyaSha:
@menski ok. I thought i already fixed it before... maybe @magnayn fix for dockerBuilderPublisher restored it...

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/client/ClientBuilderForPlugin.java
menski: But the readTimeout can be null if non is set.

KostyaSha: only if we didn't enforce default on our side...

menski: what should be the default? there was non before?

KostyaSha: hm, some timeout existed before in configuration but not used AFAIR, requires review...

menski:
@kohsuke I tried to fix your comments. I will squash the commits if the fixes are okay.

KostyaSha:
> @kohsuke I tried to fix your comments. I will squash the commits if the fixes are okay.

I think it to me :)

menski:
oh sorry :(

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
```diff
@@ -87,10 +89,16 @@ public void execute(AbstractBuild<?, ?> build) throws DockerException, IOExcepti
 
         LOGGER.info("Pulling image " + xImage);
 
-        InputStream result = client.pullImageCmd(xImage).exec();
+        PullImageResultCallback resultCallback = new PullImageResultCallback() {
+            public void onNext(PullResponseItem item) {
+                if (item.getStatus() != null && item.getProgress() == null) {
```
KostyaSha: what is `getProgress` ?

menski: if you pull a layer there will be this nice `[====>            ]` progress indicator. So the status will be the same with an updated progress until the layer was pulled completely. I thought it would be enough to log every status once and not the progress.

KostyaSha:
(edited and deleted mistaken comments)

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
KostyaSha: progress is ok because i can see then it works, also all tools to the same. What is will be ideal is progress with `append`! But then we should know some granularity or diff to previous line... Try store curLine, nextLine, and if progress is not null, then `nextLine.replace(curline, '')` and append this tail to listener!

KostyaSha: And you are printing into LOGGER, not listener, it this place has no listener, then no need printing at all imho.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -230,19 +225,20 @@ public String invoke(File f, VirtualChannel channel) throws IOException, Interru
...
-                                    String text = item.getStream();
-                                    if (text != null) {
-                                        listener.getLogger().append(text);
-                                        if (text.startsWith("Successfully built ")) {
-                                            imageId = StringUtils.substringBetween(text, "Successfully built ", "\n").trim();
+                                BuildImageResultCallback resultCallback = new BuildImageResultCallback() {
+                                    public void onNext(BuildResponseItem item) {
+                                        String text = item.getStream();
+                                        if (text != null) {
+                                            listener.getLogger().append(text);
```
KostyaSha: Could you try this https://github.com/jenkinsci/docker-plugin/pull/300/files#r37809619 trick for builder/publisher?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
KostyaSha: All this control classes need to be updated and provide listener for printing output.

KostyaSha:
@menski i summarised in https://github.com/jenkinsci/docker-plugin/milestones/0.13.0 what i want release ASAP when docker-java releases 2.1.0. Many small things. Do you agree?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
KostyaSha: created #305 will handle myself

KostyaSha:
2.1.0 released, i'm working on 0.13.0 milestone

menski:
@KostyaSha I'm currently not able to work on this. But I think the issues for 0.13.0 are okay and would be great to have this release as fast as possible. I will try to fix this pull request on sunday.

KostyaSha:
I picked your commits and reworking, because they are breaking existing things

KostyaSha:
> DockerCmdExecFactory service in the classpath otherwise it fails during initializing of the DockerClientBuilder class.

Do you really need it? I just excluded using it because plugin mostly handles options on it's side.

PR #311, opened 2015-08-31
Push to custom repo and force

jphelion:
This version adds a field for a custom/private repository to be named and also a checkbox for allowing users to forcefully tag their images. Also added a default tag (latest) if no tag has been given. This will still commit an image with the build tag (e.g. b3), but for the push to the repository it will tag it as the user has specified with the additionalTag option.


KostyaSha:
Hi, i'm in process of reviewing API and options and want resolve some things in upstream firstly. This change looks ugly: unrelated commits in PR, wrong coding style changes, constructor breakage.

Please fill issue with options that you want and i will implement myself.

jphelion:
Hi,

Sorry for the slow response and I'm very sorry for the bad quality of the pull request, it was my first like this and I managed to overlook the entire commit log. 

Regarding to the coding style I tried to keep the code looking as it was before i started editing so I think it was more of a case of badly naming the commit. All I did was mainly remove some whitespaces I had created. 

The build failure was also a huge mistake on my part as I forgot to properly test the code using the latest version of the official plugin.

I will post a feature request issue for these.

PR #323, opened 2015-09-17
Use new line separated UI. Added help file.

### LINKED ISSUE #322: help file is missing for "environments" template field
What format should I use for Environment overrides?

Per the https://docs.docker.com/reference/builder/#env I tried:
```
http_proxy='myproxy' https_proxy='myproxy'
http_proxy 'myproxy' https_proxy 'myproxy'
```
Despite this, it never seems to set / override the base image ENV http_proxy var. Am I using the correct format?

KostyaSha:
On data structure it still String[]
Fixes #322


KostyaSha:
Verified manually, works.

PR #324, opened 2015-09-17
Minimise retry delay to 2 sec, total timeout 120 sec.

KostyaSha:
Still not the best and final solution.


KostyaSha:
Tested, works fine.

PR #328, opened 2015-09-20
Update core to 1.609.3

KostyaSha:
Previously i requested to keep 1.565.3 and supported it. Now i have no such need and if somebody needs more ancient core, please inform. Will keep time window before using newer APIs.


KostyaSha:
@magnayn @thomassuckow FYI

thomassuckow:
I lag latest by about a week, so no concern from me.

PR #329, opened 2015-09-22
Remove api version and enforce default capacity in jelly

KostyaSha:


lanwen:
:+1:

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerCloud/config.jelly
```diff
@@ -9,9 +9,9 @@
         <f:textbox/>
     </f:entry>
 
-    <f:entry title="${%Docker Version}" field="version">
-        <f:textbox/>
-    </f:entry>
+    <!--<f:entry title="${%Docker Version}" field="version">-->
```
lanwen: Think you can remove it completely. Git saves the history in case of restoring

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -63,7 +63,12 @@
     public final String serverUrl;
     private int connectTimeout;
     public final int readTimeout;
-    public final String version;
+    /**
+     * Confuses users and no real use cases.
+     * TODO remove
```
lanwen: mb write when

KostyaSha: 1.0 or when i decide to break it :) Nobody depends on this plugin AFAIK and job-dsl is not implemented for plugin.

docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerCloud/config.jelly
KostyaSha: I will remove all references when Constructor will be clean-uped.

PR #331, opened 2015-09-25
Contribute vars

KostyaSha:


KostyaSha:
@lanwen

oleg-nenashev:
Any plans regarding 1.0 with a stable API?

KostyaSha:
> Any plans regarding 1.0 with a stable API?

In situation that CloudBees hijacks features and producing different docker-plugins, not sure.

oleg-nenashev:
@KostyaSha 

> In situation that CloudBees hijacks features and producing different docker-plugins, not sure.

IIRC [this plugin](https://github.com/ndeloof/docker-slaves-plugin) has been developed by @ndeloof and @ydubreuil during a hackaton on the weekend. They are CloudBees employees, but this plugin is **their own initiative**, which is **completely unrelated** to their work assignments. Any CloudBees employee is eligible to perform individual open-source contributions without prior approval, so please do not involve the company. 

If you want to discuss this duplication, start a thread in jenkinsci-dev or contact Nicolas and Yoann directly. The plugin is in the beta state, hence it's a right time to discuss such things.

Personally I'm also concerned about a new plugin, which overlaps with Docker plugin. It would e much better to contribute into the existing plugin instead of creating new ones. Probably @ndeloof and @ydubreuil selected another way, because they feel uncomfortable about the negotiation with the current plugin maintainer.

KostyaSha:
> so please do not involve the company.

Plugin is written under CB license headers and CB packages.

Instead of fixing APIs they did hacks.

@oleg-nenashev also this is off-topic for this PR.

oleg-nenashev:
> also this is off-topic for this PR.

Probably. My question was about just another API compatibility breakage, so I was just interested when is it going to finish.

KostyaSha:
`onConnected` was used anywhere, no compatibility breakage here.

ndeloof:
docke-slaves plugin indeed is a pure hacker experiment, not related to a CloudBees product plan (it's actually disturbing from this point of view), and we did it on our spare time (have a look at the commit history over the week-end if you need some proof of it). We use cloudbees namespace as this plugin might at some time become stable enough so the company would like to adopt it as an official plugin and have engineering to support and improve it, but it's licensed under MIT, not CB license.

@oleg-nenashev If you had a quick look at our plugin, you may notice it's a complete distinct experimental approach, so couldn't be developed as an evolutionary PR over docker-plugin.

@KostyaSha it's definitively a hack, we never claimed it isn't. If this experiment demonstrate to be efficient, we might discuss with community how to get jenkins API improved to cover this approach. But I don't thing one can come with a fresh new design without previous experiments.

ydubreuil:
As Nicolas already said, docker slaves plugin is really more an experiment than anything else. Our goal was to try a new way to use Docker. Our plugin doesn't use containers as a kind of lightweight VM like docker-plugin does, but see the docker API as a replacement of the API to start processes. To make it work, we had to use some Jenkins API in ways that were not intended. Our main goal was to prove it was possible.

It was not possible to propose PR on the original plugin due to this completely different design.

KostyaSha:
> see the docker API as a replacement of the API to start processes.

@stephenc ?

> Probably. My question was about just another API compatibility breakage, so I was just interested when is it going to finish.

Did you get any specific issues or just replying about backward compatibility?

PR #343, opened 2015-10-23
Hack around issue with Shade plugin and Maven 3.3.x

stephenc:
@reviewbybees

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/343)

### DIFF
--- a/docker-java-shaded/pom.xml
+++ b/docker-java-shaded/pom.xml
@@ -41,8 +41,68 @@
             <groupId>com.github.docker-java</groupId>
             <artifactId>docker-java</artifactId>
             <version>${docker-java.version}</version>
+            <optional>true</optional>
         </dependency>
-
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
+            <version>2.1.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>1.7.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>1.51</version>
+        </dependency>
+        <dependency>
+            <groupId>de.gesellix</groupId>
+            <artifactId>unix-socket-factory</artifactId>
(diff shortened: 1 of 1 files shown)


KostyaSha:
:-1: will be not required because i almost resolved https://github.com/docker-java/docker-java/issues/340
Though i don't care about this plugin anymore and you can freely ignore my comments as usual

KostyaSha:
Plus to demonstrate shading, shaded guava imports need to be replaced with normal that will be direct plugin deps. Atm plugin doing hack and using shaded guava from shaded module.

magnayn:
Is the issue the one that I've seen (mvn 3.3.3) where sometimes the hpi doesn't build correctly (which seemed to work properly if I built only the hpi) ?

I assume it's at the expense of manually having to keep the dependencies in sync?

stephenc:
Whether you want to bother with this is up to you. The main problem is that in Maven 3.3.x the shade plugin can not currently replace the project model with a reduced dependency model, so you need to ensure the effective transitive model is such that you don't need to produce a dependency reduced Pom.

We will likely fix this in maven once we get a chance (but given how long I'm waiting for 3.3.6 to get released, holding breath is not advised)

There are three solutions for this project:

* this PR
* add an enforcer rule to fail the building using maven [3.3.0,3.3.6) and provide an explanation why you cannot build... Be prepared to grow the range if we don't fix it in 3.3.7
* build the shaded jar in a separate reactor (which bypasses the bug completely)

I cannot see another solution

KS has muttered that he nearly has upstream producing a shaded jar in which case you could just depend on that and get back to a single module

magnayn:
Ok. Thanks for taking the time to explain.

I think the 4th solution is 'build the shaded artifact, then build the hpi'
as 2 separate mvn invocations, which is what I tend to do in dev.

I'm minded to add for now, then once the upstream releases a shaded
artifact use that, though it shades to a different destination package. For
guava use we're currently using the target shaded package names, because
shade can't operate in the hpi itself, it seems. It'd be nice to be able to
just shade that, and not have to live with being pinned to effectively
whatever guava version docker-java declares but one thing at a time, and at
least it's a more modern version than jenkins exports.

KostyaSha:
@stephenc may `promoteTransitiveDependencies` option be related to behaviour?

stephenc:
@KostyaSha yes it is exactly related to this behaviour because it replaces the project model during the build and Maven 3.3.x made the project model immutable from plugins (a mistake in my view, the model should be allowed to be mutated but only for a limit set of mutations, where the transitive dependency tree of the replacement is a strict subset of the transitive dependency tree of the original model)

Thus you cannot build a reactor project where the shaded module is consumed within the same reactor using Maven 3.3.x

stephenc:
[MNG-5899](https://issues.apache.org/jira/browse/MNG-5899)

Issue #206, opened 2015-04-17
Running Jenkins on Docker and using docker-plugin


fcoelho:
I'm running Jenkins on Docker using [the official image](https://registry.hub.docker.com/_/jenkins/) from the Docker Hub, and I'm trying to use the `docker-plugin` to connect to containers running on the same Docker host as the Jenkins server.

Since the `docker-plugin` can't yet connect to a unix socket, I'm running `socat` on the Jenkins container to proxy a local TCP socket to the volume-mounted `/var/run/docker.sock`. That means the Jenkins server connects to `localhost:4243`, but the real Docker host is on, say, `10.0.0.5`.

When a new container is created, port 22 is exposed to `0.0.0.0:48349` (some random port), to which Jenkins then tries to connect. But the problem is that Jenkins is trying to connect to `localhost:48349`, and obviously failing since `localhost` for the Jenkins server is a different network than `localhost` for the Docker server.

I'd like to achieve two things, if possible:
- Have Jenkins create the containers but _not_ expose any ports outside of the container
- Have Jenkins connect to the container's IP directly, not through `localhost` or other settings, i.e., I want Jenkins to connect to (for example) `172.17.0.20:22` (which is why I don't want it to expose any ports - there's simply no need for it in my case)

Is this possible with any current version of the docker-plugin? I tried fiddling with a few settings but it doesn't seem like it is possible to ignore [the exposed port](https://github.com/jenkinsci/docker-plugin/blob/master/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java#L296) setting and https://github.com/jenkinsci/docker-plugin/pull/200 deals with a similar though different issue.

KostyaSha:
Next time please provide installed versions https://wiki.jenkins-ci.org/display/JENKINS/How+to+report+an+issue#Howtoreportanissue-WhatinformationtoprovideforEnvironmentandDescription
also this should be fixed in master, feel free to try this hpi https://jenkins.ci.cloudbees.com/job/plugins/job/docker-plugin/226/com.nirima$docker-plugin/

fcoelho:
Sorry about that.

Jenkins 1.596.2
Docker-plugin: tried building from source yesterday and used the one from the build server today
Docker Server version: 1.4.1
Docker Server API version: 1.16

The hpi you linked to doesn't fix the issue, it is still trying to connect to `localhost` and exposing port 22 to 0.0.0.0.

Jenkins logs:
```
[...]

PR #344, opened 2015-10-28
Allow to use unix sockets as URL (#206, JENKINS-23301)

ermshiperete:
- allow to directly connect to the container without going through
  the mapped port on the host

This is useful if Jenkins itself runs as a container with the host's
Docker socket mapped[1]. To use the docker plugin in this case simply
set the Docker URL to "unix:///var/run/docker.sock".

[1] https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/344)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java
@@ -31,9 +31,11 @@ public class DockerBuildVariableContributor extends BuildVariableContributor {
             variables.put("JENKINS_CLOUD_ID", dockerComputer.getCloudId());
             try {
                 //replace http:// and https:// from docker-java to tcp://
-                final String dockerHost = new URIBuilder(dockerComputer.getCloud().serverUrl)
-                        .setScheme("tcp")
-                        .toString();
+                final URIBuilder uriBuilder = new URIBuilder(dockerComputer.getCloud().serverUrl);
+                if (uriBuilder.getScheme() != "unix") {
+                    uriBuilder.setScheme("tcp");
+                }
+                final String dockerHost = uriBuilder.toString();
                 variables.put("DOCKER_HOST", dockerHost);
             } catch (URISyntaxException e) {
                 LOG.error("Can't build 'DOCKER_HOST' var: {}", e.getMessage());
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ work
 *.iml
 /target
 .*
+*~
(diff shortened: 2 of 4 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java
```diff
@@ -31,9 +31,11 @@ public void buildVariablesFor(AbstractBuild build, Map<String, String> variables
             variables.put("JENKINS_CLOUD_ID", dockerComputer.getCloudId());
             try {
                 //replace http:// and https:// from docker-java to tcp://
-                final String dockerHost = new URIBuilder(dockerComputer.getCloud().serverUrl)
-                        .setScheme("tcp")
-                        .toString();
+                final URIBuilder uriBuilder = new URIBuilder(dockerComputer.getCloud().serverUrl);
+                if (uriBuilder.getScheme() == "http" || uriBuilder.getScheme() == "https") {
```
KostyaSha: if unix, then not replace. Vice versa logic with less code

ermshiperete:
Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

<sup>**[docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java, line 35 \[r1\]](https://reviewable.io:443/reviews/jenkinsci/docker-plugin/344#-K1i31YEEiAnOa8k5gR5-r1-35)** ([raw file](https://github.com/jenkinsci/docker-plugin/blob/9b0f231a3fb4bd97e0efc532230f01bbdf76a202/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java#L35)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/jenkinsci/docker-plugin/344)

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerBuildVariableContributor.java
```diff
@@ -31,9 +31,11 @@ public void buildVariablesFor(AbstractBuild build, Map<String, String> variables
             variables.put("JENKINS_CLOUD_ID", dockerComputer.getCloudId());
             try {
                 //replace http:// and https:// from docker-java to tcp://
-                final String dockerHost = new URIBuilder(dockerComputer.getCloud().serverUrl)
-                        .setScheme("tcp")
-                        .toString();
+                final URIBuilder uriBuilder = new URIBuilder(dockerComputer.getCloud().serverUrl);
+                if (uriBuilder.getScheme() != "unix") {
```
KostyaSha: @magnayn that what i call "your quality"

daniel-beck: This is Java, not Groovy.

ermshiperete: I'm new to both Java and Groovy - so tell me, how would it properly look in Java then?

magnayn: Equals()

Worry not though, I already changed it when testing locally and will apply
when next sitting at a pc.

thomassuckow: In Java == is a reference comparison so depending on how the string was created you could have ```uriBuilder.getScheme() == "unix"``` be false even if ```uriBuilder.getScheme().equals("unix")``` is true.

Then if you code in both Scala and Java, you get to be really confused...

PR #345, opened 2015-10-29
Should not hide root cause of the exception

alecharp:
When testing the connection to the docker host, it might be a good thing to see the underlying exception rather that just having the `shaded.org.apache.http.client.ClientProtocolException` (example).

@reviewbybees

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/345)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -590,7 +590,7 @@ public class DockerCloud extends Cloud {
 
                 return FormValidation.ok("Version = " + verResult.getVersion());
             } catch (Exception e) {
-                return FormValidation.error(e.getMessage());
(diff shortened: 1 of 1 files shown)


KostyaSha:
> shaded.org.apache.http.client.ClientProtocolException

It's not a message, it class name. Never saw problems with it. Fix place that throw exception without description.

alecharp:
@KostyaSha I do agree that is a class name, but currently, only the class name is used in the `FormValidation` output. Using the `FormValidation#error(Throwable e, String message)` method allow to see the class name, but also to have more details is the user wants to, as there is a _show details_ link.

In my case, I did put a wrong URL in the docker server API field and the root exception was embedded in 2 others. That is why I thought it might be a good idea to have this.

oleg-nenashev:
The change makes sense IMO

alecharp:
@reviewbybees done

KostyaSha:
:+1:

Issue #350, opened 2015-11-09
Connection timeout weirdness 


jwhitcraft:
When trying to spin up an instance, it says it can't connect to my cloud host, but when i press the test connection button in the docker cloud, it connects just fine and shows me the correct version of docker, 1.8.2, i'm using v0.15 and v0.14 of the plugin

I'm also jumping subnets, the 10.27.45 is out in AWS land where as my jenkins master is on my local machine.

Here is the log.

```
Asked to provision 1 slave(s) for: my-docker-cloud-10.27
Nov 09, 2015 9:20:15 PM INFO com.nirima.jenkins.plugins.docker.DockerCloud provision
Will provision 'registry.mycompany.net/my-jenkins-image:latest', for label: 'my-docker-cloud-10.27', in cloud: 'docker-cloud-10.27'
Nov 09, 2015 9:20:16 PM WARNING com.nirima.jenkins.plugins.docker.DockerCloud provision
Bad template 'registry.mycompany.net/refinery-git:latest' in cloud 'my-docker-cloud-10.27': 'shaded.org.apache.http.conn.ConnectTimeoutException: Connect to 10.27.45.26:4243 [/10.27.45.26] failed: connect timed out'. Trying next template...
```

pfalcon:
We experience this after upgrade from 0.12.1 to 0.15 (yep, we didn't upgrade for a while because the plugin is broken since 0.12.1 - if not one thing, then another).

Example of our log is:

~~~~
INFO: Will provision 'linaro/jessie-arm64', for label: 'docker-jessie-arm64', in cloud: 'aarch64-04' Nov 09, 2015 10:13:17 PM com.nirima.jenkins.plugins.docker.DockerCloud provision WARNING: Bad template 'linaro/jessie-arm64' in cloud 'aarch64-04': 'shaded.org.apache.http.conn.ConnectTimeoutException: Connect to 10.10.0.219:2375 [/10.10.0.219] failed: connect timed out'. Trying next template...
~~~~

It's not that a press "Test Connection", I directly telnet to 10.10.0.219, port 2375 and have HTTP speaking to me. So, "connect timed out" error is completely nonsensical, what can be wrong there is that plugin does something weird. And it's not possible to diagnose it, because the real exception is swallowed, and instead out-of-context message is printed: https://github.com/jenkinsci/docker-plugin/blob/master/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java#L227

[...]

PR #352, opened 2015-11-11
DockerCloud: Accompany "Bad template" message with actual stacktrace.

pfalcon:
That message is confusing, and oftentimes, wrong - the problem may be not
with Docker template, but with something else, and it may be not even on
Docker side, but on plugin side. So, just let people to see what really
happens and where, to have better chance to diagnose it.

Implementation uses SLF4J feature available since 1.6:
http://www.slf4j.org/faq.html#paramException (Jenkins uses 1.7)

Signed-off-by: Paul Sokolovsky <paul.sokolovsky@linaro.org>

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/352)


pfalcon:
See https://github.com/jenkinsci/docker-plugin/issues/350#issuecomment-155816836 for a case which sparked this.

pfalcon:
Thanks!

PR #367, opened 2016-01-12
[fix] running JNLP with user jenkins breaks the init command with a q…

bamthomas:
…uote (see docker logs on the container)

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/367)


bamthomas:
well RUN_CMD could be inlined...

PR #373, opened 2016-02-05
Added Node Properties to template configuration.

kmbulebu:
Enables setting node properties, such as tool locations, on a per docker image basis.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/373)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -9,7 +9,10 @@ import hudson.model.Descriptor;
 import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.labels.LabelAtom;
+import hudson.slaves.NodeProperty;
+import hudson.slaves.NodePropertyDescriptor;
 import hudson.slaves.RetentionStrategy;
+import hudson.util.DescribableList;
 import hudson.util.FormValidation;
 import jenkins.model.Jenkins;
 import org.kohsuke.accmod.Restricted;
@@ -21,6 +24,10 @@ import shaded.com.google.common.base.MoreObjects;
 import shaded.com.google.common.base.Strings;
 
 import javax.annotation.CheckForNull;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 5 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -224,6 +244,15 @@ public DockerImagePullStrategy getPullStrategy() {
     public void setPullStrategy(DockerImagePullStrategy pullStrategy) {
         this.pullStrategy = pullStrategy;
     }
+    
+    public List<? extends NodeProperty<?>> getNodeProperties() {
```
kmbulebu: This return type is meant to match the constructor parameter type of AbstractCloudSlave. Removing the usage of the generic wildcard type here would require a cast at the super() call in DockerSlave. 

See [AbstractCloudSlave](http://javadoc.jenkins-ci.org/hudson/slaves/AbstractCloudSlave.html#AbstractCloudSlave(java.lang.String, java.lang.String, java.lang.String, int, hudson.model.Node.Mode, java.lang.String, hudson.slaves.ComputerLauncher, hudson.slaves.RetentionStrategy, java.util.List)): 
`public AbstractCloudSlave(String name,
                  String nodeDescription,
                  String remoteFS,
                  int numExecutors,
                  Node.Mode mode,
                  String labelString,
                  ComputerLauncher launcher,
                  RetentionStrategy retentionStrategy,
                  List<? extends NodeProperty<?>> nodeProperties)
                   throws Descriptor.FormException,
                          IOException
`

kmbulebu:
I can split up readResolve, but I think it'll hurt the readability.

KostyaSha:
You can ignore bot messages, it was my work on enhancements, but i left this plugin development. @lanwen please disable your bot.

lanwen:
disabled. But think `readResolve` now can be splitted for 3 methods - `config1`, `config2`, `other` - so in `readResolve` will be only `if` with version selector. Big methods with lot of complexity never will be better than a couple of small and robust methods which doing one thing.

kmbulebu:
@lanwen I pulled out the config defaults from readResolve() into a configDefaults() to reduce the complexity a bit, but I stopped short of splitting it into many methods. Hope that's okay with you.

lanwen:
@kmbulebu thats much better, thanks!

ndeloof:
duplicates https://github.com/jenkinsci/docker-plugin/pull/502

PR #377, opened 2016-02-09
Support Joyent Triton and Single Host Abstractions

### LINKED ISSUE #376: Support Joyent Triton / support single host abstraction when using ssh
Joyent, VMWare (project Bonneville) and Microsoft all have Docker management implementations that are based on modeling an entire cloud as a single host. This confuses the Jenkin's Docker plugin because when it is attempting to find the host to ssh to, it assumes it is the same host as the `DOCKER_HOST` value. This is not always true.

In particular, this block in `DockerComputerSSHLauncher.java`: 
```java
protected InetSocketAddress getAddressForSSHD(String cloudId, InspectContainerRe
     //get address, if docker on localhost, then use local?
     if (host == null || host.equals("0.0.0.0")) {
        // *** This line defaults to using the DOCKER_HOST value, but that is incorrect when using
        // a system that supports a single host abstraction. ***
        host = URI.create(DockerCloud.getCloudByName(cloudId).serverUrl).getHost();
        if (host == null || host.equals("0.0.0.0")) {
             host = networkSettings.getIpAddress();
             port = sshConnector.port;
        }
     }
}
```

dekobon:
Fixes #376

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/377)


magnayn:
Cool - we're likely to start using Triton here. Ta!

PR #378, opened 2016-02-11
Allow container to get access to specific devices

SergiyMakarenko:
This is my initial attempt (still messy) to pass devices inside container.
I tried to follow code for other settings (eg 'Volumes') as an example but it looks like I'm missing something.
Although I can see 'Devices' edit box in plugin configuration section actual settings are not persisted.
Comments or direction how to fix it would be appreciated.
Thanks!

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/378)


ndeloof:
I'm not sure this is desirable to expose actual devices, and is certainly not be supported in swarm mode, is it ?

PR #383, opened 2016-03-09
JENKINS-33426: Add support for configuraion of templates on project l…

irissmann:
We like to configure the templates on project level instead in global configuration. See also
https://issues.jenkins-ci.org/browse/JENKINS-33426

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/listener/DockerQueueListener.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/listener/DockerQueueListener.java
@@ -1,0 +1,64 @@
+package com.nirima.jenkins.plugins.docker.listener;
+
+import hudson.Extension;
+import hudson.model.Project;
+import hudson.model.Queue.Item;
+import hudson.model.Queue.LeftItem;
+import hudson.model.Queue.WaitingItem;
+import hudson.model.queue.QueueListener;
+import hudson.slaves.Cloud;
+
+import com.nirima.jenkins.plugins.docker.DockerCloud;
+import com.nirima.jenkins.plugins.docker.DockerJobProperty;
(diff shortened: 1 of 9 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/listener/DockerQueueListener.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void onLeft(LeftItem li) {
+        DockerJobTemplateProperty jobTemplate = getJobTemplate(li);
+        if (jobTemplate != null) {
+            Cloud cloud = DockerCloud.getCloudByName(jobTemplate.getCloudname());
+            if ((cloud != null) && (cloud instanceof DockerCloud)) {
```
KostyaSha: instanceof is null check itself.

KostyaSha:
Yes, that may work. :+1:

KostyaSha:
The other possible way that i thought to try is doing provisioning from QTD with bypassing cloud provisioner.

irissmann:
Yes, it works :) But sometimes we have to wait quite a while before the background process picks up the the job from the queue.
Thanks for your hint with the QTD. I will try...

KostyaSha:
@irissmann i hope solve provisioning delays soon if i would have free time. (Excluding trolling from ndeloof) you may get something useful from this thread https://groups.google.com/d/msg/jenkinsci-dev/WV6sPmIh-tk/G-WfwFIpFQAJ

irissmann:
I changed the commit and removed the null check like you recommended.
I also checked out the QTD but I prefer the QueueListener because it's triggered before the QTD and I don't want bypass the cloud provisioner :)

Issue #246, opened 2015-06-10
Workspace mapping doesnt work if thera are builds wich ran not on docker


tomaxsas:
If i change my job to use docker, workspace mapping doesn't work until a manually delete all builds which ran on ordinary slaves, or let them pass along with log rotation. I could delete builds, but i want to have some history for a month or similar. I ran few test and as build wich ran on master or another slave is gone, docker maps workspace fine.

ndeloof:
This feature would benefit a full re-implementation. For sure volumes API didn't existed when plugin has been created, but today a volume to host workspace would be the best way to achieve this, and one can create a transient container to access it (readonly) and browse content without need for some mapping on master - which anyway sounds weird, need to double check but I don't think this feature can work with a remote build agent.

ndeloof:
Confirmed this "feature" just doesn't work, and can't work with a non-local docker daemon (not even to consider a swarm cluster).
I will deprecate this and search for an alternative implementation relying on volumes

PR #399, opened 2016-06-02
Add caveat about workspace mapping

wsoula:
This change adds some helper text to the remote workspace mapping
configuration option that would have saved me a lot of time.  This adds
documentation about #246


fboudra:
typo: contianers -> containers

wsoula:
@fboudra updated, and reworded part of it

PR #408, opened 2016-07-30
Improve documentation on Container Cap and Instance Capacity

eaglerainbow:
Based on the discussion at https://issues.jenkins-ci.org/browse/JENKINS-36919 and https://issues.jenkins-ci.org/browse/JENKINS-36920, I was unsatisfied with the documentation provided for the parameters Container Cap and Instance Capacity:
* The help page of Container Cap is not reachable for the end user
* Both help pages do not make clear what is being counted.
Both issues are fixed with this change.


eaglerainbow:
failed check is not related to this change

PR #409, opened 2016-07-30
JENKINS-36919: only count containers provisioned by plugin itself

eaglerainbow:
Other Docker clients may create containers on the docker host of our
Jenkins using the same image as Jenkins will do. Before, these
containers were considered when counting both for the Instance Capacity
and the Container Cap. Especially in SWARM-based situation, where
multiple Jenkins instances are using the same docker host, this may lead
to unwanted side-effects.
The commit introduces a synchronized set, which "remembers" the
containerIds, which were provisioned by this Jenkins instance. When
determining the number of running containers (Container Cap) and/or the
number of containers of the selected image (Instance Capacity), the set
is used to only count those containers, which originate from this
Jenkins instance.

For details of the issue, see also https://issues.jenkins-ci.org/browse/JENKINS-36919

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ work
 /target
 .*
 *~
+
+dependency-reduced-pom.xml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Changelog
 
-## 0.17
+## 0.16.1,2
 
-* Move to docker-java 3.x
-* Re-instate setting API version
+* Move to docker-java 3.x (3.0.6 + fixes for SDC/Triton)
+* Re-instate setting API version as some versions of docker break compatibility
 * Allow setting of registry credentials (build, push, pull)
+* Documentation clarifications
 
 ## 0.16.0
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,5 +17,5 @@
 
 # Links
 
-- https://wiki.jenkins-ci.org/display/JENKINS/contributing
-- https://wiki.jenkins-ci.org/display/JENKINS/Extend+Jenkins 
+- https://wiki.jenkins-ci.org/display/JENKINS/Beginners+Guide+to+Contributing
+- https://wiki.jenkins-ci.org/display/JENKINS/Extend+Jenkins
--- a/docker-java-shaded/pom.xml
+++ b/docker-java-shaded/pom.xml
@@ -33,7 +33,7 @@
              submodule. But for now it's probably better to allow CI to work
             as it will at least build on the CI. -->
 
-        <docker-java.version>3.0.6-jenkins</docker-java.version>
+        <docker-java.version>3.0.6</docker-java.version>
 
         <jersey.version>2.23.1</jersey.version>
         <jackson-jaxrs.version>2.6.4</jackson-jaxrs.version>
--- a/docker-java-shaded/pom.xml
+++ b/docker-java-shaded/pom.xml
@@ -6,16 +6,21 @@
      transient jenkins dependencies. -->
     <groupId>com.nirima</groupId>
     <artifactId>docker-java-shaded</artifactId>
-    <version>0.16.2-SNAPSHOT</version>
+    <version>0.16.3-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>Docker-java shaded jar for jenkins plugin</name>
 
     <distributionManagement>
         <repository>
+            <uniqueVersion>false</uniqueVersion>
             <id>maven.jenkins-ci.org</id>
-            <url>http://maven.jenkins-ci.org:8081/content/repositories/releases/</url>
+            <url>https://repo.jenkins-ci.org/releases</url>
         </repository>
+        <snapshotRepository>
+            <id>maven.jenkins-ci.org</id>
+            <url>https://repo.jenkins-ci.org/snapshots/</url>
+        </snapshotRepository>
     </distributionManagement>
 
     <properties>
@@ -28,7 +33,7 @@
              submodule. But for now it's probably better to allow CI to work
             as it will at least build on the CI. -->
 
-        <docker-java.version>3.0.2</docker-java.version>
+        <docker-java.version>3.0.6-jenkins</docker-java.version>
 
         <jersey.version>2.23.1</jersey.version>
         <jackson-jaxrs.version>2.6.4</jackson-jaxrs.version>
@@ -107,7 +112,7 @@
             <artifactId>jersey-client</artifactId>
             <version>${jersey.version}</version>
         </dependency>
-        <dependency>
+       <!-- <dependency>
             <groupId>de.gesellix</groupId>
             <artifactId>unix-socket-factory</artifactId>
             <version>${unix-socket-factory.version}</version>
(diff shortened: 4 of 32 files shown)


eaglerainbow:
failed checks are not related to this change

magnayn:
One thing I wondered was if it were possible to add metadata to the container at creation time (e.g a jenkins instance ID).

I know latter docker APIs support this. I haven't looked too closely at the detail of java support though. It may be more efficient to search on that than having to remember the list..

magnayn:
e.g https://docs.docker.com/engine/userguide/labels-custom-metadata/

eaglerainbow:
@magnayn: docker supports this in general. However, given the state of "maintenance only" of this plugin, this would be a much more invasive correction. That is why I abanonded this idea quite early during my implementation. 
However, I could imagine that for your successor-folk, this could be a valid approach (because then this piece of information would also be available to the outside world)

magnayn:
Well, it's not maintenance-only - but there's a dearth of actual code contributions vs the endless wishlist for free implementation!

I'm using SDC/Triton, which is swarm-like; it doesn't have this issue as it only returns containers/images tied to a particular set of credentials (so jenkins instance has its own). I'd imagine it's using a similar metadata approach.

I'd imagine (modulo docker-java support) it'd be relatively easy to add a tag and latterly filter on it.

eaglerainbow:
Then we should clarify the approach vice versa: Does DC/Trition support adding metadata labels like Docker does? If yes, I would have a second look at it. If no, we would only have to set the labels in case of a non-Triton environment, which would make the implementation more complex :-|

magnayn:
In general, SDC ought to be API compatible with docker (though in practice
it is deficient as it can't do 'docker build').

However, we already detect if the cloud is SDC because of that, so I can
turn off the filtering if you're actually on SDC.

I'm having a quick look - I suspect this may not be that hard.

There is a slight question in my mind over the container cap. Current
behaviour will cap it to the total number of containers on the host
(regardless if started by us). If we filter out non-started-by Jenkins,
that behaviour changes (and may not be what everyone wants..)

eaglerainbow:
> There is a slight question in my mind over the container cap. Current behaviour will cap it to the total number of containers on the host (regardless if started by us). If we filter out non-started-by Jenkins, that behaviour changes (and may not be what everyone wants..)

I see your point. However, that is exactly the issue I have in my environment: The cluster is also "used by some other folks", and the load they produce is entirely different from what I am producing. Essently, the problem also is that there may be multiple Jenkins instances around and their Container Cap need to be in sync. Otherwise, the server with the lowest Container Capp will be the one which is suffering the most. All in all, I am not very satisfied with the Container Cap thing in general -- I would have expected that this kind of feature was available already on docker side (it's not reasonable to assume that there "inifinity capacity" -- neither on a single docker host and even not on a docker swarm cluster).

What about the idea to make it a configurable option how to interprete the Container Cap? We could then default that one to "use old behaviour - consider all containers on the server" to keep compatibility. Shouldn't be that complicated, I think...

magnayn:
SO
https://github.com/jenkinsci/docker-plugin/commit/459a88f03fb09a9121e066a4e4f897a9cfefcd75
adds a JenkinsID that can be used to filter.

The cloud API isn't really so useful if (like I did) you have a pool of
machines that you want to fill up, as it really just asks each in turn "can
you provision a <x>" and not "where would be the best location to provision
an <x>", so it's down to providing enough features to get the best answer
to that.

I don't know if docker-swarm provides this kind of behaviour - certainly
SDC/Triton does (and has a lot of tweakable heuristics to control where to
provision next).

eaglerainbow:
I could verify that your commit works with docker (even native / without swarm) and adds a corresponding label to the contianer: 
`docker ps --format "{{.ID}} {{.Image}} {{.Labels}}"`
`2dbf93baf7e0 jenkins-1 JenkinsId=5213d6f01ba52f7958cb90a6340c9caf`

However, I noticed that configuration compatibility is broken (different story) though.
I will have a look if I can achieve the same thing with using labels during querying.

eaglerainbow:
I have reworked the pull request: It now makes use of the newly introduced label of the Jenkins instance by filtering the list command request. I tested it against my local docker instance (Server Version: 1.11.2), by running the same image as Jenkins is using manually (thus not providing any label), whilst triggering two jobs on the jenkins instance. The count of running containers war properly returned as expected.

This only leaves the point open on

> Current behaviour will cap it to the total number of containers on the host

which I would tackle next (but will take some more days to complete).

eaglerainbow:
Still managed to get something coded -- but the stuff still needs some further testing...
However, early feedback is already welcome.

eaglerainbow:
Ok, this looks better now. 
The current implementation adds another configuration parameter called "Local Container Cap" (in contrast to the already existing "(Global) Container Cap"), which provides a Container Cap only counting those instances, which the local Jenkins Id has provisioned before. By default, the new limit is disabled. If you want, you may also enable both limits at the time (something which I wouldn't see a scenario for, though). Help documentation is included as well.
Feedback is appreciated.

eaglerainbow:
Given the discussion in #425, we should verify my statement from above:

> By default, the new limit is disabled.

In #425 we learned that "disabling limits" may be significantly trickier that we thought...

eaglerainbow:
Build error is related to the fact that apparently the pom.xml of docker-plugin-shaded is not properly bumped to 3.0.6.
At least http://repo.jenkins-ci.org/public/com/github/docker-java/docker-java/3.0.6/ is available, whilst still the coding refers to 3.0.6-SNAPSHOT
Error message:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project 
docker-java-shaded: Could not resolve dependencies for project com.nirima:docker-java-shaded:jar:0.16.3-SNAPSHOT: 
Could not find artifact com.github.docker-java:docker-java:jar:3.0.6-jenkins in repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/)
```

h0nIg:
whats the status of this? will this get merged, since PR signals closed (" eaglerainbow closed this 8 days ago") instead of merged

berni2288:
I also wondered what's going on with this, this is a real issue for us.

eaglerainbow:
closing PR was done by mistake - reopening it

eaglerainbow:
@magnayn I just found another case where even for the case of Triton you might run into the issue: 
* Let us have two separate Jenkins instances, which are referring to the same Triton host.
* Let the two instances share the same credentials.
* Let the two instances use the same image name.
Then, also in the Triton scenario, the instance with the lower (Global) Container Cap will starve...

eaglerainbow:
I have now changes this PR to show the following behaviour:
* Zero for both Global and Local Container Cap means "Do not provisioning any container"
* Default value of Local Container Cap is set to 2147483647

This makes it
a) compatible
and 
b) the two types of Container Caps consistent with each other.

Moreover, I have tried to enrich the online documentation to help with the possibly confused users...

eaglerainbow:
After some further testing, IMO this PR is ripe for being merged.

ndeloof:
lgtm from technical point of view but I'm reluctant from a feature PoV. Limiting resources consumption should not be driven by # of containers but by the amount of resources they allocate, i.e I could have 200 containers with 64Mb allocated memory, but only few of them to run with 4Gb

PR #410, opened 2016-07-31
Improve documentation on Retention Strategy/Availability drop down

eaglerainbow:
The field for defining the retention strategy for a DockerTemplate
refers to the generic availability help. However, the values described
there only do have little in common with the values which the end user
can select. 
This commit provides a rough summary about the retention strategies and
tries to describe their behavior in a nutshell to the end user. The text
is newly associated to the field "Availability" in the configuration
dialog.


KostyaSha:
@eaglerainbow can i pick your description into https://github.com/KostyaSha/yet-another-docker-plugin ?

eaglerainbow:
sure, feel free to!

Issue #415, opened 2016-08-12
NullPointerException in DockerCloud.pullImage() (Line 328)


fraz3alpha:
When adding a new Template to the Docker Cloud any attempt to try and instantiate a copy of it would fail and a NullPointerException was produced in the log:

```
java.lang.NullPointerException
        at java.util.Arrays$ArrayList.<init>(Arrays.java:2853)
        at java.util.Arrays.asList(Arrays.java:2839)
        at com.nirima.jenkins.plugins.docker.DockerCloud$2.apply(DockerCloud.java:331)
        at com.nirima.jenkins.plugins.docker.DockerCloud$2.apply(DockerCloud.java:328)
        at shaded.com.google.common.collect.Iterators.indexOf(Iterators.java:778)
        at shaded.com.google.common.collect.Iterators.any(Iterators.java:684)
        at shaded.com.google.common.collect.Iterables.any(Iterables.java:623)
        at com.nirima.jenkins.plugins.docker.DockerCloud.pullImage(DockerCloud.java:328)
        at com.nirima.jenkins.plugins.docker.DockerCloud.provisionWithWait(DockerCloud.java:363)
        ...
```

Docker plugin version: 0.16.1 (also 0.16.0)

Docker slave version:
```
Server Version: 1.11.2
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 294
 Dirperm1 Supported: true
Logging Driver: json-file
Plugins: 
 Volume: local
 Network: bridge null host
Kernel Version: 4.4.0-28-generic
Operating System: Ubuntu 16.04 LTS
OSType: linux
Architecture: x86_64
CPUs: 32
Total Memory: 62.73 GiB
```

The symptom was identical to Issue #315 , but in that report the raiser managed to work around his problem by changing the API version and using a different Jenkins. In this case the problem persisted if a different cloud template or Jenkins was used - as long as the image was being targeted at the same Docker host.

Looking at the area around the problem here is the snippet:

```
    private void pullImage(DockerTemplate dockerTemplate)  throws IOException {
        final String imageName = dockerTemplate.getDockerTemplateBase().getImage();

        List<Image> images = getClient().listImagesCmd().exec();

        NameParser.ReposTag repostag = NameParser.parseRepositoryTag(imageName);
        // if image was specified without tag, then treat as latest
[...]

Issue #414, opened 2016-08-12
What is the intended build process given your use of a private fork of docker-java?


fraz3alpha:
Since commit ac2797a3966aac90152b0d78a75bf04c60469ec1 and the use of version `2.2.4-jenkins` of the docker-java dependency the plugin will not build as-is.

What is the intended build process now that you are using a non-standard dependency -do we have to build that library ourselves first and ensure it is in our local maven cache?

KostyaSha:
That how @magnayn maintain, he solved his problem by using custom patched dependency. His "custom" change is in docker-java PR but he can't provide requested information. 
If you want to build custom hpi you can freely switch to mainstream docker-java version as i sure that you don't need his custom fix.
Have a nice day!

magnayn:
Well, the process was going to be that I submitted the fix to docker-java and it wouldn't be the need.

Sadly I did, but was met with the response that "2.x branch is no longer supported" (despite the fact that I was - literally - supporting it by submitting a PR (!) 

Moving to 3.x required a bunch of, somewhat arbitary IMO, API changes that I didn't have time for + figuring out how to fix the same thing in docker-java 3.x (this free software is justified and funded by it's use in my $dayjob).

We now _are_ on the 3.x branch, as work needed some of the auth functionality.

I've once again submit PRs to cover the missing / broken parts; I'd clearly like to be on mainline. If they don't get merged, I'll add a fork as a submodule to keep things in one place.

PR #412, opened 2016-08-09
Upgrade to docker-java 3.x

bschelberg:
Docker-java 3.x allows the use of the ':Z' option for volumes. I don't know what's in your private fork (for wildcard SSL?), so this is based off the public release.

I haven't created an issue for this, what I wanted to do was specify this in a Docker template:

    Volumes: /tmp/fortify/:/home/jenkins/.fortify/:Z

Version 2.x of docker-java threw a parse error with volumes configured like this. Version 3.x supports it.

PR #416, opened 2016-08-12
Guard against null image tags during pull logic

fraz3alpha:
For issue #415 

  - In the case where there happens to be a badly stored image on the
    target host it is possible that the image.getRepoTags() call will
    return null. This causes an exception to be thrown during processing
    which can only be recovered by cleaning up the Docker host. In the
    case where you either use a host/swarm managed by someone else, or
    share the resources with other teams, you cannot always guarantee
    all hosts will be clean, so it is important to include this check

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -334,7 +334,11 @@ public class DockerCloud extends Cloud {
         boolean imageExists = Iterables.any(images, new Predicate<Image>() {
             @Override
             public boolean apply(Image image) {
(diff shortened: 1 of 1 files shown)


fraz3alpha:
Presumably unintended consequence of using a private docker-java library, the automatic PR builder doesn't work :(

```
Caused by: org.apache.maven.project.DependencyResolutionException: Could not resolve dependencies for project com.nirima:docker-java-shaded:jar:0.16.2-SNAPSHOT: Could not find artifact com.github.docker-java:docker-java:jar:2.2.4-jenkins in repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/)
```

Also, see issue #414 , and PR #412 for the kinds of problems that this is also causing

KostyaSha:
Yes, it bug.

Issue #413, opened 2016-08-10
Start/Stop Docker containers build step 'Run Container' multiple environment properties


gupbeheer:
There is no possiblity to add multiple environment properties to the "Run Container" build step. The split is done based on "\n" but there is only the option to put in one line. 
In the Template settings there is an option to "expand" the "environment" input field to a textarea. This should also be possible within the 'Start/Stop Docker container' build step 'Run Container'.

KostyaSha:
@gupbeheer why do you need this UI based start/stop commands? Could you describe your use case?

gupbeheer:
@KostyaSha we use the start/stop commands (run in our case) to run multiple containers before our integration (black box testing) tests run. This guarantees that we have a clean environment (database, server) that is exactly the same for each test run and we use the plugin to stop and remove the started containers after the tests.
We use the UI because that way we do not need an SSH server configured on Jenkins to run the commands on the Docker machine and for the "automatic" cleaning at the end of the build.

KostyaSha:
But you can use docker CLI and DOCKER_HOST is already injected into build process.

gupbeheer:
@KostyaSha To use Docker CLI the Docker CLI must be available. This is not the case on our build server. We need to SSH to the Docker machine (or another one with the Docker CLI) to execute the command, we then need to store the id's of the running containers so that we can stop and remove them later. The UI is far more simple and is offered in the plugin, so why not use it?

PR #419, opened 2016-08-15
Fixes jenkinsci/docker-plugin#413 Start/Stop Docker containers build step 'Run Container' multiple environment properties

gupbeheer:
The DockerSimpleTemplate UI is now based on the DockerTemplateBase UI as they share the same codebase. Multiple (line) entries for Volumes, VolumesFrom and Environment are now possible.

This fixes jenkinsci/docker-plugin#413


gupbeheer:
failed checks are not related to this change

PR #425, opened 2016-08-18
Container Cap = 0 blocks provisioning of new containers

eaglerainbow:
see also https://issues.jenkins-ci.org/browse/JENKINS-37526

### DIFF
--- a/docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerCloud/help-containerCap.html
+++ b/docker-plugin/src/main/resources/com/nirima/jenkins/plugins/docker/DockerCloud/help-containerCap.html
@@ -1,5 +1,6 @@
 <div>
     <p>The maximum number of <b>containers</b> that this provider is allowed to run in total (value is independent of the template/image).</p>
     <p>Note that also containers, which have <b>not</b> been created by Jenkins, are counted as well.</p>
-    <p>Defaults to blank which disables this limit.</p>
+    <p>Defaulted to 100, if not specified otherwise.</p>
+    <p>A value of zero (0) disables provisioning of containers altogether. If you want to disable this limit, enter a non-reachable high value, such as 2147483647.</p>
 </div>


magnayn:
Hm. The history of this is a bit chequered.

Originally the semantics were
blank = not specified (No limit)
0 = zero = can't provision
>0 = provision up to that limit.

"Someone" then decided that blank ought to mean integer.max, and so to set the cap to that.

The only problem I can see with making zero work as "not specified" is how then can I temporarily stop a cloud from provisioning (I.E: I sometimes really *do* want it to be 0...)

eaglerainbow:
> The only problem I can see with making zero work as "not specified" is how then can I temporarily stop a cloud from provisioning (I.E: I sometimes really do want it to be 0...)

mmmh... But that sounds to me like that you want to have checkbox with "cloud enabled/disabled" or something like this. Setting the container cap to 0 is also a solution for this case, but a very implicit one (and obviously isn't something where the user immediately understands what you want). 
So, what do you think about enhancing this commit with another configuration option/checkbox, such that the administrator can (temporarily) disable the specified DockerCloud altogether?

eaglerainbow:
Allowing to prevent provisioning containers entirely, I have introduced another checkbox in the configuration to disable the entire DockerCloud. I think with that it is possible to resolve this ambigious value of "0".
What currently gives me a little headache (but the root cause already happened in the past) is the fact that today we are not sure what a "Container Cap = 0" really should mean:

1. If it was an "old zero", it meant "do not provision" :arrow_right: `enabled = false`
2. If it was a blank before which got converted, it meant "unlimited" :arrow_right: `enabled = true` with (new) `Container Cap = 0`. 

My understanding was variant 2 (based on the documentation which is still provided), as I am a "new user". Old users who have entered zero (meaning to disable provisioning) still can live with their configuration today. With this pull request, which synchronizes the code again with the documentation, suddenly this zero would mean "unlimited" -- which is quite the contrast of what they most likely will expect.

In short, it remains unclear to me which one was right :confused:
* the coding, or
* the documentation.

Or to put it in some other words: it is more a matter of what we declare to be the bug :confounded:

KostyaSha:
> "Someone" then decided that blank ought to mean integer.max, and so to set the cap to that.

"Someone" copy-pasted logic from AbstractCloud or EC2 plugin where default logic was fallback to max integer, this end into sucking provisioning upto 32k nodes that happened unexpectedly to user. That's why i changed this logic to strict 0 - zero nodes, 1 - one. Because of backward compatibility this was done with migration. But I missed help files because they are not tied statically to code.

Designing plugins with checkboxes is way to the hell.  Blank - default magic that will follow user to unexpected behaviours.

thomassuckow:
My question: do we need unlimited?

How much trouble is it for someone to put in 99999 if they have some massive instance?

Should blank not be allowed and coerced to something like 100 as a default? Or maybe 10?

KostyaSha:
> Should blank not be allowed and coerced to something like 100 as a default? Or maybe 10?

https://github.com/KostyaSha/yet-another-docker-plugin/blob/master/yet-another-docker-plugin/src/main/java/com/github/kostyasha/yad/DockerSlaveTemplate.java#L68

magnayn:
Yes, it's sadly a bit of a mess. A change was introduced that converted
"blank" to INTEGER_MAX, which I think is pretty awful.

From a UI/UX PoV, there's 3 choices:

1) I do not care about limiting instances at all
2) I want to limit it to some nonzero number of instances
3) I want to limit it to zero (I.e disable, possibly temporarily)

In general,I prefer fewer "knobs to twiddle" - hence a single input
allowing [empty], a number, and zero seem(s, ed) (to me, anyway) the
simplest mechanism. (FWIW, I've often run in mode1 as the cloud I used had
on-demand capacity).

I guess I don't mind a 'disable' checkbox - though of course what would
'enabled, but a value of 0' mean?

I don't particularly like 'just set 99999' - as when J. Random other user
comes to our Jenkins server they puzzle over 'why this arbitary number' -
but it does of course achieve the goal.

The other thing to consider is what the default should be. I would choose
(1) - therefore if you go with 'just set 99999', the plugin ends up placing
a magic number in that field which will inevitably be too low for some, and
too high for others :-/

thomassuckow:
I don't think anyone truly can "handle" unlimited, everyone should in theory have a physical limitation or a monetary limitation. When Jenkins goes rogue (because someone triggered a matrix job in an loop) there should be some limit other than integer max.

If someone can actually run 10K jobs at once with there Jenkins install and that's "normal", I'd love a screenshot.

If the field is required, it forces their hand to do the rough mental calculation of how many containers is reasonable.

magnayn:
The cloud we use ends up failing at the docker API level anyway ("no
compute resource available"), so it's a non-issue for us. Indeed, as one of
the other threads that's engaged in jenkins-spawned-container-counting,
it's actually very difficult to come up with a "should I allow provisioning
on this cloud" answer without quite a lot of external information (e.g:
Should the limit be <n> containers? <n> Jenkins containers? <n> containers
started by a particular user?)

This perhaps should be a pluggable strategy :-/

SDC/Triton (and I assume Swarm) is doing some of this anyway when the
request comes in - hence why I (personally) do not want Jenkins enforcing a
limit there because there's no sensible way it can work it out. My *guess*
is that, over time, more people are going to end up using some sort of
clustered Docker provider (be it public or private cloud), so I actually
suspect (1) is going to end up being what most people will end up wanting
(or - if you prefer - 'this limit is enforced by some other mechanism').

I can live with a numeric default (it is, after all, only a default!), but
I do think making arbitary numbers mean "I don't care about this value"
isn't right.

KostyaSha:
> If someone can actually run 10K jobs at once with there Jenkins install and that's "normal", I'd love a screenshot.

@thomassuckow you can't run even more 200 slaves without hitting core issues, so it LOL to see such settings/defaults in this plugin.

eaglerainbow:
> This perhaps should be a pluggable strategy :-/

Before coming to the other aspects of this discussion, I'd like to briefly comment on this, although it tends to become off-topic: I am not so in favor of this idea, because the set of configuration options already provided today is quite overwhelming for an average user (at least I was gulping the first time I saw the list of options in the config - especially when I was puzzled by the help in combination with many missing details on the plugin's wiki page :grimacing: - I had to set up scenarios for hours to determine the exact behavior verifying if the plugin suits my needs...). 

That is why I also do not feel so fancy about my own idea of introducing the "enabled" checkbox. However, I could understand the value-add in that sense that disabling it while keeping the well-considered Container Cap alive, is something valuable (temporarily entering a zero to disable the provisioning destroys the original value).  And BTW it would help to sort out this ambiguity which now caused this debate...

My own scenario for instance is that my Jenkins job load is very CPU-intensive, consuming more cores than executors are running, and thus docker swarm has issues to determine the limits based on the load average. Moreover, I am on a private cloud and therefore know quite well, what my capacity is -- I even know this better on instance level. That is why the Container Cap is used here for something like a "last resort safty belt": If the number of containers on the cluster is getting larger than this value, then something is going brutaly wrong and the cluster should be protected from being unnecessarily overstressed (thus queueing via the build job queue appears to be a good idea anyway then ;) ). In that sense, in general I agree to @thomassuckow's argument that it's worthwhile to think about what at least "a reasonable upper boundary value" for the existing installation should be.
However, on the other hand, if the limitting is already implemented on the cloud's cluster infrastructure (which I would also agree to be the better design due to the additional knowledge that should be available there), it is useless (and in the sense of UX even nonsense) to introduce another limiter (especially, if the value is considered to be used as global value for the target -- which is the case here). And that's exactly where 

> you can't run even more than 200 slaves without hitting core issues, so it LOL to see such settings/defaults in this plugin.

(@KostyaSha's argument) comes into the play: The Container Cap does not necessarily mean that you are alone on the cloud/cluster! For instance, in my scenario, I have a cluster which is used by five different Jenkins instances (each running with docker-plugin). Whilst IMHO I think of the "instance limit" to be local for each Jenkins instance, the Container Cap should IMHO be an superordinated "safety belt" protecting the entire cluster (calculated across all Jenkins instances involved) -- even better, if that protection was available by the cluster software (unfortunately, I have seen the opposite with docker swarm).

eaglerainbow:
> though of course what would 'enabled, but a value of 0' mean?

Well, that was exactly my intention: a literal 0 can't mean "don't provision anything" - as this would be quite contradicting (though due to our compatibility/migration issue a combination which may happen in wildlife). That is why we could use this combination to mean "infinite capacity limit", which would be in sync with the original documentation. 
On the other side, with the statement

> But I missed help files because they are not tied statically to code.

we know from @KostyaSha that the documentation was apparently missed out. That is why the coding apparently is holding the truth (which answers my question "In short, it remains unclear to me which one was right: the coding, or the documentation?").
Given that, I would suggest that

* we must consider the existing coding to be the intention,
* the documentation needs to be fixed, stating now that "zero means zero" and not infinite, and
* my pull request should be rejected, as it will introduce a compatibility issue after upgrading (zero to suddenly mean infinite). 

At least, this would be consequent -- however, it does not fix the question, how to describe "infinite" / "I don't care to limit"...

KostyaSha:
It may also happen that total cloud limit may have bit different logic vs template limit.


> I was puzzled by the help in combination with many missing details on the plugin's wiki page 😬

Part of the issue that jenkins core sucks in UI features, in yad-plugin i covered almost everything with help files and good default configuration, but jelly in some parts still screwing UX!

eaglerainbow:
> It may also happen that total cloud limit may have bit different logic vs template limit.

Well, we have another discussion on that as well: please refer to #409. As a matter of fact, this leads to the approach that there are "Global Container Caps" and "Local Container Caps" (or alternatively "strategies" as @magnayn has proposed).

eaglerainbow:
> That is why we must assume that the coding is holding the truth

Being strict to myself, I havead justed this pull request such that only the documentation gets changed, now reflecting the current state of implementation.

Issue #401, opened 2016-06-16
Containers sometimes don't get removed after job is done


farahfa:
Hello,

From time to time, I notice that there are a lot of containers that aren't removed after the job is done, is there an explanation as to why that happens? Is it a bug or am I missing something? When I do docker rm $(docker ps -a -q) and remove the containers and then run a bunch of jobs, then it works as expected.

Fadi

HadesArchitect:
Hi,

which exactly containers aren't removed? Slaves? Can you show _docker ps_?

KostyaSha:
There are a lot of bugs in docker-plugin around provisioning :)

HadesArchitect:
For sure :)

But I clearly see a great future in this plugin. Docker popularity &, therefore, community will grow!

HadesArchitect:
I've asked for details because I don't have any troubles with containers remaining... Except my own containers which were spawned during job execution.

KostyaSha:
> Except my own containers which were spawned during job execution.

That fixable with cleanup mechanisms. 

> But I clearly see a great future in this plugin.

It had future until CB employees started rewriting their own plugins with the same functionality (even i was forced to rewrote it).

> From time to time, I notice that there are a lot of containers that aren't removed after the job is done, is there an explanation as to why that happens? Is it a bug or am I missing something?

That can be everything. When i was maintainer here i placed https://github.com/jenkinsci/docker-plugin/blob/master/CONTRIBUTING.md , but nobody providing well formed issues with use cases and logs. If i right remember it should be not required to do manual clean-ups because in code i made automatic cleanups after timeout (or in other plugin... but should be in this, i remember that deal with such cases).

farahfa:
@HadesArchitect, Sorry for not replying sooner. But yes I was referring to Slaves still showing up when I do docker ps -a (they wouldn't be running but they're not completely removed which ends up taking space). But this happens randomly and I cannot reproduce the problem anytime I want and right now it's working just fine, but I'm sure in the future this will happen again.

jommgoncalves:
I have the same problem. @farahfa How you solve this problem?

farahfa:
Unfortunately, I haven't solved the issue.. I have a cronjob that runs `docker rm -f $(docker ps -a -q)` at a time where I know no-one is going to be using Jenkins.. Which removes all containers..

Also, not sure if this helps or not but I've updated linux-image, headers & image-extras to latest...

jommgoncalves:
[...]

PR #426, opened 2016-08-19
[FEATURE] Docker Watchdog Async Periodic Work - first version

eaglerainbow:
As being discussed in
https://github.com/jenkinsci/docker-plugin/issues/401, there seems to be
issues with containers which are not cleaned up properly. Therefore, a
Watchdog-like periodic work instance is being introduced, which verifies
the consistency of the containers which are reported to be running on
the docker server (and were provisioned by the current Jenkins instance)
compared with the slaves attached to the current Jenkins instance.
Moreover, containers which have exited, but have not been removed yet,
are reported.

So far, only reports to the logs are being generated. After further
analysis on the issue has been performed, there is the intention to also
trigger clean up commands to the docker server, fixing container leaks.


magnayn:
Seems a reasonable approach to me if the reason for containers not being removed can't be figured out.

I'd have some sort of configurable "reap if older than X" value

eaglerainbow:
> Seems a reasonable approach to me if the reason for containers not being removed can't be figured out.

Well, I surely also prefer to find the root cause(s) for the hanging containers instead of just "blowing away the symptom"...

> I'd have some sort of configurable "reap if older than X" value

So far the pull request only suggests to report on the situation. Yet, it does not stop/remove any containers. However, in principle I agree with you: The TTL should be a configuration option (with some very meaningful default, as I fear that it's very advanced -- and the average user in fact shouldn't need to know what this is all about...)

eaglerainbow:
I have implemented a first version of a TTL-based cleanup for both exited container and running containers, which do not have a proper association to the Jenkins anymore. 
However, I was not able to fully test this yet, as doing so is quite time-consuming.

fraz3alpha:
Just a note to say that trying to match one-to-one the containers running on a machine with those registered in a particular Jenkins instance will not work in the case where the slave node is shared. This is quite a common pattern where multiple teams (distinct Jenkins servers) share one or more Docker hosts, and in particular would not work on a Swarm.

The removal of exited containers is almost certainly still a valid strategy in every case I can immediately think of, but to remove running containers not registered as computers in a particular Jenkins instance has lots of pitfalls that would at least need flagging up in the help, and should definitely remain off by default

eaglerainbow:
> will not work in the case where the slave node is shared.

Sorry if that might sound ignorant, but how do you achieve to share a single slave with multiple Jenkins instances, if the "computers" are pulled up and taken down on-demand, as this plugin usually does?

> This is quite a common pattern where multiple teams (distinct Jenkins servers) share one or more Docker hosts, and in particular would not work on a Swarm.

Please note that since recently the docker plugin sets a property on the container which identifies the Jenkins server from where the container was started uniquely. This approach also runs smooth on swarm clusters (as a matter of fact that was exactly my original case I had: multiple Jenkins instances sharing one single swarm cluster, which gets jammed by orphaned slave instances).

fraz3alpha:
My apologies for not being clear @eaglerainbow , I meant "slave node" in the sense that one large VM running Docker with the API exposed may be used by multiple Jenkins servers - not the internal Computer/Slave object that Jenkins - part of the problem with the terms overlapping! From what you've said, it seems you've already thought about that then.

I wasn't aware that the plugin now tagged containers with the originating Jenkins server, I'll take a look  at that feature now - thanks for mentioning it

eaglerainbow:
@fraz3alpha never mind - it's good that someone third is looking over it. Often you are able to detect a glitch in concepts even before the pull request is merged.
NB: tagging containers with JenkinsId property is available with docker-plugin v0.16.2 (and later). The feature was added by @magnayn with https://github.com/jenkinsci/docker-plugin/commit/459a88f03fb09a9121e066a4e4f897a9cfefcd75 .

eaglerainbow:
Build error is related to the fact that apparently the pom.xml of docker-plugin-shaded is not properly bumped to 3.0.6.
At least http://repo.jenkins-ci.org/public/com/github/docker-java/docker-java/3.0.6/ is available, whilst still the coding refers to 3.0.6-SNAPSHOT
Error message:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project 
docker-java-shaded: Could not resolve dependencies for project com.nirima:docker-java-shaded:jar:0.16.3-SNAPSHOT: 
Could not find artifact com.github.docker-java:docker-java:jar:3.0.6-jenkins in repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/)
```

eaglerainbow:
> Hi, I don't suppose you could update this PR to make it merge again, could you?

So far, response here wasn't that bright - given that we have one comment on it within almost 2 years now. If I didn't know it better, I could think that people out there do survive happily without it... 

Yet, I will give it at least a try, as I have seen yet another use case in real life where such a "watchdog" could be of help: I sometimes see Jenkins servers which had crashed while running dozens of docker-bound slaves, and thus leaving a significant volume of dirt behind on the (swarm/docker) cluster. On boot up the capacity is lost, as nobody is taking care on house-keeping automatically. 
Even with the best "Cloud" configuration integration of any docker-plugin to Jenkins-core, you can't fix this problem, as the Jenkins isn't running anymore at the point of time when it crashes. The watchdog here at least could detect this, and clean up the mess a little after the next reboot. That's for sure not the perfect solution, but at least way better than nothing (which typically ends up in limiting your container capacity and you don't really know from where it's coming).

pjdarton:
Yes, I agree that this plugin languished, unloved, for a long time.  However it's receiving attention again now (it's even been chosen to be part of "[Jenkins Essentials](https://jenkins.io/blog/2018/04/06/jenkins-essentials/)"), and having some form of automated tidy-up functionality is _long_ overdue.
The code has also moved on quite a bit, with a new connection method, improved robustness and cosmetic UI improvements.

More importantly for garbage-collection, when provisioning new containers, in addition to the JenkinsId label (`DockerTemplateBase.CONTAINER_LABEL_JENKINS_INSTANCE_ID`) the plugin now puts in a label (`DockerTemplate.CONTAINER_LABEL_NODE_NAME`) specifying the slave-node name so that any clean-up code can detect if this Jenkins instance still has a slave node corresponding to that container.
i.e. it's now possible to trace _either_ from slave node to container _or_ from container to slave node.  This should make it possible to implement a clean-up process with some pretty good logging to allow good traceability.

It's also now possible to use `listContainersCmd().withLabelFilter(Map<String, String>)` (e.g. see `DockerCloud`'s `countContainersInDocker` method) instead of having to hand-craft a JSON filter, which makes the code a bit simpler ;-)

On the downside, the code churn that's happened since you wrote this has made updating this PR a non-trivial exercise, such that it'd probably be simpler to re-implement from the current master than to do a git merge.

If you're willing to give it a go then that would be appreciated, and you can then bask in the fame of having implemented more awesomeness on this popular plugin :-)
If not then I (and/or others) can probably get something equivalent working by using this PR as an example.

eaglerainbow:
After having battled now 2.5hrs with my former DevEnv, I'll have a look into this topic today. I like the idea with `DockerTemplate.CONTAINER_LABEL_NODE_NAME`, but (this is a more general matter and not related to this PR) I don't like that the prefix container id is gone from the slave's name. I don't know how many times this naming convention has saved me before....

pjdarton:
The reason that the disconnect between slave name, container name and container id happened was because it wasn't possible to keep them 100% connected reliably.

Container IDs are unique within their docker engine/swarm, but not globally unique; Container names have to be unique within their docker engine/swarm, but not globally unique; Slave names have to be unique within a single Jenkins server, but not globally unique.

These differences aren't a problem when you have a one-to-one mapping between Jenkins server and docker engine, however things get problematic if you've got multiple docker engines or multiple Jenkins servers.

If you have more than one Jenkins server running containers in a single docker engine/swarm then you can get duplicated names if the Jenkins server tries to specify the container names.
If you have more than one docker engine/swarm in a single Jenkins server then you can get duplicated container IDs.
The answer is for Jenkins to choose the name it's going to call the Jenkins slave (as it can then ensure uniqueness), and to leave it up to docker to decide what the container name and ID will be (as it can ensure uniqueness)...

... and then to ensure that both ends know all the necessary details to find the other end:
 - The Jenkins slave node knows (in fields) the docker engine/swarm URL and the ID of the docker container, 
 - The docker container knows (in labels) the URL of the Jenkins server and the slave node name.

i.e. the old naming conventions may be gone *but* there are new conventions that take their place.

eaglerainbow:
I managed to get something built up in the meantime, which seems to start ticking already. I did not manage to implement the removal of containers yet, though.

You are right, having the name of the node as label makes some things easier. However, due to other things (esp. `DockerTransientNode` doesn't have any reference to its `DockerTemplate` anymore, which makes verification of the mapping impossible) are also a little trickier, as it's now harder to properly compare the right levels (now, nodes are on the levels of containers - this was different before) against each other (e.g. we must consider that someone noob could have configured two DockerClouds pointing to the same docker instance). 

The stuff I have currently implemented isn't ripe for being publishing, but I am at least on it. Don't expect to see anything before end of this weekend, though...

eaglerainbow:
> i.e. the old naming conventions may be gone but there are new conventions that take their place.

yeah, you are right (and that change might also have been very necessary). However, the new convention is much more complex and significantly less practical (which is a property of this plugin, which I liked the most).

PS: Be careful: The reverse URL, which is used as value of a label isn't necessarily the URL, from which everybody may reach the Jenkins (had some trouble with a similar scenario just recently). However, I also currently do not see a better value to take for the plugin.

pjdarton:
If DockerTransientNode contained a reference to the template it was built from then Jenkins would persist a copy of that template within the Slave Node XML every time the slave node was modified, which isn't ideal.
What we'd need is for every cloud and every template definition to have a unique name or ID and then the node could store that, allowing for each lookup later.
FYI the templates have a similar problem - they don't know the cloud they came from either so they have to look it up by asking each cloud "am I one of your templates?".
I don't have a good answer to this conundrum :-(

Nodes & containers are on the same level, yes:  Given a container's labels, we can ask Jenkins if there's a slave node of the right name.  If there's a container (that was started by this Jenkins server) without a matching slave node then we should consider removing that container (i.e. if it's been like that a long time).  Containers that have slave nodes, even offline slave nodes, should be spared - they'll get removed when the slave node gets removed.

Two clouds, one docker:
Yes, that's possible, albeit rather unusual.  It may even be deliberate in some situations, e.g. if you had two sets of templates and wanted to have "up to 10 containers from the first set of templates and up to 50 of the second set" but didn't mind what the individual instance counts were of individual templates.
I agree that we shouldn't make assumptions, but I'm also of the opinion that we don't need to necessarily go to great effort to ensure optimum performance with weird configurations either.  i.e. we don't need to do a uniqueness-filter before looping over each cloud to check for orphaned containers ;-)

Re: Jenkins URL: I'm not using it programatically.
From my point of view, it's a human-friendly string, not a program-friendly URL.  As a human sysadmin, I much prefer to see a URL than the "JenkinsID" string as I can recognize the URLs, but (as proven by the need to override the Jenkins URL in the JNLP connection config) I am aware of the occasional mismatch between the URL Jenkins has been told it has and the URL that other subsystems see it as.

Re: new convention less practical
I assure you that the change _was_ necessary to avoid naming collisions (I didn't implement this "just for fun").
I am open to suggestions about changes that could improve practicality as long as they don't impact reliability.

Re: before weekend:  No worries; I'm not setting deadlines.
I'm grateful that you're looking at this and I'll choose "well written" over "delivered early" every time (which is why I'm a software developer, not a software project manager ;-) )

eaglerainbow:
> If DockerTransientNode contained a reference to the template it was built from then

I agree to all that; when I had a look at this matter before, it had been totally unclear how all this de-/serialization properly worked out and had quite some troublesome hours with understanding the side-effects. Your current model is much clearer to me, so it's definitively an improvement. However, it also means that I can't implement the "watchdog" here as dirty as it had been (which we can now discuss, whether this wouldn't be a good thing either 😁 ).

>  If there's a container (that was started by this Jenkins server) without a matching slave node then we should consider removing that container 

That's the 90% target of the entire endeavor here.
==> 100% taken.
 
> Containers that have slave nodes, even offline slave nodes, should be spared - they'll get removed when the slave node gets removed.

Yep, also this is agreed. 

There is yet another class, though: We might have slave nodes, which do not have a container anymore: I had this case several times already - the agents then need to deleted manually (and sometimes), this is even a mess (e.g. if you are persisted workflow scripts/Jenkins 2.0 Pipeline scripts, which had serialized the name of the slave, but Jenkins crashed as a whole).
We can detect that situation (and at least log them), but it will be hard to really fix those cases.

> [Two cloud configurations to the same docker] Yes, that's possible, albeit rather unusual.

Unfortunately, I know one server at least, which has this 😅 

> I agree that we shouldn't make assumptions, but I'm also of the opinion that we don't need to necessarily go to great effort to ensure optimum performance  [,..]

Well, I am not striving for performance in this case, but code must not break either. We need to ensure that we are resilient enough to survive such configurations. Otherwise, you will have troubles with staying an "essential's part of Jenkins"... 😛 

> [new convention less practical] 
> I am open to suggestions about changes that could improve practicality as long as they don't impact reliability.

I see your need and forces, and as I currently don't have a better idea (in the sense of a PR), I can't complain...
Just one brief thought: Now, you are more or less generating a random id for the slaves. Would it be an option to 

a) either prefix the container id to the generated value (e.g. `slave-<6-digits-of-container-id>-<locally-generated-id>`)
or
b) only generate a new id, if a collision is detected?

Just some food for the brain -- Though very interesting, we should stop this discussion (at least) here at this place, as w.r.t. to this PR it simply is off-topic...

> I'll choose "well written" over "delivered early" every time (which is why I'm a software developer, not a software project manager ;-)  )

Me, too 😀

eaglerainbow:
If you are interested, you may have a brief glimpse into https://github.com/eaglerainbow/docker-plugin/compare/master...DockerContainerWatchdog2018 - I have uploaded the state as far as I could get up to now. Major obstacle, which I think I am quite through with in the meantime, is the matter to properly unit test the code. Right now I have twice as much test code than productive code 😞 and it will become even more...

pjdarton:
De-/serialization:  Basically anything that isn't transient gets stored.
This means you can't have "just a pointer back up the tree" without repercussions.
In the case of the DockerTransientNode I suspect that we may be able to live with those repercussions but I've not tried it.

Re: slave nodes without containers
If we find a Docker slave node that's offline, and we're 100% confident that we're in communication with the correct docker API for that container and we're sure that there is no container for it, then I think we can be reasonably confident that it's ok to get rid of it.
...and I'm pretty sure that it is possible to get rid of slave nodes programatically - `Jenkins.getInstance().removeNode(Node n)` should do the trick.  At least, I think that's that would be required.

Re: slave names/ids
This is a chicken/egg problem - we need to know what the slave name will be _before_ we start the container, which is before we've got a container-id for it.
For JNLP to work, we need to pass the slave name and "secret" ( = a hash of the slave name) to the container, and those arguments are provided _to_ the "docker run" command and we only find out the container-id _from_ the "docker run" command's result.
i.e. by the time we know what the container-id is, it's too late to change the slave node name.

We _could_ make Jenkins set the container name to be the Jenkins slave node name and then retry whenever Docker refused because of the duplicate name _but_ that does result in really messy code - there's quite a lot that needs to happen between deciding what the slave name is and calling docker, and we'd have to re-run all of that, and some of that includes a "before X" ... "after X" pattern so it would make that API really messy to have to cope with a "not doing X after all, please do before Y instead" scenario.
Letting the docker engines choose names is a lot simpler.

If you can figure out a fairly clean way of having the docker plugin set container names _and_ automatically cope with naming collisions then feel free to submit a PR (or even an issue describing an approach that we could jointly turn into a PR).

Re: test code
In my experience, a 2:1 ratio is not unreasonable, although a 1:1 is more common (although a total absence of test code is also quite common on Jenkins plugins!).
At the risk of telling you stuff you already know...
Start with writing (in human text) the test scenarios you want to test, then write test code, then refactor to extract out common setup and/or test code, and repeat, refactoring to extract common code so that each @Test method is mostly human-readable scenario description and all the lengthy setup code is hidden in common methods shared between tests.
Also, if the code is difficult to test, consider refactoring the code-under-test so that the main algorithm being tested is more accessible to the test code - feel free to use @Restricted(NoExternalUse.class) on package-level methods to give the test access to the inner-workings.
e.g. refactor the code so that it has a method that's given a list of containers (that our Jenkins instance started) and a list of (docker) nodes, it returns a list of containers that should be removed.  And another method that decides on nodes to be removed.  Keep the "find all nodes", "find containers" and container/node removal code separate.
If you do that then the complex test scenarios can be tested with a simple "pass these arguments to that method, expect that result" white-box test.  You'll still need an end-to-end basic test to verify that the well-tested-algorithm methods get called, but you can do the complex tests purely on the inner algorithm which will save on setup code.

eaglerainbow:
First of all, I have created #658  in the meantime, as I believe that it makes sense to start reviewing the code (I believe that there are several iterations of the coding necessary anyway). That is why I would suggest to close this PR here and continue over there.

> Re: slave nodes without containers
> If we find a Docker slave node that's offline, and we're 100% confident 

That's what I have done now... I am interested in hearing whether the checks before the removal of the node are strict enough for you. 

> Re: test code

I decided to use a slightly different approach with the `TestableDockerContainerWatchdog`: The interface of the coding towards Jenkins only requires some "global variables" as input parameters plus some further "global methods" for notifying about the actions to be performed. I isolated those in methods of visibility `protected`. Afterwards I could derive `TestableDockerContainerWatchdog` from `DockerContainerWatchdog` and overwrite those methods, which represent the interface from/to Jenkins. The alternative would have been to virtually make every method `public` + ` @restricted(NoExternalUse.class)`, which would have made the productive coding quite ugly (now, it's just a mocking orgy, but that's at least buried in the test coding only).

PR #440, opened 2016-09-22
Fix reading of network settings from docker.

amerkel2:
This pull request fixes https://issues.jenkins-ci.org/browse/JENKINS-38135 by reading the container's IP address from the `network` map of the container settings instead of using the deprecated top-level `IPAddress` field in case a custom network is defined for the docker template.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java
@@ -139,11 +139,12 @@ public class DockerComputerSSHLauncher extends DockerComputerLauncher {
         //get address, if docker on localhost, then use local?
         if (host == null || host.equals("0.0.0.0")) {
             String url = DockerCloud.getCloudByName(cloudId).getServerUrl();
-            if( url.startsWith("unix") ) {
+            host = getDockerHostFromCloud(cloudId);
+
+            if( url.startsWith("unix") && (host == null || host.trim().isEmpty()) ) {
                 // Communicating with local sockets.
                 host = "0.0.0.0";
             } else {
-                host = getDockerHostFromCloud(cloudId);
 
             /* Don't use IP from DOCKER_HOST because it is invalid or we are
              * connecting to a system that supports a single host abstraction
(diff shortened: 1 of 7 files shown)


docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java
```diff
@@ -150,7 +153,11 @@ protected InetSocketAddress getAddressForSSHD(String cloudId, InspectContainerRe
              * like Joyent's Triton. */
                 if (host == null || host.equals("0.0.0.0") || usesSingleHostAbstraction(ir)) {
                     // Try to connect to the container directly (without going through the host)
-                    host = networkSettings.getIpAddress();
+                    if (network != null && network.length() > 0) {
+                        host = networkSettings.getNetworks().get(network).getIpAddress();
```
lucamilanesio: That wouldn't work either when the Jenkins master is running in a Docker container itself. Slaves ports are exposed through the Docker hostname and not on the virtual network IP. I tried your PR and didn't work for me. I actually needed to add a new "Docker hostname" configuration to specifically tell what the physical IP looks like. See https://github.com/jenkinsci/docker-plugin/pull/443

amerkel2: In this code path, the slave is not connected to via the exposed port, but directly via port 22 (cf. https://github.com/amerkel2/docker-plugin/blob/edd1a87f32ec216d573b0c5cf63b299b2607d8ba/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java#L161). The objective of this PR is to improve detection of the virtual network IP of the container. That the container cannot be reached via this IP in all scenarios (e.g., if the master runs on a different physical host) is a different matter that could indeed be solved by adding the host to connect to via a separate config option.

lucamilanesio:
@amerkel2 if you rebase this PR now, it should be building successfully again 😄

pjdarton:
This looks like it was obsoleted by other changes a long time ago (e.g. the "Docker Hostname or IP address" field has existed for at least 2 years).

PR #443, opened 2016-09-26
Configure Docker hostname when using unix socks

lucamilanesio:
When using Docker URL with unix-domains sockets we cannot access the
docker hostname from the host part of the URL. Even using the docker
inspection of the machine created dynamically would not expose the
right value accessible from the outside.

Allow to explicitly define the Docker hostname you want to use when
Jenkins master needs to connect to slave instances created.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -72,6 +72,7 @@ public class DockerCloud extends Cloud {
     public final int readTimeout;
     public final String version;
     public final String credentialsId;
+    public final String dockerHostname;
 
     private transient DockerClient connection;
 
@@ -104,7 +105,8 @@ public class DockerCloud extends Cloud {
                        int connectTimeout,
                        int readTimeout,
                        String credentialsId,
-                       String version) {
+                       String version,
+                       String dockerHostname) {
         super(name);
         Preconditions.checkNotNull(serverUrl);
         this.version = version;
@@ -112,6 +114,7 @@ public class DockerCloud extends Cloud {
         this.serverUrl = sanitizeUrl(serverUrl);
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
+        this.dockerHostname = dockerHostname;
 
         if (templates != null) {
             this.templates = new ArrayList<>(templates);
@@ -134,7 +137,8 @@ public class DockerCloud extends Cloud {
                        int connectTimeout,
                        int readTimeout,
                        String credentialsId,
-                       String version) {
+                       String version,
+                       String dockerHostname) {
         super(name);
         Preconditions.checkNotNull(serverUrl);
         this.version = version;
@@ -142,6 +146,7 @@ public class DockerCloud extends Cloud {
         this.serverUrl = sanitizeUrl(serverUrl);
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
+        this.dockerHostname = dockerHostname;
 
         if (templates != null) {
             this.templates = new ArrayList<>(templates);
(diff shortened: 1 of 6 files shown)


lucamilanesio:
@magnayn do you recall why you were using a docker-java.3.0.6-jenkins version? Do you recall the list of hot-fixes you applied? Seems that those fixes broke the unix:/// docker socket connections :-( This PR would fix the issue but I need to know what are the fixes you included and are not in my PR. Thank you in advance for your feedback.

magnayn:
The 3.0.6-jenkins is built from
https://github.com/magnayn/docker-java/commits/master

The only significant diff is
https://github.com/magnayn/docker-java/commit/efd5f2c76e429a2ca589cd435e29bebb365e6c9d

If https://github.com/docker-java/docker-java/pull/660 got applied I could
return to mainline, or possibly a resolution for
https://github.com/docker-java/docker-java/issues/160.

I don't predict either of those things to happen any time soon.

FWIW, 3.0.6-jenkins should work with unix sockets, as I use it on a
different project in that way.

lucamilanesio:
@magnayn thanks for your feedback, I'll try to build your 3.0.6-jenkins from source as it may be possibly just a temporary build hiccup on latest Jenkins plugins central distribution.

What I've seen was that when using unix:/// in the Docker URL I got an exception saying that UNIX domain sockets were not supported. 

Additionally is not clear to me how can then the master discover the slave's IP to use for SSH connectivity, as the docker hostname isn't in the URL anymore when using Unix sockets. For this reason I've added an explicit "Docker Hostname" field to be used. How did you manage to allow discovery of the hostname / ip without that field?

magnayn:
so
https://github.com/jenkinsci/docker-plugin/commit/954ebdab1a6c79e32eb9d7691145574c8624a3a2
changed
it so that if the url started with unix, to determine that the
host was your local one.

I defaulted it to 0.0.0.0. I guess docker could be bound somewhere else
(don't know in practice if anyone does that), in which case a hostname to
talk to probably makes sense

lucamilanesio:
It's not clear to me however how can https://github.com/jenkinsci/docker-plugin/commit/954ebdab1a6c79e32eb9d7691145574c8624a3a2 work: trying it on my Jenkins/Docker master I got:
```
[09/27/16 17:32:30] SSH Launch of local-c7f8ee51f6ed on 0.0.0.0 failed in 118 ms
```

In the code the 0.0.0.0 is translated into:
```
              if (host == null || host.equals("0.0.0.0") || usesSingleHostAbstraction(ir)) {
                   host = networkSettings.getIpAddress();
                   port = sshConnector.port;
              }
```
... which isn't necessarily true. When the Jenkins master itself is a Docker image (which shares the socket with the Docker host) the container IP address is not the Docker hostname.

We definitely need an extra configuration parameter to say what is the hostname (or IP) to use in that case.

lucamilanesio:
@magnayn I've removed the commit about the 3.0.6-jenkins version as I verified that rebuilding it locally from your GitHub repo, the Unix sockets actually work. It was then possibly just a hiccup, apologies.

Using the 3.0.6-jenkins dependency however, will make this Pull Request build fail 😢  because the dependency isn't available anywhere. See https://jenkins.ci.cloudbees.com/job/plugins/job/docker-plugin/660/console

Do we need then to include the customised docker-java as submodule and building it as part of the docker-plugin build? Or can we just use the "plain vanilla" ?

lucamilanesio:
@magnayn any news on this PR? Can we merge it? Does it makes sense?

magnayn:
Mostly - the only thing we should probably do is default the case of
unix:// sockets if you don't specify a dockerHost, as 0.0.0.0 is a
reasonable default and is probalby true in 90% of cases. That's what it's
doing ATM

I.E: if url is unix://
AND dockerHost is null
host = 0.0.0.0

lucamilanesio:
@magnayn done, I made the change more conservative and, in case Docker Hostname is not configured, will behave exactly as before, using 0.0.0.0

I've rebased as well on master, so ,this time, the Jenkins build shouldn't fail 😎

lucamilanesio:
... and, finally, the build is *green* ... oops *blue*  🇯🇵

PR #444, opened 2016-09-27
Bumping docker-java.version to 3.0.6

eaglerainbow:
Version 3.0.6-jenkins of java-docker apprently is no longer available for CI builds. That is why pull-requests are currently failing with their builds (thus, making everyone blind to real errors which need to be fixed).

This change bumps the docker-java version to the final version of 3.0.6 which is currently in place (needs to be checked, if this version is also ok from a delivery perspective).

### DIFF
--- a/docker-java-shaded/pom.xml
+++ b/docker-java-shaded/pom.xml
@@ -33,7 +33,7 @@
              submodule. But for now it's probably better to allow CI to work
             as it will at least build on the CI. -->
 
-        <docker-java.version>3.0.6-jenkins</docker-java.version>
+        <docker-java.version>3.0.6</docker-java.version>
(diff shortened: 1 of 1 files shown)


lucamilanesio:
Look at the discussion on https://github.com/jenkinsci/docker-plugin/pull/443: it seems that @magnayn explicitly needs its custom build because of a one-line change on the registry SSL validation.

eaglerainbow:
Damn it  :worried:

magnayn:
No, I think it's fine - I'll just have to remember to apply the patch
before releasing. I think it's better to not fail the CI build, and hauling
in a submodule is probably overkill for such a minor diff.

PR #446, opened 2016-09-30
Give high priority to Docker Hostname definition

lucamilanesio:
When the Docker Hostname is defined at Cloud level, it should be used
as preferential setting over the other automatisms.


lucamilanesio:
@magnayn I need to apologise, but there is a small follow-up to my previous PR. The fix I made works but *not* for the unix:/// sockets. I made this extra PR to fix the issue, do you have any time to have a look? Thank you in advance.

lucamilanesio:
@magnayn ping ...

magnayn:
thanks!

PR #449, opened 2016-10-12
Pull image improvement

slawekjaranowski:
Correct implementation for PULL_LATEST strategy
which doesn't pull if image name doesn't contains latest tag.

### DIFF
--- a/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategyTest.java
+++ b/docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategyTest.java
@@ -1,0 +1,106 @@
+package com.nirima.jenkins.plugins.docker;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


KostyaSha:
M.... seems right. `latest` and docker image names is real pain.

docker-plugin/src/test/java/com/nirima/jenkins/plugins/docker/DockerImagePullStrategyTest.java
```diff
@@ -0,0 +1,106 @@
...
+@RunWith(Parameterized.class)
+public class DockerImagePullStrategyTest {
+
+    private final List<Image> imageList;
+    private final String imageName;
+    private final DockerImagePullStrategy pullStrategy;
+    private final boolean shouldPull;
+
+    @Parameterized.Parameters(name = "existing image: ''{0}'', image to pull: ''{1}'', strategy: ''{2}''")
+    public static Collection<Object[]> data() {
```
KostyaSha: I initially wrote this strategies, can i borrow your fix & test into https://github.com/KostyaSha/yet-another-docker-plugin ?

slawekjaranowski: Yes, you can if you wish.

PR #454, opened 2016-10-20
Docker Port Binding during SSH-Connection startup

gamma:
Do not require the port binding for the ssh service to be on the 0.0.0.0 address.

The problem is that Windows Server 2016 docker implementation does not support "host IP addresses in NAT settings" - Removing the fixed IP solves the problem and makes no problems with *NIX based servers.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java
@@ -71,7 +71,7 @@ public class DockerComputerSSHLauncher extends DockerComputerLauncher {
             createCmd.withCmd("bash", "-c", "/usr/sbin/sshd -D -p " + sshPort);
         }
 
-        createCmd.getPortBindings().add(PortBinding.parse("0.0.0.0::" + sshPort));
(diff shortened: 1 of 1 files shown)


gamma:
Is someone reviewing the pull requests?

fboudra:
+the maintainer: @magnayn

laenzlinger:
This change has the side effect that the hostname can not be resolved anymore in a swarm based setup. 

With version 0.16.2 the slave name was resolved to [cloud-name]-[container-id]-[docker-host-name]
With this change the slave name will be  [cloud-name]-[container-id]-localhost in my environment.

The following code is defining the slave name: https://github.com/jenkinsci/docker-plugin/blob/master/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerSwarmSlave.java

docker inspect on the created container shows that the HostIp is empty, and therefore the hostname can not be resolved.

```
  "HostConfig": {
            "Binds": null,
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "json-file",
                "Config": {}
            },
            "NetworkMode": "default",
            "PortBindings": {
                "22/tcp": [
                    {
                        "HostIp": "",
                        "HostPort": ""
                    }
                ]
            },
```

Before this change, the host could "HostIp" was filled with the IP of the docker host where the image was started.

I have to do further research for why docker swarm does not set HostIp anymore with this change.

PR #456, opened 2016-11-03
[FIXED JENKINS-39473] Switch to bouncycastle-api plugin for bouncycastle jars

stephenc:
See [JENKINS-39473](https://issues.jenkins-ci.org/browse/JENKINS-39473)

@jenkinsci/code-reviewers @reviewbybees

### DIFF
--- a/docker-plugin/pom.xml
+++ b/docker-plugin/pom.xml
@@ -28,11 +28,34 @@
         </developer>
     </developers>
 
+    <properties>
+        <jenkins.version>1.648</jenkins.version>
+    </properties>
+
     <dependencies>
-         <dependency>
+        <dependency>
+            <!-- TODO remove this once we upgrade baseline Jenkins to 2.16 and bouncycastle-api to 2.16 -->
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
LGTM 🐝

magnayn:
Is the BC dependency in 0.16.2 breaking things currently (I.E: should a 0.16.3 be rolled )?

stephenc:
@reviewbybees done

stephenc:
@magnayn I suspect that people running Jenkins 1.625 - 1.647 may at random be having issues as it all depends on plugin initialization order. I do not have any specific customer issues driving this.

On Jenkins 1.648+ the bundling of bouncycastle should be idempotent (for now) as everyone is on the same version...

Rolling a 0.16.3 now will not help the people on 1.625 - 1.647... there is no fix for them other than getting a version of the docker plugin that uses the older bouncycastle. The people on 1.648+ will just have duplicate jars on the classpath (which may break things if they are running Java 9)... as such I see this as your call w.r.t. your time priorities

Issue #407, opened 2016-07-21
security issue with binding all started containers with 0.0.0.0::port


pault28:
There is a major security issue with the way containers are provision using the plugin.
All started containers with Docker ssh launcher results in the provisioned container accessible from the outside world - as long one knows the computer name from which they are running.

At the moment the code I can attribute to this is https://github.com/jenkinsci/docker-plugin/blob/1b4b2546567bdb0e1ffcb2a64075a5ed7a1c457e/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher.java#L71

Can we make this configurable please?

`Jenkins ver. 1.646`
`docker plugin 0.16.0`

To see what I mean after starting a container with the plugin run `docker ps`:

```
sudo docker ps
CONTAINER ID   IMAGE            COMMAND     CREATED        STATUS        PORTS                   NAMES
211a95fb6373  IMAGE_NAME  "bash -c '"   2 seconds ago   Up 1 seconds  0.0.0.0:32777->22/tcp   sick_meitner
```

Loki-Afro:
duplicate of #20?

ndeloof:
closing as a duplicate of https://github.com/jenkinsci/docker-plugin/issues/20

Issue #20, opened 2014-05-13
Be able to fix public IP


guilhem:
Currently all ports are bind on `0.0.0.0`
It may be useful (and secured) to be able to fix another IP (for example the IP used to call API ;) )

In my case, Host server have a public address and a internal bridge. I only want to see docker vm accessible by internal bridge :)

More generally, be able to pu our own config in API can be great.

KostyaSha:
@guilhem do you want enter address manually or resolve from interface name, etc?

pjdarton:
I believe that the ability to specify which network a container is visible on was added in [this commit](https://github.com/jenkinsci/docker-plugin/commit/d5553d47c95defbb70079a5eb263fd6a61ebbc31) in March 2016.

PR #472, opened 2017-01-12
[FIX docker-plugin-407] Make ssh port binding configurable

pault28:
- Fix issue #407  - security issue with binding all started containers with 0.0.0.0::port
- This adds an option box that controls how the ssh port is bound when starting the container. Ticking it results in `127.0.0.1:32770->22/tcp`.
Default  and current behaviour (not ticking) is `0.0.0.0:32770->22/tcp`
- related to fix for #20

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/docker-plugin/472)

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
@@ -120,6 +120,9 @@ public abstract class DockerTemplateBackwardCompatibility {
     private transient boolean bindAllPorts;
 
     @Deprecated
+    private transient boolean bindSshPortLocalhost;
+
+    @Deprecated
     private transient Integer memoryLimit;
     @Deprecated
     private transient Integer memorySwap;
@@ -128,6 +131,7 @@ public abstract class DockerTemplateBackwardCompatibility {
 
     @Deprecated
     private transient boolean privileged;
+
     @Deprecated
     private transient boolean tty;
 
@@ -222,6 +226,7 @@ public abstract class DockerTemplateBackwardCompatibility {
                         cpuShares,
                         bindPorts,
                         bindAllPorts,
+                        bindSshPortLocalhost,
                         privileged,
                         tty,
                         macAddress)
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -90,6 +90,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase> {
 
     public final String bindPorts;
     public final boolean bindAllPorts;
+    public final boolean bindSshPortLocalhost;
 
     public final Integer memoryLimit;
     public final Integer memorySwap;
@@ -119,6 +120,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase> {
                               Integer cpuShares,
                               String bindPorts,
                               boolean bindAllPorts,
+                              boolean bindSshPortLocalhost,
                               boolean privileged,
                               boolean tty,
                               String macAddress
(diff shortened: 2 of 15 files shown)


pault28:
Open for review @kohsuke  @magnayn @Loki-Afro  @guilhem. Thanks.

Loki-Afro:
Reviewed 15 of 15 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/jenkinsci/docker-plugin/472)*

robfraz:
Hello - it seems that this has been reviewed now?  Is it likely to get merged soon?

Loki-Afro:
@magnayn any update? ;)

ndeloof:
I agree binding to `0.0.0.0` expose the build container to arbitrary network, and we might want some finer grained control on it. Not sure binding to localhost would be the best solution to this issue.
I think we at least should generate a dedicated key pair when a new docker agent is created, injected as `authorized_keys` into container during setup.

pault28:
@ndeloof - This PR does not bind to localhost by default. It rather introduces an option to bind to localhost whilst keeping the default behavior of binding to `0.0.0.0` - https://github.com/jenkinsci/docker-plugin/pull/472/files#diff-ab1f960647743db55b4a9363588341adL74

ndeloof:
@pault28 well understood, I'm not complaining about lack of backward compatibility, just wonder we have a better way to secure such a container - in addition to avoid binding to all available network interfaces. Also, IIUC binding to localhost will limit use for local containers, preventing use on docker swarm cluster or comparable. So at least a more flexible option would be necessary but just a boolean flag.

pault28:
Thanks @ndeloof - I personally think adding a better security is great but it may well fall out of the scope of the plugin. 
Our use case is we build several docker images with different dependencies which are then used to build different jenkins job. 
The problem was whilst the jenkins server had its security settings well configured, our port scanners picked up this vulnerability which makes it possible for anyone with the jenkins server hostname/ip and the port used for any of the containers to log in as long as they know the username and password of the jenkins user.

Adding strongly security is desired but I really think this should be merged to address the original issue. We've been using the patched version of this plugin since this vulnerability was discovered. This limits of us from picking new releases/changes.

Any chance of getting this merged soon?

ndeloof:
I have a working fix that relies on Jenkins master's indentity ssh keys to secure traffic to the sshd service running inside container, and as such does not require any "open secret" to be bundled into the docker image. With this setup, container can only be accessed by jenkins master, even if ssh listen on `0.0.0.0`.
That doesn't mean I'm opposed to restricting port binding, but limiting to `localhost` only assume a minimalist use-case, I'd prefer a more advanced option to let user configure network to match various scenarios.

pault28:
Well, if you are not opposed to port binding restriction can you get this merged please? Our security team has deemed this port exposure as a no no irrespective of how secure access to the containers is.
Your fix can then be an improvement/addition to the security of containers launched via the plugin.

ndeloof:
@pault28 as I explained I've asked for more flexibility, not just a boolean flag for localhost-only binding. If you need this security fixed on your own ASAP then feel free to deploy internally a patched version of this plugin.

pault28:
No problem @ndeloof. That's exactly what we have been doing for the past 9-10 months.

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -119,6 +120,7 @@ public DockerTemplateBase(String image,
                               Integer cpuShares,
                               String bindPorts,
                               boolean bindAllPorts,
+                              boolean bindSshPortLocalhost,
```
ndeloof: I'd prefer a more flexible binding definition, falling back to `0.0.0.0` when not set, but offering more flexibility for various use-cases than just "limit to localhost"

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBackwardCompatibility.java
```diff
@@ -120,6 +120,9 @@
     private transient boolean bindAllPorts;
 
     @Deprecated
+    private transient boolean bindSshPortLocalhost;
```
ndeloof: why introduce this here ?

PR #473, opened 2017-01-17
JENKINS-41035: fix browsing slave workspace with nested folders

cpfeiffer:
(or otherwise custom workspace directories)

This maps the workspace folder from the slave path to the master path instead of assuming the path is JENKINS_HOME/workspace/job-name.


ndeloof:
unfortunately this whole feature seems to be broken by design as it assumes a local docker daemon. A better approach relying on volumes is to be investigated on https://issues.jenkins-ci.org/browse/JENKINS-47576.

Issue #478, opened 2017-02-22
IdleMinutes configuration not properly restored on restart


eaglerainbow:
On several occasions I could observe that docker-plugin does not restore the idleTimeout value properly for Retention Strategies other than "Docker Once" after Jenkins has performed a restart. After some debugging I ended up in the following case:

![idleminutesnotpropagated](https://cloud.githubusercontent.com/assets/12811079/23232414/788d1fee-f94b-11e6-933c-8a29a8710b5c.png)
(screenshot taken running core 2.25)

Apparently, the root cause is that `DockerCloudRetentionStrategy` extends `CloudRetentionStrategy` and both do have a variable called `idleMinutes`. Note that
* the `config.xml` contains enough information to properly restore the variables
* the `idleMinutes` variable is restored properly for `DockerCloudRetentionStrategy` but not for its parent
* `CloudRetentionStrategy` only provides an API to define the `idleMinutes` on construction, but does not have any setter method.

As consequence, I have seen several Jenkins servers running wild with the following behaviour:
1. A new container is requested and started, because a job is requesting execution capacity for a corresponding label.
2. However, there is no time for the job to be scheduled to an executor before...
3. the idle timeout of the slave/computer strikes and the docker container is removed.
4. Yet, the job still is requesting an instance to be executed on. Therefore, another container is started and the game starts over.

This leads to very high load averages on the server infrastructure, as hundreds of containers are being started and immediately removed without doing anything productive. 

I played around with the coding and - for the sake of testing - changed the name of the `idleMinutes` attribute in `DockerCloudRetentionStrategy` locally (it's clear that this change is incompatible and thus not a publishable workaround). The new name of the attribute in `DockerCloudRetentionStrategy` was `dockerIdleTimeout`. The result was prompt: 

![dockeridleminutes](https://cloud.githubusercontent.com/assets/12811079/23232726/a16ad108-f94c-11e6-8fc9-fcbdbd87597c.png)

The content of the variable in both instances could be restored properly (NB: please ignore that the value is 6 in the screenshot instead of 5).

## Known Workaround
(using delivered state by the docker-plugin)
A known workaround for the problem mentioned above is to start the `/configure` screen after Jenkins was restarted. Change the value of `idle timeout` to some other value, click `save` and repeat to get back to the old value. Once performed, the idle timeout is obeyed properly again.

eaglerainbow:
related to #270 - seems that the correction there did not fix it all

[...]

PR #479, opened 2017-02-23
Fix: idleMinutes = 0 in CloudRetentionStrategy after XStream deserialization

eaglerainbow:
see also https://github.com/jenkinsci/docker-plugin/issues/478


ndeloof:
fixed

PR #488, opened 2017-04-19
Improve container garbage collection

kepeket:
Very often, when containers are unexpectedly killed, the plugin keeps tracks of them, preventing Jenkins to pop up additional slaves.

This workaround allow the plugin to get the exact amount of containers running by polling the docker api from time to time

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -51,6 +51,7 @@ import javax.annotation.CheckForNull;
 import javax.servlet.ServletException;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


pjdarton:
Hi @kepeket 
I think that the mass of other changes that have occurred within the plugin between when you raised this PR and now (version 1.1.4) have probably obsoleted the need for this PR.
e.g. the plugin doesn't permanently keep track of containers.

I believe that the plugin could do with some additional garbage-collection functionality (to remove any old containers started by Jenkins that don't have a slave node, and to remove any old slave nodes that don't have a container), but this PR doesn't address that need.

Do you believe that we still need this PR?
If so, can you resolve the merge issues and provide more detail as to why this PR is still required?

kepeket:
I don't really know if this PR is still needed, I've stopped using the plugin and I'm poping docker directly instead.

Closing.

PR #493, opened 2017-06-02
Display and exit on curl failures.

jaryn:
Display error when curl fails to download the slave.jar.


ndeloof:
looks good but the current approach is just not pleasant. Docker plugin could rely on `docker cp` to install `slave.jar` as well as it's init script. I'd also like to get this aligned with https://github.com/jenkinsci/docker-jnlp-slave as much as possible

ndeloof:
obsolete with new JNLP launcher design

PR #494, opened 2017-06-06
Exclude jersey-apache-connector from docker-java-shaded to fix "org.apache.http.conn.UnsupportedSchemeException: unix protocol is not supported"

erioness1125:
`org.glassfish.jersey.apache.connector.ApacheConnector` could be loaded from either jersey-apache-connector.jar or docker-java-shaded.jar. As docker-java-shaded includes all the classes in the org.glassfish package, it should be guaranteed to be loaded from the shaded jar, not the original jersey jar. I excluded it from docker-java-shaded.


erioness1125:
Issue #495

ndeloof:
obsolete (moved to docker-java-api-plugin)

PR #512, opened 2017-09-20
make DOCKER_HOST optional

pudi1975:
seems the initial intent was to let container access it's containing host, but actually this prevent a build to rely on /var/run/docker.sock bind mount. Also, there's no guarantee the hostname/ip as declared in jenkins for connectivity from master is relevant from container.


ndeloof:
relying on `/var/run/docker.sock` is a terrible idea anyway, this just open the security pandora box. For longer terms I'd like to introduce some composition support in agent templates so one can define a set of container to match mostly all needs, and if actual docker infra is required, the TLS keys is a better way to expose docker daemon as you can control resources using some authz-plugin.

ndeloof:
fixed by https://github.com/jenkinsci/docker-plugin/commit/9421e3d571008ada10e1c9ac28e2fd1adb0859aa

PR #514, opened 2017-10-04
Fix name of slave on swarm.

Typz:
When using recent versions of Swarm, hostIP does not seem to be
available in the PortBindings section of HostConfig, but in the ports
section of NetworkSettings. docker inspect response looks something like
this:

   [...]
   "HostConfig": {
       "Binds": null,
       "ContainerIDFile": "",
       "LogConfig": { ... },
       "NetworkMode": "default",
       "PortBindings": {
           "22/tcp": [
               {
                   "HostIp": "0.0.0.0",
                   "HostPort": ""
               }
           ]
       },
       ...
   }
   "NetworkSettings": {
       "Bridge": "",
       "SandboxID": "e73dee6dcda0985515a2b21c8ffa9690db633278b033409658a809149ddf0f41",
       "HairpinMode": false,
       "LinkLocalIPv6Address": "",
       "LinkLocalIPv6PrefixLen": 0,
       "Ports": {
           "22/tcp": [
               {
                   "HostIp": "172.20.182.49",
                   "HostPort": "37032"
               }
           ]
       },
       ...
   }
   [...]

To handle both situations, now try to get host hostip from
NetworkSettings.Ports if we only get 0.0.0.0 from HostConfig.portBindings.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/JenkinsUtils.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/utils/JenkinsUtils.java
@@ -121,17 +121,23 @@ public class JenkinsUtils {
         });
     }
 
-    public static String getHostnameFromBinding(InspectContainerResponse inspectContainerResponse) {
-        Map<ExposedPort, Ports.Binding[]> bindings = inspectContainerResponse.getHostConfig().getPortBindings().getBindings();
+    private static String getHostnameFromBinding(Map<ExposedPort, Ports.Binding[]> bindings) {
         if (bindings != null && !bindings.isEmpty())  {
             Ports.Binding[] binding = bindings.values().iterator().next();
             if (binding != null && binding.length > 0) {
                 String hostIp = binding[0].getHostIp();
-                return getHostnameForIp(hostIp);
+                if (!"0.0.0.0".equals(hostIp))
+                    return getHostnameForIp(hostIp);
             }
         }
-
         return null;
+    }
+
+    public static String getHostnameFromBinding(InspectContainerResponse inspectContainerResponse) {
+        String hostname = getHostnameFromBinding(inspectContainerResponse.getHostConfig().getPortBindings().getBindings());
+        if (hostname != null)
+            return hostname;
+        return getHostnameFromBinding(inspectContainerResponse.getNetworkSettings().getPorts().getBindings());
     }
 
     private static String getHostnameForIp(String hospIp) {


ndeloof:
My plan is to extract getHostnameFromBinding into some configurable strategies so end-user can define the sshd IP or how to retrieve it, and we stop doing such guess on user infrastructure, which demonstrated to just be wrong in many situations

Typz:
I'll happily use any improvement, though I am not really keen on leaving this to the user: I would prefer the tool to automatically find out the address. But indeed having different strategies and a way to (automatically, please!) select the right would be nice.

But until that time has come it would be nice to have things working: we used to have a everything setup, and address retrieval failed when we upgraded swarm (from 1.1.3 to 1.2.8, I think)

ndeloof:
> tool to automatically find out the address
This is just not possible. In many deployment the IP you get from docker inspect is not the one available from external network. There's no magic here

Typz:
The IP you get from docker inspect should be reachable from Jenkins (otherwise you cannot connect to the slave!), which should be enough IMO : we don't try to connect to this address, simply to display an "identifier" of the actual host.

ndeloof:
You got the point : in many scenarios the IP you get from docker inspect is _not_ reachable from Jenkins and user report issues. Your request is a side-case for a larger issue.

ndeloof:
Actually I haven't been able to create a test setup where I get docker inspect to return any hostIp. As container is created with `-P` to expose ssh port 22 on random available port, I only get 
```
"PortBindings": {
                "22/tcp": [
                    {
                        "HostIp": "",
                        "HostPort": "1234"
                    }
                ]
            }
```
Do you get an actual IP reported here ?

Typz:
Using an older version of swarm, we indeed had the host IP in HostConfig/PortBindings.
Using latest swarm version we have the actual IP in NetworkSettings/Ports, and 0.0.0.0 in HostConfig/PortBindings.

ndeloof:
ok I can confirm this indeed happens with Docker swarm standalone, but is specific to this setup. 
Sounds weird this was reported under `HostConfig/PortBindings` as this is the container configuration spec, not actual state reported here. Maybe an API implementation mistake, corrected by latest swarm. So we need to cover both :-\

ndeloof:
as your goal is to report the node where the container is running, sounds to me it would make more sense to get the `Node.Name` from docker inspect output, don't you think so ?

Typz:
`Node.name` works for me; but changing this would conflict with #515 : should I merge the pull requests, and always add the Node.name (if available, without checking isSwarm) ?

ndeloof:
`Node.Name` will bet undefined on non-swarm docker daemon, so easy to filter
docker-java as a strongly typed client API is restricting us to supported APIs, and doesn't even support formats. Sounds we will have to go one step deeper and directly call the docker API

ndeloof:
So, we will need a custom docker inspect command. For this to happen we also need https://github.com/docker-java/docker-java/commit/fe3d2225d7b35be58b0d03d4165bed96d9bb69e7
so we need to wait for a docker-java release, then can implement `Node` support in docker inspect and get the docker node name added to slave display name.

Typz:
Or wait for this patch: https://github.com/docker-java/docker-java/pull/931, which adds the Node field.

ndeloof:
nice one :P

Typz:
docker-java/docker-java#383 has been integrated, which adds the Node field!

ndeloof:
we have to wait for a release so we can rely on this feature from docker-java

ndeloof:
bca9036fb517069165772fcae3996a74114505a1 do now attach full docker inspect to a build.
When Node.name is available in API we could include this useful info in summary

PR #515, opened 2017-10-04
Use same node name as before

Typz:
Node name is <cloud_name>-<containerId>[-<hostName>], with hostname
being added only when running on a swarm.

### DIFF
--- a/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -643,7 +643,7 @@ public class DockerCloud extends Cloud {
 
     }
 
-    /* package */ boolean isSwarm() {
+    public boolean isSwarm() {
         Version remoteVersion = getClient().versionCmd().exec();
         // Cache the return.
         if( _isSwarm == null ) {
--- a/docker-plugin/src/main/java/io/jenkins/docker/DockerSlaveProvisioner.java
+++ b/docker-plugin/src/main/java/io/jenkins/docker/DockerSlaveProvisioner.java
@@ -2,10 +2,12 @@ package io.jenkins.docker;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.CreateContainerCmd;
+import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.command.PullImageCmd;
 import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.model.AuthConfig;
 import com.github.dockerjava.core.command.PullImageResultCallback;
+import com.google.common.base.Strings;
 import com.nirima.jenkins.plugins.docker.DockerCloud;
 import com.nirima.jenkins.plugins.docker.DockerImagePullStrategy;
 import com.nirima.jenkins.plugins.docker.DockerSlave;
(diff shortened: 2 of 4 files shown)


docker-plugin/src/main/java/io/jenkins/docker/DockerSlaveProvisioner.java
```diff
@@ -46,6 +48,21 @@ public DockerSlaveProvisioner(DockerCloud cloud, DockerTemplate template, Docker
...
 
+    public String getDisplayName(String containerId, InspectContainerResponse inspect) {
+        String slaveName = containerId.substring(0, 12);
+        try {
+            slaveName = cloud.getDisplayName() + "-" + slaveName;
+        } catch (Exception ex) {
+            LOGGER.warn("Error fetching cloud name");
+        }
+        if (cloud.isSwarm()) {
+            String hostName = JenkinsUtils.getHostnameFromBinding(inspect);
```
ndeloof: getHostnameFromBinding is some voodoo I'd like to remove from this plugin

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -643,7 +643,7 @@ public boolean equals(Object o) {
 
     }
 
-    /* package */ boolean isSwarm() {
+    public boolean isSwarm() {
```
ndeloof: no way, planned for removal.
swarm standalone should be replace with docker swarm mode as soon as we get some implementation working.

ndeloof:
Can you please describe your needs in more details ? 
Why do you care so much about the agent name ?

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
Typz: Please don't !
Swarm mode is not designed for such use case, and some people (me included) may be happy with classic swarm, with no time/resources to switch over to swarm mode.

Typz:
In an ideal world, I would not care about the name (except it displays 0.0.0.0), and that information should be removed altogether from the name.

In practice, all machines in the swarm are not identical (for exemple, due to "recylcing" of existing machines into the swarm), and there are sometimes build issues which end up being related to the exact machine the build was run on: to debug these issues, the name of the slave is essential, as it is the only thing that persists in the build log after the build has completed (and the effemeral slave has been discarded).

docker-plugin/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
ndeloof: I'm not saying we won't support swarm standalone anymore (it still make sense), just we will avoid specific code for this use case without some explicit added value.

ndeloof:
ok, then what you're looking for is node informations, to be displayed on the /computer/id page. There's many more we can put there, that would be useful for diagnostic

Typz:
The /computer/id page does not really help with effemeral nodes: once the build is complete, the node is removed and the information cannot be accessed anymore.

ndeloof:
right. We could indeed offer some better display name, even I'm not a big fan to use this for such diagnostic. Anyway I'd prefer we don't rely on `getHostnameFromBinding` which is only a guess on external dockerhost IP (there's many reported issue caused by this). Not sure yet how docker API can provide such information

Typz:
I can try to make some changes, but how do you suggest we proceed?

Typz:
@ndeloof : how do you suggest we proceed?

ndeloof:
@Typz need to wait for docker-java release so we can use the swarm Node info and include this in the node description.

ndeloof:
@Typz I just noticed `DockerBuildAction` do keep reference to the DockerHost used to provision the agent, maybe this would be a good place to store docker swarm Node name for further diagnostic when the node is gone

ndeloof:
fixed by bca9036fb517069165772fcae3996a74114505a1

PR #521, opened 2017-10-12
Fixed NPE in DockerSlaveProvisioner.java if registry is not used

mstipanov:
Fixed NPE in DockerSlaveProvisioner.java if registry is not used


ndeloof:
thanks for this !

Issue #533, opened 2017-10-18
SSH Connection is tried too early


gamma:
Please don't use the issue tracker to ask questions, join jenkins-users mailing list.

If you get some troubles with docker-plugin, please report
 - docker-plugin version you use: 1.0.1
 - jenkins version you use: 2.85
 - docker engine version you use:
  - Version = swarm/1.2.6, API Version = 1.22
  - Version = 17.05.0-ce, API Version = 1.29

I tested several setups throughout the day. In the end only the ones worked correctly when the SSHD started very quick. I was able to even solve problems mit supervisord.

But I can not get the Windows Container to work (which they do with version 0.16 of the plugin). Windows Container will be started using the command `powershell -Command C:\\ServiceMonitor.exe sshd` (where ServiceMonitor.exe is a listener to make sure sshd still runs).

It seems that the startup of the daemon takes a little too long for Jenkins and docker-plugin to be accepted as running. When I hit the "Launch Again" button just seconds later, everything works.

There should be some way to make sure that not only the container is running but the SSHD service too. Maybe some kind of automatic connection timeout (10s?) and retry for 5 times ... only then should the container be flagged as not usable (and maybe torn down again).

ndeloof:
fixed by 60d40bfb8a7388d8e4cc6d458d12ea085496932b

PR #536, opened 2017-10-19
Revert missing waitFor for sshd

gamma:
Revert the missing (regression) waitFor function to make sure that the daemon is actually running.

This functionality got lost during the restructuring. See #533.


ndeloof:
Good catch, thanks for heavy testing.
Won't merge this one as I'm currently rewriting DockerComputerSSHLauncher for better UI, but will get this in.

ndeloof:
I reported your changes to new class `io.jenkins.docker.connector.DockerComputerSSHConnector`.
This made me think we could offer a more generic mechanism based on container's HEALTHCHECK so we know if service is ready (from his point of view). This doesn't prevent to run this network connectivity check, that would help diagnostic

Issue #550, opened 2017-11-08
Tunnel connection through is broken


bulanovk:
- [x] docker-plugin version 1.0.4
 - [x] jenkins version 2.86
 - [x] docker engine version  17.09.0-ce

On container start following command are going to be executed:
`java -cp /home/jenkins/remoting-3.13.jar hudson.remoting.jnlp.Main -headless --tunnel 192.168.8.39:50001 -url http://myserver/ 53f0c95ea2e95cc1dab7f37025e873b29473b7f3842e94e949e2fa3d6ed22b23 Node1-4ba115963d83`
but it fails with error 
`"--tunnel" is not a valid option`

according to remoting.jar documentation `-tunnel` option should be used instead of `--tunnel`

```java  
java -jar slave.jar [options...] <secret key> <slave name>
 -agentLog FILE                  : Local agent error log destination (overrides
                                   workDir)
 -cert VAL                       : Specify additional X.509 encoded PEM
                                   certificates to trust when connecting to
                                   Jenkins root URLs. If starting with @ then
                                   the remainder is assumed to be the name of
                                   the certificate file to read.
 -credentials USER:PASSWORD      : HTTP BASIC AUTH header to pass in for making
                                   HTTP requests.
 -failIfWorkDirIsMissing         : Fails the initialization if the requested
                                   workDir or internalDir are missing ('false'
                                   by default)
 -headless                       : Run in headless mode, without GUI
 -internalDir VAL                : Specifies a name of the internal files
                                   within a working directory ('remoting' by
                                   default)
 -jar-cache DIR                  : Cache directory that stores jar files sent
                                   from the master
 -loggingConfig FILE             : Path to the property file with
                                   java.util.logging settings
 -noKeepAlive                    : Disable TCP socket keep alive on connection
                                   to the master.
 -noreconnect                    : If the connection ends, don't retry and just
                                   exit.
 -proxyCredentials USER:PASSWORD : HTTP BASIC AUTH header to pass in for making
                                   HTTP authenticated proxy requests.
 -tunnel HOST:PORT               : Connect to the specified host and port,
                                   instead of connecting directly to Jenkins.
                                   Useful when connection to Hudson needs to be
                                   tunneled. Can be also HOST: or :PORT, in
                                   which case the missing portion will be
                                   auto-configured like the default behavior
 -url URL                        : Specify the Jenkins root URLs to connect to.
 -workDir FILE                   : Declares the working directory of the
[...]

PR #551, opened 2017-11-08
According to remoting.jar documentation -tunnel option should be used…

bulanovk:
According to remoting.jar documentation `-tunnel` option should be used instead of `--tunnel`

Fix for issue https://github.com/jenkinsci/docker-plugin/issues/550


spbkelt:
When are you going to release this?

PR #562, opened 2017-11-30
Use expanded tags and expand dockerfile directory

bschelberg:
Recently, our built images became unable to use $BUILD_NUMBER in their tags. I also wanted to use a build parameter in the docker file directory, which didn't work. This change fixes both of those issues.

I upgraded the TokenMacro dependency to make it easier to expand any variables in dockerFileDirectory. I upgraded to the latest, version 2.3. The method I wanted to use was added in 2.0. I don't know what the convention is with upgrading dependencies - do you typically upgrade to the latest, or to the minimum required to try to avoid dependency issues with installations?

The change to the Test was required because the revision that I pulled from master wouldn't compile without it.


ndeloof:
I've cherry picked your commit to workaround conflict

PR #564, opened 2017-12-06
docker-plugin 1.1 does not support SSH connection to standalone swarm

Typz:
When running against standalone swarm, the SSH server is not available on the same host as the docker end point: the docker endpoint starts a container on any node in the swarm, and the SSH server will only be reachable through that hosts' name/IP address.

This used to work fine, but got broken between 1.04 and 1.1.

(I reproduced the issue with docker 17.09.0-ce and swarm/1.2.8)

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -86,11 +86,6 @@ public class DockerCloud extends Cloud {
     private int containerCap = 100;
 
     /**
-     * Is this cloud actually a swarm?
-     */
-    private transient Boolean _isSwarm;
-
-    /**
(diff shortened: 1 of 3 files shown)


ndeloof:
doesn't swarm standalone report the assigned IP in NetworkSettings ?
```js
"NetworkSettings": {
            "Ports": {
                "22/tcp": [
                    {
                        "HostIp": "<node IP>",
                        "HostPort": "32768"
                    }
                ]
            }
```

if confirmed, I'd prefer to avoid extra args and rely on https://github.com/ndeloof/docker-plugin/commit/c3316ee339faf45e1fa602adefaba6fcd440b08a

ndeloof:
such a pain docker-java does not report `Node` information on inspect, as we would get external IP without much effort :-\

Issue #584, opened 2017-12-15
Unit Tests were broken for LTS 2.89.1


bulanovk:
If you get some troubles with docker-plugin, please report
 - docker-plugin version you use  1.1.1
 - jenkins version you use 2.89.1

All unit test trying to run slave are broken when testing on LTS 2.89.1, because CommandLuncher moved from Jenkins core to CommandLauncher Plugin starting from 2.86 version of Jenkins

ndeloof:
If you run tests by changing jenkins.version property in pom.xml, then this is expected.
Any reason you want to change core requirement ?

bulanovk:
It's requires to pass tests on LTS 2.89.1, because Command Launcher was removed from Jenkins core starting from Jenkins 2.86. By default docker plugin are tested againsts 2.60, 2.73 and 2.89 Jenkins version.

If we ommit  this changes, we will got false negative unit test results and error on PR

ndeloof:
right, but docker-plugin does not rely on CommandLauncher ...
So, shouldn't this be transparent for plugin ? What I mean is : it makes sense to test plugin on various LTS, but this should not impact plugin (or all of them will need to add this hack !). Shouldn't this be managed by plugin parent pom.xml (or `buildPlugin` shared library) ? 



I'm also confused by the profile configured by `jenkins.version != any`. What's the intent here ?

bulanovk:
Yes, in right way it should come from some of transitive, and now it just WA to got Green test while it will be fixed some where it realy reuqires this plugin.

I'm also confused by the profile configured by jenkins.version != any. What's the intent here ?
it just maven WA to got multiple maven profile activated in the same time, to avoid duplicate depends

ndeloof:
Sorry still don't get it. If multiple profiles are enable, dependencies will just get merges in effective pom.

The "jenkins_after_2_86" looks good as a temporary workaround. Would better test jenkins.version >= 2.86 but I guess this is the best maven let us do (sic).

bulanovk:
Maven documentation said:
```
This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated
```

So if we just left <activeByDefault>true</activeByDefault> it will be diactivated when second profile will be triggered by a condition. To avoid such case there is a WA you need to specify any condition that alway will be true.

[...]

PR #585, opened 2017-12-15
Unit Tests were broken for LTS 2.89.1

bulanovk:
Introduce default profile to test with Jenkins prior 2.86
Introduce jenkins_after_2_86 profile to test with Jenkins 2.86+

Fix for https://github.com/jenkinsci/docker-plugin/issues/584


ndeloof:
partially merged your patch as https://github.com/jenkinsci/docker-plugin/commit/cf206f152b11a06367e229a07f50f9e150ef0c44

ndeloof:
thanks a lot for fixing this !

PR #596, opened 2017-12-28
New Option for JNLP: Pass Slave Connection Arguments to Entrypoint

odavid:
Hi,

I am using the docker jnlp slave to to build docker images by attaching the /var/run/docker.sock to the slave container. 
In order for it to work, I created an [image](https://github.com/odavid/jenkins-jnlp-slave) derived from [jenkins/jnlp-slave](https://github.com/jenkinsci/docker-jnlp-slave). 

Basically the user that runs the entrypoint is 'root' and during startup, it checks for the GID that owns the /var/run/docker.sock on the host. Then, it changes the 'jenkins' user to be member of that group.  

```bash
# To enable docker cloud based on docker socket,
# we need to add jenkins user to the docker group
if [ -S /var/run/docker.sock ]; then
  DOCKER_SOCKET_OWNER_GROUP_ID=$(stat -c %g /var/run/docker.sock)
  groupadd -for -g ${DOCKER_SOCKET_OWNER_GROUP_ID} docker
  id jenkins -G -n | grep docker || usermod -aG docker jenkins
fi
```    

At the end, the entrypoint replaces the user to be 'jenkins' and passes the arguments it got back to the original entrypoint
```bash
...
exec gosu jenkins "jenkins-slave" "$@"
``` 

It works very well with [kubernetes-plugin](https://github.com/jenkinsci/kubernetes-plugin/) and [amazon-ecs-plugin](https://github.com/jenkinsci/amazon-ecs-plugin/), however, with docker-plugin it does not work in lot of situations.

After investigation, I noticed that the problem is the way the docker-plugin runs the entrypoint with /bin/sh in detached mode and then exec the slave.jar. 
Although I understand the reasoning behind that - (i.e to be able to inject the slave.jar), it causes the 'jenkins' user to start the java process before it was a member of the docker group.

Pseudo Example:
```bash
docker run -t -d -v /var/run/docker.sock:/var/run/docker.sock --name my-jenkins-slave odavid/jenkins-jnlp-slave /bin/sh
docker exec -t -u jenkins my-jenkins-slave bash -c id

# can result (depends on how fast it ran...)
# uid=10000(jenkins) gid=10000(jenkins) groups=10000(jenkins)

docker run -t -d --name my-jenkins-slave -v /var/run/docker.sock:/var/run/docker.sock odavid/jenkins-jnlp-slave bash -c id
# will always result
# uid=10000(jenkins) gid=10000(jenkins) groups=50(docker),10000(jenkins)

```

In order to solve that, I wanted to have an additional option (while keeping the current behaviour as the default one): 
* __Pass Slave Connection Arguments to Entrypoint__  - When this option is enabled, the plugin will only run the container without any "docker exec" processes and pass the slave.jar arguments to the entrypoint. Using this way can help people do some initial preparation within the entrypoint ___before___ running the slave.jar. This also simplifies the execution and make it aligned with with [kubernetes-plugin](https://github.com/jenkinsci/kubernetes-plugin/) and [amazon-ecs-plugin](https://github.com/jenkinsci/amazon-ecs-plugin/).

In order to implement this feature, I had to re-factor the code in the following steps:
* Make the ability for the Launcher to start a container and not just get a container Id
* Move the beforeContainerCreated(), beforeContainerStarted(), afterContainerStarted() to a dedicated DockerContainerLifecycleHandler class, so it will be used as a callback
* Move the docker container start to a dedicated class: i.e. DockerContainerExecuter
* Enable the CreateContainerCmd to be manipulated with a reference to the SlaveComputer ___before___ the container starts, so I can pass the computer.JnlpMac and computer.name to the entrypoint
* In order to keep everything work with Serialization, I needed also to pass the template to the createLauncher and use DockerTemplate.createContainerCmd within the launcher.launch()
* Since the containerId is not available during the laucher object creation, I had to change the unique slave name to a different unique

I will be very happy if I can this PR approved, since currently I have only 2 options:
* Use root user instead of jenkins - this makes it hard with cache volumes. Also I don't really like root to be the process owner.
* Use jenkins-swarm or other static slaves (There are several places I work who do not have kubernetes or ECS) 

I would really appericiate if you can approve this PR.

Best,
Ohad

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -4,7 +4,6 @@ import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.CreateContainerCmd;
 import com.github.dockerjava.api.command.PullImageCmd;
 import com.github.dockerjava.api.exception.DockerClientException;
-import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.model.PortBinding;
 import com.github.dockerjava.api.model.PullResponseItem;
@@ -15,19 +14,13 @@ import com.nirima.jenkins.plugins.docker.launcher.DockerComputerLauncher;
 import com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy;
 import hudson.Extension;
 import hudson.Util;
-import hudson.model.Describable;
-import hudson.model.Descriptor;
-import hudson.model.DescriptorVisibilityFilter;
-import hudson.model.Label;
-import hudson.model.Node;
-import hudson.model.TaskListener;
+import hudson.model.*;
 import hudson.model.labels.LabelAtom;
 import hudson.slaves.ComputerLauncher;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.NodePropertyDescriptor;
 import hudson.slaves.RetentionStrategy;
 import hudson.util.DescribableList;
-import hudson.util.FormValidation;
 import io.jenkins.docker.DockerTransientNode;
 import io.jenkins.docker.client.DockerAPI;
 import io.jenkins.docker.connector.DockerComputerConnector;
@@ -37,19 +30,14 @@ import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
-import org.kohsuke.stapler.QueryParameter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.io.Serializable;
+import java.util.*;
 
 
 public class DockerTemplate implements Describable<DockerTemplate> {
@@ -412,35 +400,40 @@ public class DockerTemplate implements Describable<DockerTemplate> {
 
     }
 
+    public class ContainerCommandCreator implements Serializable{
+        private final String containerUniqueName;
+
+        ContainerCommandCreator(){
+            // Since container can now be launched during slave launch, we need an alternative unique node name, which will be also the container name
+            this.containerUniqueName = Long.toHexString(System.nanoTime());
+        }
+
+        public String getContainerUniqueName() {
+            return containerUniqueName;
+        }
+
+        public CreateContainerCmd createContainerCmd(DockerAPI api){
+            final DockerClient client = api.getClient();
+            CreateContainerCmd cmd = client.createContainerCmd(getImage());
+            fillContainerConfig(cmd);
+            cmd.withName(containerUniqueName);
+            return cmd;
+        }
+
+    }
+
     @Restricted(NoExternalUse.class)
     public DockerTransientNode provisionNode(DockerAPI api, TaskListener listener) throws IOException, Descriptor.FormException, InterruptedException {
 
-        final DockerClient client = api.getClient();
         final DockerComputerConnector connector = getConnector();
         pullImage(api, listener);
 
         LOGGER.info("Trying to run container for {}", getImage());
-        CreateContainerCmd cmd = client.createContainerCmd(getImage());
-        fillContainerConfig(cmd);
+        ContainerCommandCreator containerCommandCreator = new ContainerCommandCreator();
+        LOGGER.info("Trying to run container for image: {}, with unique name: {}", getImage(), containerCommandCreator.getContainerUniqueName());
+        final ComputerLauncher launcher = connector.createLauncher(api, containerCommandCreator, remoteFs, listener);
 
-        connector.beforeContainerCreated(api, remoteFs, cmd);
-
-        String containerId = cmd.exec().getId();
-
-        try {
-            connector.beforeContainerStarted(api, remoteFs, containerId);
-
-            client.startContainerCmd(containerId).exec();
-
-            connector.afterContainerStarted(api, remoteFs, containerId);
-        } catch (DockerException e) {
-            // if something went wrong, cleanup aborted container
-            client.removeContainerCmd(containerId).withForce(true).exec();
-            throw e;
-        }
-
-        final ComputerLauncher launcher = connector.createLauncher(api, containerId, remoteFs, listener);
-        DockerTransientNode node = new DockerTransientNode(containerId, remoteFs, launcher);
+        DockerTransientNode node = new DockerTransientNode(containerCommandCreator.getContainerUniqueName(), remoteFs, launcher);
         node.setNodeDescription("Docker Agent [" + getImage() + " on "+ api.getDockerHost().getUri() + "]");
         node.setMode(mode);
         node.setLabelString(labelString);
--- a/src/main/java/io/jenkins/docker/DockerTransientNode.java
+++ b/src/main/java/io/jenkins/docker/DockerTransientNode.java
@@ -7,7 +7,6 @@ import com.nirima.jenkins.plugins.docker.DockerOfflineCause;
 import com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy;
 import hudson.model.Computer;
 import hudson.model.Descriptor;
-import hudson.model.Queue;
 import hudson.model.Slave;
 import hudson.model.TaskListener;
 import hudson.slaves.Cloud;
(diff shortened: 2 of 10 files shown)


ndeloof:
The reason launcher is executed in an exec is not to copy the remoting.jar (this can happen before container is started) but to ensure we can reconnect after jenkins is restarted, typically within a pipeline.

I personally dislike this approach for two reasons: 
1. using gosu require to initially run container as root
1. exposing /var/run/docker.sock inside container is a major security issue for underlying docker infrastructure.

That being said I understand the intent here.

src/main/resources/io/jenkins/docker/connector/DockerComputerJNLPConnector/config.jelly
```diff
@@ -10,5 +10,14 @@
         <f:textbox/>
     </f:entry>
 
+    <f:entry title="${%Jenkins URL}" field="jenkinsUrl">
```
ndeloof: I don't get the goal of this parameter. Jenkins external URL can be configured globaly, and tunnel option is designed to cover reverse-proxy/firewall constraints

ndeloof:
maybe we could re-implement https://github.com/moby/moby/blob/master/pkg/namesgenerator/names-generator.go in java

src/main/resources/io/jenkins/docker/connector/DockerComputerJNLPConnector/config.jelly
odavid: Hi @ndeloof,
I've added that parameter (actually just exposed it, since it was there before for the sake of Unit Tests), just because it existed in both kubernetes-plugin and aws-ecs-plugin, so I thought it would be nice as standard way to override the current Jenkins URL. 
It is not related to this pull request, so I can remove it if you think it is not needed. 
As a side comment, sometime you don't have a real DNS name, but just an IP and you don't want it to be defined in the global JENKINS configuration, since it is affecting other plugins. IMHO, have this parameter as an optional one would be nice, but again, I can remove it if think it should be removed.

odavid:
@ndeloof, I believe that hex value of system.nanotime would be enough here and will serves us well.
In any case, names-generator can be an overkill for this purpose. I do think that we can add the template name as a prefix for the slave name.

src/main/java/io/jenkins/docker/DockerTransientNode.java
```diff
@@ -7,44 +7,51 @@
...
 /**
  * A {@link Slave} node designed to be used only once for a build.
  * 
  * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
  */
 public class DockerTransientNode extends Slave {
-    
-    private final String containerId;
-    
+
```
ndeloof: please do not remove this useful information. Just make this non-final and assign as container has been created

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -412,35 +409,25 @@ public void onNext(PullResponseItem item) {
...
-            connector.beforeContainerStarted(api, remoteFs, containerId);
-
-            client.startContainerCmd(containerId).exec();
-
-            connector.afterContainerStarted(api, remoteFs, containerId);
-        } catch (DockerException e) {
-            // if something went wrong, cleanup aborted container
-            client.removeContainerCmd(containerId).withForce(true).exec();
-            throw e;
-        }
```
ndeloof: You don't need to move this code away and introduce `DockerContainerLifecycleHandler`(which makes DockerComputerConnector role unclear). Just compute `containerName`, create `DockerTransientNode ` object, then enforce here `CreateContainerCmd.name` is set to this unique ID and pass node object to connector methods for launcher setup.

src/main/java/io/jenkins/docker/DockerTransientNode.java
```diff
@@ -7,44 +7,51 @@
...
 
     private boolean removeVolumes;
 
     private String cloudId;
 
-    public DockerTransientNode(@Nonnull String containerId, String remoteFS, ComputerLauncher launcher) throws Descriptor.FormException, IOException {
-        super("docker-" + containerId.substring(0,12), remoteFS, launcher);
+
+    public DockerTransientNode(@Nonnull String uniqueName, String workdir, ComputerLauncher launcher) throws Descriptor.FormException, IOException {
+        super("docker-" + uniqueName, workdir, launcher);
```
ndeloof: wouldn't hurt to also store containerName as an attribute

src/main/java/io/jenkins/docker/connector/DockerComputerJNLPConnector.java
```diff
@@ -76,46 +101,80 @@ public boolean isLaunchSupported() {
...
+                            final DockerClient client = api.getClient();
+                            injectRemotingJar(containerId, workdir, client);
+                        }
+                    }
+                };
+                CreateContainerCmd cmd = template.createContainerCmd(api);
+                if(!passSlaveConnectionArgs){
+                    launchAndInjectJar(computer, listener, handler, api, containerExecuter, workdir, cmd);
+                }else{
+                    launchStandardJNLPImage(computer, listener, handler, api, containerExecuter, workdir, cmd);
```
ndeloof: just replace the exec-based implementation with `java remoting.jar` set as main process in container

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
odavid: I moved this code to the DockerComputerConnector, since it must be called from the launcher itself and cannot run before the launcher is being created. In this iteration, I tried to keep the original code as it was when I could (sorry for the many refactoring - did not try to make your review hard, it was just multiple iterations).
Also removed the DockerContainerLifecycleHandler and restored the original methods inside the DockerComputerConnector)

src/main/java/io/jenkins/docker/DockerTransientNode.java
odavid: Restored.

src/main/java/io/jenkins/docker/DockerTransientNode.java
odavid: Done.

src/main/resources/io/jenkins/docker/connector/DockerComputerJNLPConnector/config.jelly
```diff
@@ -10,5 +10,14 @@
...
+        <f:textbox/>
+    </f:entry>
+
+    <f:entry title="${%Pass Slave Connection Arguments to Entrypoint}" field="passSlaveConnectionArgs">
```
ndeloof: not needed,  just make this the default behaviour for JNLP connector

src/main/resources/io/jenkins/docker/connector/DockerComputerJNLPConnector/help-passSlaveConnectionArgs.html
```diff
@@ -0,0 +1,17 @@
+<div>
+    <p>
+        <span class="info">Prerequisites:</span> Docker image must derive from <a href="https://github.com/jenkinsci/docker-jnlp-slave/">jenkins/jnlp-slave</a>
```
ndeloof: wrong. This isn't a MUST requirement. Just need java available on arbitrary image.
(also not required assuming you make "passSlaveConnectionArgs" the standard behaviour, without such an option)

src/main/java/io/jenkins/docker/connector/DockerComputerJNLPConnector.java
```diff
@@ -34,6 +37,7 @@
     private String user;
     private final JNLPLauncher jnlpLauncher;
     private String jenkinsUrl;
+    private boolean passSlaveConnectionArgs;
```
ndeloof: not necessary, just make this the standard behaviour

ndeloof:
please drop "passSlaveConnectionArgs" option to just make this the default mechanism, and please also squash your commits so we can have a more readable history for those changes.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -412,35 +400,40 @@ public void onNext(PullResponseItem item) {
 
     }
 
-    @Restricted(NoExternalUse.class)
-    public DockerTransientNode provisionNode(DockerAPI api, TaskListener listener) throws IOException, Descriptor.FormException, InterruptedException {
+    public class ContainerCommandCreator implements Serializable{
```
odavid: Hi @ndeloof, I had to add this ```ContainerCommandCreator```, since the CreateContainerCmd is not Serializable and therefore when passing it to the anonymous ```DelegatingComputerLauncher(new JNLPLauncher()))``` through the ```createLauncher``` method, Jenkins will fail serialization. 
So the trick was to do it lazily using the ```DockerTemplate``` which is serializable. So this class is a holder for the uniqueContainerName and when needed, the it creates the CreateContainerCmd and fills it will all the required data from the DockerTemplate instance.

odavid:
Hi,
I've dropped the passSlaveConnectionArgs and made it the default one.
Also squashed my commits into one.
Best regards,
Ohad

odavid:
Thank you!

Issue #594, opened 2017-12-21
Jenkins gets stuck when there are no resources to schedule a container


gamma:
If you get some troubles with docker-plugin, please report
 - [ x ] docker-plugin version you use: **1.1+**
 - [ x ] jenkins version you use: **2.96**
 - [ x ] docker engine version you use: **Version = swarm/1.2.6, API Version = 1.22**
 - [ ] stack trace / logs / any technical details that could help diagnose this issue

I have my docker cloud images configured to only spawn 5 to 10 instances max. There are several jobs that start 5+ jobs at once. The first jobs get started until the limit is reached and Jenkins would wait for resources being available.

Starting with docker-plugin 1.1 this does not work as expected. The jobs get stuck. Even stopping and starting new jobs manually does not work, only a restart of Jenkins.

MatthewChiappa:
I'm experiencing the same issue with with the same setup only with Jenkins at version 2.73. Every day I can slowly see the number of jobs stuck in the queue. The only way to get the jobs to run is to kick off more builds and it gets forced out of the queue. The only cure to this is to restart Jenkins, but after a day or so the problem keeps happening. We also run 20+ jobs at a time at some points, which hits our container cap.

ndeloof:
I still don't fully understand the NodeProvisioned logic and how to get eventually rid of such issues.

pjdarton:
@ndeloof My understanding is that Jenkins' handling of all this is rather basic (possibly _too_ basic).  All responsibility for making it work falls upon Cloud's [provision(Label label, int excessWorkload)](https://github.com/jenkinsci/docker-plugin/blob/master/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java#L234) method.

This method will be called, repeatedly, as long as Jenkins has jobs in the queue. e.g. if Jenkins has a job queued that needs the label "unix" then provision will be called and expected to return a PlannedNode that'll (eventually) give a Node that's got the "unix" label that can be used to run the job.  Actual provisioning runs in [another thread](https://github.com/jenkinsci/docker-plugin/blob/master/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java#L263) but eventually gives Jenkins a new Node, that'll then run one or more jobs and reduce demand.

[...]

PR #610, opened 2018-01-23
Enhancement: (re-)Add "Read Timeout" to DockerCloud.

pjdarton:
It's possible for a docker API endpoint to seize up, accepting incoming connections but never replying.  This causes docker-plugin's calls to the docker API to hang, causing Jenkins' call to the docker-plugin to hang, causing Jenkins' cloud functionality to hang, crippling Jenkins.
To handle this, the docker-plugin must timeout so that it won't wait forever for the docker-api to return, and treat a timeout as a failure.  This will help mitigate #594.
This code change re-introduces the functionality as follows:
 - Re-add readTimeout to the docker cloud configuration
 - Re-add readTimeout to the docker API
 - Enhance docker-java code to support readTimeout.
 - Improve help text for both connectTimeout and readTimeout.

Note: This functionality was present in older versions of the plugin but was lost due to later versions of the docker-java library missing this functionality.  There's a separate issue to put that back again.  Once the official docker-java release is enhanced to support readTimeout, this plugin can remove its own enhanced NettyDockerCmdFactory and go back to using the docker-java version directly.

### DIFF
--- a/src/main/java/io/jenkins/docker/client/DockerAPI.java
+++ b/src/main/java/io/jenkins/docker/client/DockerAPI.java
@@ -3,12 +3,13 @@ package io.jenkins.docker.client;
 import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.command.VersionCmd;
 import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.model.Version;
 import com.github.dockerjava.core.DefaultDockerClientConfig;
 import com.github.dockerjava.core.DockerClientBuilder;
 import com.github.dockerjava.core.SSLConfig;
-import com.github.dockerjava.netty.NettyDockerCmdExecFactory;
+
 import hudson.Extension;
 import hudson.model.AbstractDescribableImpl;
 import hudson.model.Descriptor;
@@ -55,7 +56,11 @@ public class DockerAPI extends AbstractDescribableImpl<DockerAPI> implements Ser
 
     private DockerServerEndpoint dockerHost;
 
+    /** Connection timeout in seconds */
     private int connectTimeout;
+
+    /** Read timeout in seconds */
+    private int readTimeout;
 
     private String apiVersion;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/jenkins/docker/client/DockerAPI.java
```diff
@@ -122,21 +137,24 @@ public boolean isSwarm() {
...
         if (client == null) {
+            final Integer readTimeoutInMillisecondsOrNull = readTimeout > 0 ? Integer.valueOf(readTimeout * 1000) : null;
+            final Integer connectTimeoutInMillisecondsOrNull = connectTimeout > 0 ? Integer.valueOf(connectTimeout * 1000) : null;
             client = DockerClientBuilder.getInstance(
                 new DefaultDockerClientConfig.Builder()
                     .withDockerHost(dockerHost.getUri())
                     .withCustomSslConfig(toSSlConfig(dockerHost.getCredentialsId()))
                 )
                 .withDockerCmdExecFactory(new NettyDockerCmdExecFactory()
-                .withConnectTimeout(connectTimeout))
```
pjdarton: Note: I'm pretty sure that this old code was buggy - withConnectTimeout expects to be given milliseconds (or null for "no timeout"), but connectTimeout is in seconds.

src/main/java/io/jenkins/docker/client/DockerAPI.java
```diff
@@ -205,24 +224,35 @@ public ListBoxModel doFillCredentialsIdItems(@AncestorInPath ItemGroup context,
...
-
-                return FormValidation.ok("Version = " + verResult.getVersion() + ", API Version = " + verResult.getApiVersion());
+                final DockerServerEndpoint dsep = new DockerServerEndpoint(uri, credentialsId);
+                final DockerAPI dapi = new DockerAPI(dsep, connectTimeout, readTimeout, apiVersion, null);
+                final DockerClient dc = dapi.getClient();
+                final VersionCmd vc = dc.versionCmd();
+                final Version v = vc.exec();
+                final String actualVersion = v.getVersion();
+                final String actualApiVersion = v.getApiVersion();
+                return FormValidation.ok("Version = " + actualVersion + ", API Version = " + actualApiVersion);
```
pjdarton: Splitting the test code up so that we only get one method call per line means that, in the event of a failure, the exception trace will tell us exactly where the problem came from.
e.g. If everything is on one line and we get a NullPointerException then it's impossible to see which bit failed.

PR #616, opened 2018-02-09
Per-cloud instance caps

pjdarton:
1) BugFix: if the plugin is talking to a very busy docker daemon, it can get more than 1meg of data back from the ListContainersCmd and throw an exception, which stops it from starting further containers.
This code changes the parameters we pass to ListContainers so that we only ask about containers that we started ourselves, thus excluding any other containers from the returned results.
2) BugFix: if the plugin has multiple clouds mentioning the same image, it was possible for it to get confused when counting the number of containers being constructed.  This change ensures that each cloud has its own counters so there's no longer pollution between them.
3) BugFix: the code responsible for tracking the count of containers in progress was able to loose track.  We now use try/finally code to ensure it cannot get out of step even if exceptions happen.
4) Behavioral change:  cloud-level instance caps now only count instances started by "our" Jenkins and no longer count containers running within docker that weren't started by our Jenkins.
5) Performance improvement:  If you don't set a template container instance cap, we don't ask docker to count them.  If you don't set a cloud-level container instance cap, we don't ask docker to count them.  If you don't set either cap then we don't hammer the "ListContainers" API at all, resulting in improved speed and stability of the Jenkins cloud provisioning process overall.
6) Logging changes:  DockerCloud's "Will provision ..." log messages now provide more information, showing current counts and caps, when those caps are being used.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -10,7 +10,6 @@ import com.github.dockerjava.api.command.PullImageCmd;
 import com.github.dockerjava.api.command.PushImageCmd;
 import com.github.dockerjava.api.command.StartContainerCmd;
 import com.github.dockerjava.api.model.AuthConfig;
-import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.api.model.Version;
 import com.google.common.base.Throwables;
 import hudson.Extension;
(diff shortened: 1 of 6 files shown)


pjdarton:
@ndeloof I've been running this for a few days now, it seems to work quite well.
I would recommend that it be merged (along with #617) and we do a release.

I've also come to the conclusion that the docker-java library's support of Netty is buggy - it's apparent that Netty isn't a priority for docker-java and that it isn't tested prior to release - every time I try to use functionality we've not used before, I find yet another bug.  I suspect that we (the docker-plugin) are the only user of its Netty functionality and I therefore question whether or not it's worthwhile raising bug reports on it.
 - You mentioned before that you'd like to swap to another docker library:  I'm coming to the conclusion that this is a good idea!

Issue #615, opened 2018-02-08
Test Connection button leaking file descriptors


JordanGlassley:
- [1.1.2] docker-plugin version you use
 - [2.89.3] jenkins version you use
 - [17.05.0-ce] docker engine version you use

There seems to be consistently leaking file descriptors when pressing the "Test Connection" button on the Configure System page.

![final-button](https://user-images.githubusercontent.com/36273203/35995951-99346254-0cda-11e8-959d-f398055074d2.png)

### Effects
There were occasional connection issues with our docker API endpoint, so in spamming the test connection button, we ended up reaching the max open file limit for our system. Here is a graph of open file descriptors during that period of time.

![final-graph](https://user-images.githubusercontent.com/36273203/35994599-7f39d590-0cd6-11e8-8714-55b65c5273c6.png)

### Leaking file descriptors
Each time the button is pressed, there is a consistent increase in the number of file descriptors open on the system. Each button press increased the [eventpoll] fd by 4 and the pipe fd by 8. These file descriptors don't seem to ever be reclaimed until restarting the Jenkins instance.

```
# before button press
[vagrant@master0 ~]$ sudo lsof -p 15090 | awk -F' ' '{print $NF}' | sort | uniq -c
     ...
      7 [eventpoll]
     14 pipe
     ...
[vagrant@master0 ~]$ sudo lsof -p 15090 | awk -F' ' '{print $NF}' | wc -l
1401

# after button press
[vagrant@master0 ~]$ sudo lsof -p 15090 | awk -F' ' '{print $NF}' | sort | uniq -c
     ...
     11 [eventpoll]
     22 pipe
     ...
[vagrant@master0 ~]$ sudo lsof -p 15090 | awk -F' ' '{print $NF}' | wc -l
1413
```

### Logs/Stacktraces
I didn't see any exceptions or stacktraces for the following patterns around this situation. This was the only logging for the button press.

```
# patterns
io.jenkins.docker
com.nirima
com.github.dockerjava

# logging
Feb 08, 2018 2:06:02 PM FINE com.github.dockerjava.core.command.AbstrDockerCmd exec
Cmd: com.github.dockerjava.netty.exec.VersionCmdExec@73b7073d
Feb 08, 2018 2:06:02 PM FINEST com.github.dockerjava.netty.exec.VersionCmdExec execute
GET: com.github.dockerjava.netty.WebTarget@16c0441b
```

pjdarton:
[...]

PR #617, opened 2018-02-09
Bugfix for issue #615

pjdarton:
Fix for #615 
The docker cloud "Test Connection" button causes socket resources to be leaked.  The doTestConnection method in DockerAPI creates a new DockerClient and then forgets about it, leaving it open.
This change adds a finally clause that closes the connection after the test to ensure that it doesn't leak so easily.


pjdarton:
Note that this fix was tested by JordanGlassley (who raised it originally) and it looks like this fixes it.

Issue #614, opened 2018-02-07
UnsupportedOperationException: Refusing to marshal com.github.dockerjava.api.command.InspectContainerResponse


wumpz:
Seems to be a problem due to JEP 200. I use the latest version of Jenkins and this plugin.

```
java.io.IOException: java.lang.RuntimeException: Failed to serialize hudson.model.Slave#launcher for class io.jenkins.docker.DockerTransientNode
jenkins_1    |  at hudson.XmlFile.write(XmlFile.java:200)
jenkins_1    |  at jenkins.model.Nodes.save(Nodes.java:274)
jenkins_1    |  at hudson.util.PersistedList.onModified(PersistedList.java:173)
jenkins_1    |  at hudson.util.PersistedList.replaceBy(PersistedList.java:85)
jenkins_1    |  at hudson.model.Slave.setNodeProperties(Slave.java:299)
jenkins_1    |  at com.nirima.jenkins.plugins.docker.DockerTemplate.provisionNode(DockerTemplate.java:448)
jenkins_1    |  at com.nirima.jenkins.plugins.docker.DockerCloud$1.run(DockerCloud.java:268)
jenkins_1    |  at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
jenkins_1    |  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
jenkins_1    |  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
jenkins_1    |  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
jenkins_1    |  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
jenkins_1    |  at java.lang.Thread.run(Thread.java:748)
jenkins_1    | Caused by: java.lang.RuntimeException: Failed to serialize hudson.model.Slave#launcher for class io.jenkins.docker.DockerTransientNode
jenkins_1    |  at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:256)
jenkins_1    |  at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:224)
jenkins_1    |  at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:138)
[...]

PR #619, opened 2018-02-14
Add Class filter to support JEP-200 whitelist

bulanovk:
Add whitelisted class for JEP-200

Fix for issue https://github.com/jenkinsci/docker-plugin/issues/614

### DIFF
--- a/src/main/resources/META-INF/hudson.remoting.ClassFilter
+++ b/src/main/resources/META-INF/hudson.remoting.ClassFilter
@@ -1,0 +1,14 @@
+com.github.dockerjava.api.command.InspectContainerResponse
+com.github.dockerjava.api.model.ContainerConfig
+com.github.dockerjava.api.model.HostConfig
+com.github.dockerjava.api.model.Binds
+com.github.dockerjava.api.model.Bind
+com.github.dockerjava.api.model.Volume
+com.github.dockerjava.api.model.LogConfig
+com.github.dockerjava.api.model.Ports
+com.github.dockerjava.api.model.RestartPolicy
+com.github.dockerjava.api.model.NetworkSettings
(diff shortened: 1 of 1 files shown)


sadovnikov:
Hi, Are there plans for releasing new version of the plugin with this fix? 
We use docker-plugin-1.1.3, however it's released a week too early :)

pjdarton:
You can download the latest docker-plugin.hpi file from https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/master/ and then upload that into your Jenkins via "Manage Jenkins -> Plugins -> Advanced".

sadovnikov:
Thank you @pjdarton for your suggestion. However we are using a number of Jenkins Masters and distribute plugins' updates through Operations Center. We do need a released plugin version

oleg-nenashev:
`com.github.dockerjava.api.command.InspectContainerResponse$Node` is not whitelisted. The serialization may fail if the field is not null

pjdarton:
I don't think any class in the docker-plugin that needs serialization
should possess any fields from the docker-java library.
Anything in need of serialization should be extracted out of docker-java
classes.
The docker-plugin's use of docker-java should be purely "a means to an end"
rather than appearing in its API.

It's a public holiday in the UK so I don't have my computer to hand (hence
the vague reply) but I think that any fix for this should be to make all
use of docker-java be in local variables (or, if really necessary,
transient fields). Any returned data that we need to transport across
serialization should be extracted from the docker-java classes.

ndeloof:
I fully agree none of the dockerjava classes should be part of plugin's serialization.

oleg-nenashev:
There are also some discussion in https://issues.jenkins-ci.org/browse/JENKINS-50480

PR #622, opened 2018-02-22
Prevent over-provisioning

pjdarton:
Jenkins can keep on asking for more executors even though it has enough slaves promised to it that will satisfy demand once they come online.
We already keep track of the number of slaves that we are in the process of creating but have yet to give to Jenkins, so if we include that count when determining demand, we can ensure we do not provision more
containers than Jenkins really wanted.

Note: This is mostly useful if the docker host is being slow at creating containers, causing Jenkins to ask again before we've created the last bunch.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -288,6 +288,22 @@ public class DockerCloud extends Cloud {
             final List<DockerTemplate> templates = getTemplates(label);
             int remainingWorkload = numberOfExecutorsRequired;
 
+            // Take account of the executors that will result from the containers which we
+            // are already committed to starting but which have yet to be given to Jenkins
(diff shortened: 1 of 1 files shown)


pjdarton:
Personally I wasn't going to press the "merge" button on this until I was sure it was a good change, but I guess as long as we don't release it until we can be _sure_ that it's all good then that'll suffice.

ndeloof:
`git revert` is your best friend if something sounds weird during pre-release testing :)

pjdarton:
OK, I'll take your word for it - personally I'm not hugely comfortably with git beyond "commit" and "push" :-)

FYI I've added a few RFEs (#624, #625, #626) that I intend to start work sometime next month (probably in that order).  If you have any opinions on them then please do share - I don't want to take things in a direction you disapprove of, and I'd also really value having a second opinion - it's all too easy to get "tunnel vision" and end up going down the wrong path sometimes...

PR #627, opened 2018-02-26
manage pipeline resumability

ndeloof:
re-run `start` on pipeline resume if node hasn't been created yet


src/main/java/io/jenkins/docker/DockerTransientNode.java
```diff
@@ -35,6 +36,8 @@
 
     private String cloudId;
 
+    private AtomicBoolean acceptingTasks = new AtomicBoolean(true);
```
jglick: I think a simple `boolean` (perhaps `volatile`) would suffice here.

ndeloof: a `volatile boolean` would then be equivalent

PR #632, opened 2018-03-08
More info in DockerComputerSSHConnector

HugoY:
Display more information in the Prerequisites section of
DockerComputerSSHConnector.

Only sshd is shown as mandatory but JDK is also mandatory

Note that this info is available on the plugin page
https://plugins.jenkins.io/docker-plugin
(Section "Creating a docker image">"Launch via SSH")

### DIFF
--- a/src/main/resources/io/jenkins/docker/connector/DockerComputerSSHConnector/config.jelly
+++ b/src/main/resources/io/jenkins/docker/connector/DockerComputerSSHConnector/config.jelly
@@ -2,7 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" >
 
     <f:block>
-        <span class="info">Prerequisites:</span> Docker image must have <a href="https://www.openssh.com/">sshd</a> installed.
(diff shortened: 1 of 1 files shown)


pjdarton:
Are you 100% sure of this?
The SSH connection plugin (that the docker plugin calls on to do this) contains functionality to locate a JVM and, if it can't find one, can attempt to download one automatically onto the target.
If you ensure that you've got a long idle timeout set so the container stays around long enough for you to go onto the GUI and see the node's connection log, you can see it looking for "java" and then, if it doesn't find it, trying to download one suitable for that OS.

HugoY:
Not 100% sure.
But it's clearly stated in the doc that a JDK is needed. 
untitled
The fact that only sshd is mentioned on the "Configure System" page in Jenkins mislead me (no auto download for me, maybe idle timeout too short IDK)

pjdarton:
OK, I'm happy to make the help in the UI consistent with the docs.
I'm not 100% convinced that the docs (and now the UI help) are correct _but_ it would be silly to create a docker image without a JVM as that'd add a large delay to container startup if it had to download and install one every time - it's certainly _sensible_ (and common practice) to put everything you need into the container.

So, right or wrong, your change to the UI help will help folks to follow best practice :-)

PR #633, opened 2018-03-08
[REVIEW REQUESTED] Improve readtimeout functionality

pjdarton:
This PR provides improved readTimeout functionality and connection caching, implementing #624.

Docker connection readTimeouts no longer apply to docker pull and docker build commands.
Docker build build-steps no longer have a timeout - builds have their own timeout functionality.
Docker templates now have their own pullTimeout field that's used as the max time for the docker pull to complete.
Deprecated DockerCloud.getClient() in favor of DockerCloud.getDockerApi().takeClient() and .releaseClient().  Refactored code so that all plugin code now calls DockerAPI.takeClient()
and finally { releaseClient }.

Issue #626, opened 2018-02-22
RFE: Avoid using broken templates/clouds


pjdarton:
I've found that docker hosts crash and burn - sometimes they just stop accepting connections, sometimes they accept connections but never answer (hence the need for a readTimeout), sometimes they answer but always fail.
I've also found that it's dead easy to enter in a template that doesn't work and always fails to spin up, e.g. the image can't be pulled.

In both cases, the docker plugin will blindly ignore the failures and continue to try to spin up containers, and will continue to tell Jenkins "yes, I can do this" (i.e. ``DockerCloud.canProvision(Label)``` will return true) even when it's doomed to failure, and this means that Jenkins doesn't ask any other cloud to try, so one failing cloud/template can take out your entire Jenkins server.
We need to fix that - one failing template or DockerCloud shouldn't stop the other templates/clouds from being able to shoulder the load.

Proposal:
 - Add a back-off period (in seconds) to the cloud configuration.  Default it to a minute or two.
 - If a template fails, we record the failure and the failure time in a transient field (so it isn't persisted) in the template.
 - When ```DockerCloud.canProvision(Label)``` is called, it takes into account the time of the last failure and excludes any templates that failed too recently.  This may well mean that we answer "no" when we would otherwise say "yes".
 - When ```DockerCloud.provision(...)``` is called, we sort our templates in order of last failure so that templates that have never failed come first and the most recently failed templates are considered last.  That way we'll allow all templates to be tried in sequence until we find one that works or end up ruling them all out.
 - More serious issues like timeouts, a failure to log in etc, should be recorded against the cloud in similar (transient) fields and should rule out the entire cloud for a period.
 - The configuration UI should show these issues, allowing the administrator to see the last exception(s) that occurred and when they happened, to guide them in how to fix things.  e.g. if a template can't work because we've got invalid (or missing) credentials to the registry, that should show up in the UI for the template.  If a docker host locks up and stops responding, causing timeouts when talking to it, that should show up in the UI for the cloud as a whole.  (this would tie in nicely with #625)

pjdarton:
Merged.
It'll be in the next release.

PR #634, opened 2018-03-09
[REVIEW REQUESTED] Avoid broken clouds and templates

pjdarton:
Automatically temporarily disable malfunctioning templates/clouds #626 

DockerCloud and DockerTemplate now have a "Enabled" field so a user can disable an entire cloud or template until they choose to re-enable it.
If a template fails to provision, this flag is cleared (on the template) for 5 minutes.
If a cloud is unable to interrogate the docker container counts then the flag is cleared (on the entire cloud) for 5 minutes.
The cloud and template configuration pages will show when they have been temporarily disabled, why this has happened, and how long the disabling will last for.
The net effect should be that, in the event that a template or cloud fails, it'll be disabled for a few minutes, thus allowing Jenkins to try another template or cloud, thus allowing Jenkins to continue to start containers even in the event that one of its clouds/templates is unable to start containers.

Note: This doesn't currently detect containers failing to come online after they've been successfully started, it's aimed at detecting and coping with docker hosts failing rather than user configuration errors.
Also note: This PR builds upon the code from #633 so most of the changed files are actually from that PR.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -46,6 +46,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.firstOrNull;
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.withId;
@@ -103,6 +104,7 @@ public class DockerCloud extends Cloud {
      */
     private boolean exposeDockerHost;
 
+    private @CheckForNull DockerDisabled disabled;
 
     @DataBoundConstructor
     public DockerCloud(String name,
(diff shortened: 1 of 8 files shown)


ndeloof:
"disable" management looks good to me
not a big fan of the take/releaseClient approach, I prefer to rely on `Closable` and try with resources.

pjdarton:
I agree that take/release is less than ideal.  Personally it was having to add try blocks around everything that I found most objectionable, and using Closable doesn't avoid that (it just avoids the finally block) - I just wish there was a way of letting the garbage collector handle it all "by magic", but I don't think we can get away with that in this case.

OK, I'll change PR #633 so that it wraps DockerClient and makes the close method do what "releaseClient" does at present ... but it won't be this month due to other work pressures.

I'm glad you like the "disable" stuff tho'.  If you're a genius with git merge then I guess you could merge https://github.com/jenkinsci/docker-plugin/pull/634/commits/dcce7eda2518a67450edda0568e693964c1b6df4 and https://github.com/jenkinsci/docker-plugin/pull/634/commits/870252f6dc5cd771c3b380ebe9c3984048334546 to get the disable functionality without pulling in the take/release functionality.  Alternatively, you could wait a few weeks and I'll do it ;-)

Issue #644, opened 2018-04-30
No such file / directory exceptions being thrown creating nodes when persisting XML


fraz3alpha:
- [x] docker-plugin version you use - `1.1.4`
 - [x] jenkins version you use - `2.107.1`
 - [x] docker engine version you use - `Version = swarm/1.2.8, API Version = 1.22, Docker 17.09.0-ce`
 - [x] stack trace / logs / any technical details that could help diagnose this issue

We have recently moved up to Jenkins version `2.107.1`, and have also updated our plugin to the latest version - `1.1.4`. We are being persistently hit by the feature that disables provisioning of a template for 5 minutes after an exception is thrown. We are regularly seeing the following exception which originates down in the code that persists the slave's configuration to an XML file on disk :

```
Apr 30, 2018 12:02:44 PM com.nirima.jenkins.plugins.docker.DockerCloud$1 run
SEVERE: Error in provisioning; template='DockerTemplate{configVersion=2, labelString='old-sl sl swarm_node sl_swarm_node sl_always_online sl_swarm_node_latest', connector=io.jenkins.docker.connector.DockerComputerSSHConnector@2432d04f, remoteFs='/home/jenkins', instanceCap=500, mode=EXCLUSIVE, retentionStrategy=com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy@5ecbf51, dockerTemplateBase=DockerTemplateBase{image='taas-docker-public.artifactory.swg-devops.com/swarm/dk-jenkins-slave:latest', pullCredentialsId='public.shared.artifactory.docker.registry.username.password', registry=null, dockerCommand='./setup_slave.sh', hostname='', dnsHosts=[9.9.9.9], network='', volumes=[], volumesFrom2=[], environment=[], bindPorts='', bindAllPorts=false, memoryLimit=null, memorySwap=null, cpuShares=null, privileged=true, tty=false, macAddress='null', extraHosts=[]}, removeVolumes=true, pullStrategy=PULL_NEVER, nodeProperties=[], disabled=BySystem,0 ms,4 min 59 sec,Template provisioning failed.}' for cloud='taas-internal-swarm'
java.nio.file.NoSuchFileException: /var/jenkins_home/nodes/docker-22f02f15d4988b/atomic3648298142656815004tmp
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
[...]

PR #645, opened 2018-04-30
Fix for issue #644

PeppyHare:
With Jenkins 2.107.2 and docker-plugin 1.1.4, we find that during node provisioning, if many nodes are simultaneously requested, it is possible to have the node's persistent filesystem yanked from under it while the provisioner is trying to set node properties causing an exception to raise. The exception will trigger a disableBySystem, which we don't want since the provisioning could just proceed normally. This is currently our best workaround.

See issue #644


fraz3alpha:
Thanks for taking the time to reply so quickly Peter :)

To avoid any confusion - we don't expect this 'fix' to be merged, it is very much our attempts to improve things and get debug information out of an area of Jenkins that has no logging available to turn on.

pjdarton:
As discussed elsewhere, this PR was only raised as an aide to debugging and was never intended to be merged.
As the actual answer to #644 has been merged, this can be closed.

Issue #642, opened 2018-04-05
Environment variables are not passed to containers


sadovnikov:
In order to avoid Jenkins JEP-200 troubles we upgraded the plugin from `1.0.4` to `1.1.3`. However the new version seems to have a problem with passing environment variables to the containers.

Below are 
- a fragment of Docker Agent template configuration in Jenkins and 
- the value of the same environment variable as it's seen in the running container (copied from `docker inspect`).

![image](https://user-images.githubusercontent.com/4749332/38354356-9df5ed84-38b9-11e8-83cc-89c6986e9666.png)

````
"Config": {
	"Hostname": "93eccfe73b47",
	"Domainname": "",
	"User": "jenkins",
	"AttachStdin": false,
	"AttachStdout": false,
	"AttachStderr": false,
	"Tty": true,
	"OpenStdin": false,
	"StdinOnce": false,
	"Env": [
		"JENKINS_HOME=/var/jenkins_home",
		"JENKINS_LOG=/var/jenkins_log",
		"HOST_HOSTNAME=luuNNNd",
		"LANG=en_US.UTF-8",
		"LANGUAGE=en_US.UTF-8",
		"LC_CTYPE=en_US.UTF-8",
		"LC_ALL=en_US.UTF-8",
		"JAVA_HOME=/usr/lib/jvm/zulu-8"
	],
````

`luuNNNd` is default value, which is defined in the docker image and which is expected to be set through ` --env HOST_HOSTNAME` in `docker run` command. This worked as expected in `1.0.4`

Below is `DockerCloud` configuration from `config.xml`
````
<com.nirima.jenkins.plugins.docker.DockerCloud plugin="docker-plugin@1.1.3">
  <name>docker.luu182d</name>
  <templates>
	<com.nirima.jenkins.plugins.docker.DockerTemplate>
	  <configVersion>2</configVersion>
	  <labelString>jenkins-agent</labelString>
	  <connector class="io.jenkins.docker.connector.DockerComputerJNLPConnector">
		<user></user>
		<jnlpLauncher>
		  <vmargs>-Dhudson.diyChunking=false -Djava.security.krb5.conf=/etc/krb5.conf -Djava.security.auth.login.config=/etc/jaas-krb5.conf -Djavax.security.auth.useSubjectCredsOnly=true -Dsvnkit.http.methods=Negotiate -Dsvnkit.negotiate.type=spnego</vmargs>
		  <workDirSettings>
			<disabled>false</disabled>
[...]

PR #649, opened 2018-05-11
Fix issue #642

pjdarton:
JNLP connector was replacing the container-start cmd's env vars with just the JAVA_OPT var from its own setting.
It needed to add that var to existing vars.


pjdarton:
This needs to be merged before #650, and then we only want the unit-tests from #650 as this provides a better bugfix.

PR #650, opened 2018-05-11
fix for #642. Add JAVA_OPT var instead of replacing existing vars

sadovnikov:
jenkinsci/docker-plugin#642 explains the problem this PR resolves.
Comment from @pjdarton explains the code change


sadovnikov:
Hi, I fully agree. Thanks for keeping the unit test in ;)
Suggested by you work-around works well too, so we keep on using `1.1.3` until new version is available

pjdarton:
Thank you for taking the time to write the unit-test - it makes every developer's life easier when regressions are caught by unit-tests instead of manual testing :-)
I'm glad the workaround works - that helps to confirm that we've found and fixed the right bug with these code changes.

FYI version 1.1.4 is available, which adds some extra functionality and robustness - see the [changelog](https://github.com/jenkinsci/docker-plugin/blob/master/CHANGELOG.md) for details - but it'll probably be 1.1.5 that fixes this issue (and other issues).

sadovnikov:
We've got Dockerfile to "build" image with the build server and 1.1.4 was not available on https://jenkins-updates.cloudbees.com/download/plugins/docker-plugin/ for a couple of weeks after its release, so we could not use it

PR #657, opened 2018-05-30
[JENKINS-51614] Add some docs

batmat:
[JENKINS-51614](https://issues.jenkins-ci.org/browse/JENKINS-51614)
I'm sure I made some mistakes with my explanations, but I'm using this PR so that it can be corrected and completed anyway with the review of maintainers (and anyone actually).

Basically I did that after it took me some time figuring out the various connection methods.

Feel free to correct, or add more information as inline review comments, I'll be happy to make it more accurate with the feedback here.

This is the current rendering:

image

Thanks! @pjdarton @ndeloof 

cc @reviewbybees

### DIFF
--- a/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
+++ b/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
@@ -14,7 +14,7 @@
         <dd>This method runs a container, then connects to it using <code>docker exec</code>,
             all by using the Docker API.
             The agent does not need to be able to reach the master through the network layers
-            to communicate, all will go through Docker API.
+            to communicate ; all will go through Docker API.
         </dd>
 
         <dt>Connect with JNLP</dt>
--- a/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
+++ b/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
@@ -1,6 +1,6 @@
 <div>
 
-    <p>Basically, for all connection methods, Jenkins will start by triggering a <code>docker run</code>.
+    <p>For all connection methods, Jenkins will start by triggering a <code>docker run</code>.
         Then, after this step, there will optionally be more steps to establish the connection.
         There is currently three alternative ways to connect your Jenkins master
         to the dynamically provisioned Docker agents.
--- a/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
+++ b/src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
@@ -1,0 +1,31 @@
+<div>
+
+    <p>Basically, for all connection methods, Jenkins will start by triggering a <code>docker run</code>.
+        Then, after this step, there will optionally be more steps to establish the connection.
+        There is currently three alternative ways to connect your Jenkins master
+        to the dynamically provisioned Docker agents.
+    </p>
+
+    <p>There are different pros and cons for each connection method. Depending on your environment, choose the one matching your needs. More detailed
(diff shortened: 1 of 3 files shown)


src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
```diff
@@ -0,0 +1,31 @@
...
+        And the remoting agent will establish the connection with the master through the network.
+        Hence, the agent must be able to access the master through its address and port.</dd>
+
+        <dt>Connect with SSH</dt>
+        <dd>The specified image is expected to run an SSH server.
+            Then, it will treat that computer like it usually does for any SSH connected agent:
+            the master will log into it, copy the remoting agent, then start it.
+        </dd>
+    </dl>
+</div>
```
Wadeck: nit: add a trailing blank line

src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
```diff
@@ -0,0 +1,31 @@
+<div>
+
+    <p>Basically, for all connection methods, Jenkins will start by triggering a <code>docker run</code>.
```
pjdarton: While I, too, have a habit of starting many an explanation with the word "Basically", I think we can remove it in text.  It's horribly informal.
Maybe "In summary" instead?

src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-connector.html
```diff
@@ -0,0 +1,31 @@
...
+
+    <p>There are different pros and cons for each connection method. Depending on your environment, choose the one matching your needs. More detailed
+        prerequisites are provided once you select a given method.</p>
+
+    <dl>
+        <dt>Attach Docker container</dt>
+        <dd>This method runs a container, then connects to it using <code>docker exec</code>,
+            all by using the Docker API.
+            The agent does not need to be able to reach the master through the network layers
+            to communicate, all will go through Docker API.
```
pjdarton: Nit: Should be semicolon, not comma; we're making a separate, but related, statement.
(or one could use a - instead)

src/main/resources/com/nirima/jenkins/plugins/docker/launcher/DockerComputerSSHLauncher/config.jelly
```diff
@@ -3,6 +3,7 @@
 
     <f:block>
         <span class="info">Prerequisites:</span> Docker image must have <a href="https://www.openssh.com/">sshd</a> installed.
+        And the agent container (address and random port) must be reachable from the master.
```
pjdarton: Hmm, this file looks like it's very similar in nature to https://github.com/jenkinsci/docker-plugin/blob/master/src/main/resources/io/jenkins/docker/connector/DockerComputerSSHConnector/config.jelly which I edited very recently.

If we're reviewing files in https://github.com/batmat/docker-plugin/tree/fd2022e45db33d922886b76915ca8709217bdd62/src/main/resources/com/nirima/jenkins/plugins/docker/launcher then I think that, at the very least, they should be kept in step with the other files in https://github.com/batmat/docker-plugin/tree/fd2022e45db33d922886b76915ca8709217bdd62/src/main/resources/com/nirima/jenkins/plugins/docker/launcher
...or, ideally, some means found that removes the duplication.

(When I did my edit, I didn't realize that there was a very similar file elsewhere)

pjdarton: Handle it in a dedicated PR.
What you've done here is a step in the right direction, but it's highlighted existing data duplication that should be addressed if we are to make the docker-plugin a gleaming example of "code done right" ;-)

PR #660, opened 2018-06-08
Namespaces for Container Label Keys

eaglerainbow:
As by-product of #658, we have detected that https://docs.docker.com/config/labels-custom-metadata/ requests consumers of docker to use namespaces for their containre label keys. 

This PR proposes to prefix container label keys used with `com.nirima.jenkins.plugins.docker.`. 
Moreover, it refactors the existing coding such that all labels are centrally available at a constant interface.

## Mapping between old and new label keys

| Old Label Key Identifier | New Label Key Identifier
|--------------------------|-----------------------------------------------------------|
| JenkinsId                | com.nirima.jenkins.plugins.docker.JenkinsId               |
| JenkinsServerUrl         | com.nirima.jenkins.plugins.docker.JenkinsServerUrl        |
| JenkinsContainerImage    | com.nirima.jenkins.plugins.docker.JenkinsContainerImage   |
| JenkinsNodeName          | com.nirima.jenkins.plugins.docker.JenkinsNodeName         |
| JenkinsTemplateName      | com.nirima.jenkins.plugins.docker.JenkinsTemplateName     |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,5 @@
 # Changelog
 
-## 1.1.x [draft only!]
-* The label keys used for containers created have changed to ensure that they comply with [Docker's guidelines](https://docs.docker.com/config/labels-custom-metadata/) [#660](https://docs.docker.com/config/labels-custom-metadata/). Find below the mapping between the old label keys and the new one's in case you made use of them in foreign tools:
-   | Old Label Key Identifier | New Label Key Identifier
-   |--------------------------|-----------------------------------------------------------|
-   | JenkinsId                | com.nirima.jenkins.plugins.docker.JenkinsId               |
-   | JenkinsServerUrl         | com.nirima.jenkins.plugins.docker.JenkinsServerUrl        |
-   | JenkinsContainerImage    | com.nirima.jenkins.plugins.docker.JenkinsContainerImage   |
-   | JenkinsNodeName          | com.nirima.jenkins.plugins.docker.JenkinsNodeName         |
-   | JenkinsTemplateName      | com.nirima.jenkins.plugins.docker.JenkinsTemplateName     |
-
 ## 1.1.4 (2018-04-19)
 * Automatically avoid using broken clouds/templates [#626](https://github.com/jenkinsci/docker-plugin/issues/626)
 * Improved "read timeout" handling and added template "pull timeout" [#624](https://github.com/jenkinsci/docker-plugin/issues/624)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 1.1.x [draft only!]
+* The label keys used for containers created have changed to ensure that they comply with [Docker's guidelines](https://docs.docker.com/config/labels-custom-metadata/) [#660](https://docs.docker.com/config/labels-custom-metadata/). Find below the mapping between the old label keys and the new one's in case you made use of them in foreign tools:
+   | Old Label Key Identifier | New Label Key Identifier
+   |--------------------------|-----------------------------------------------------------|
+   | JenkinsId                | com.nirima.jenkins.plugins.docker.JenkinsId               |
+   | JenkinsServerUrl         | com.nirima.jenkins.plugins.docker.JenkinsServerUrl        |
+   | JenkinsContainerImage    | com.nirima.jenkins.plugins.docker.JenkinsContainerImage   |
+   | JenkinsNodeName          | com.nirima.jenkins.plugins.docker.JenkinsNodeName         |
+   | JenkinsTemplateName      | com.nirima.jenkins.plugins.docker.JenkinsTemplateName     |
+
 ## 1.1.4 (2018-04-19)
 * Automatically avoid using broken clouds/templates [#626](https://github.com/jenkinsci/docker-plugin/issues/626)
 * Improved "read timeout" handling and added template "pull timeout" [#624](https://github.com/jenkinsci/docker-plugin/issues/624)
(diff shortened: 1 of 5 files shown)


eaglerainbow:
Disclaimer: I am currently a little unsure about the lifecycle aspects of this PR: Adding the prefix to the label keys must be (more or less) considered an incompatible change. Though, we are not aware of any consumer of the labels besides ourselves. As such, the (consistent) change of this constant value would not be harming as such. However, with this change applied, we assume that at the point of time the update happens, no containers are left on the docker instance. Otherwise, those docker containers would get "lost". 

@pjdarton: What is the committment to our consumers and the administrators w.r.t. updating this plugin? Is there a committment out there that we must be "blue/green deployable"?

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -33,48 +55,13 @@
...
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
```
pjdarton: I think you've allowed your IDE to re-arrange all the import statements here.
This makes it very difficult to see what changes have actually been made.

I'm not against changes that re-arrange the code for aesthetic reasons, but they should be kept separate from functional changes.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -29,51 +52,12 @@
...
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import static org.apache.commons.lang.StringUtils.isEmpty;
-import static org.apache.commons.lang.StringUtils.trimToNull;
```
pjdarton: Import statement re-arrangement should not be part of this PR.
(Personally, I prefer to have static imports at the start too, but this is not what this PR is for)

eaglerainbow: see above.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
eaglerainbow: I will comply to your request, but still I think it's wrong:
You are right that I have set up my IDE to automatically adjust imports. Imports, IMO, are (usually) just syntactic sugar, which I, as developer, do not need to take care of (I have plenty other things to do). These kind of activities do not change anything on semantic level. 
If, by chance of this PR, we also clean up the code and don't need to start the entire CI for such a small thing, then this is good (I'd understand your concerns, if I had changed something in the coding (e.g. changes on sequence of methods, or such-a-like). 
Would you have approved a PR, with only these aesthetic changes, if I had sent it separately?

reverted with 4ae4722e8419f5ffe27fabcde0198a02b4fa1b92

src/main/java/com/nirima/jenkins/plugins/docker/DockerContainerLabelKeys.java
```diff
@@ -0,0 +1,64 @@
...
+public interface DockerContainerLabelKeys {
+
+    /**
+     * As requested by https://docs.docker.com/config/labels-custom-metadata/, keys of labels used
+     * in docker shall be prefixed by a namespace using the reverse DNS notation.
+     * All label keys of this plugin shall use this namespace as a prefix. 
+     * 
+     * Label keys defined in this interface already have this namespace prefixed.
+     */
+    static final String PLUGIN_LABEL_KEY_NAMESPACE = "com.nirima.jenkins.plugins.docker.";
```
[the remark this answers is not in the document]
eaglerainbow: > I don't think this needs to be visible outside this class.

`public` interfaces in java do not permit to specify an attribute to be private. It'll mean that we would have to make it an `abstract class`, which also has a co-notation which I don't like (as we don't want to implement this artifact at all anyway). 

On the other hand, we have the counter-example already in this plugin with `com.nirima.jenkins.plugins.docker.utils.Consts`. If we already have an approach for such cases, we should stick to one concept. I have adjusted it with 30c937bcbfe3903b32972f207d30395457054329 .


> `.class.getPackage().getCanonicalName() + "." `

I did not think over that aspect too much yet. The only thing is that I always dislike, if constants change their value. That might happen, if someone is moving our own namespace around. However, on the other side, if somebody changes the namespace of this plugin, then it'll also be a good idea, if the constants will change (as most likely the "new version" of the plugin will be "something different"). 

I browsed a little through the coding. My impression is that it more tends to the concept of class-based reference than to absolute referencing. That is why I have adjusted the coding accordingly with 983fdc36db6679c13415508c668619a1a9d27e95.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 1.1.x [draft only!]
```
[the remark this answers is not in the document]
eaglerainbow: That is wh I had marked it as "[draft only!]". 
But I am ok with it that we remove the stuff in `CHANGELOG.md` and move it to a comment here. 

I have done so with b0585093fc03b547daa6298ce601429a92ba16fe

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
pjdarton: The idea is to ensure that important changes are kept to a minimalist set of changed lines of code so that they're easy to merge and will only cause merge conflicts of there is an actual conflict of functionality.

Aesthetic changes are also necessary to keep the code from getting too ugly but they do tend to touch lots and lots of lines of code when they happen, so they're almost impossible to merge with anything, so it's important that nobody ever *needs* to, which means that these kinds of changes have to be kept apart from functionality changes.

i.e. if you find the code intolerably ugly when working, and can't bear the thought of leaving it like that, do a tidy-up PR first ... or afterwards ... but not at the same time ;-)

PR #674, opened 2018-07-31
Add option to restrict the access on docker hosts for non-admin users

cnantch:
Hello everybody,

I develop a functionality in order to restrict some actions on the docker clouds for non-admin users.


oleg-nenashev:
Closing the pull request according to the conversation with @cnantch

PR #677, opened 2018-08-15
SSH retry count and wait time correction

ccureau:
DockerComputerSSHConnector now correctly retries a SSH connection up to _maxRetryCount_ times and waits _retryWaitTime_ seconds before attempting another connection.  Added some documentation to make the action of these fields more evident.

Neither the maxRetryCount or the retryWaitTime were applied to the SSH connection logic despite being collected on the Jenkins screen since 0.1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ work
 *~
 
 dependency-reduced-pom.xml
+.DS_Store
--- a/src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
+++ b/src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
@@ -225,10 +225,10 @@ public class DockerComputerSSHConnector extends DockerComputerConnector {
         // TODO we could (also) have a more generic mechanism relying on healthcheck (inspect State.Health.Status)
         final PortUtils.ConnectionCheck connectionCheck =
                 PortUtils.connectionCheck( address )
-                        .withRetries( 30 )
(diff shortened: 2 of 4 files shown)


daxgames:
@ccureau Well that was too easy! 🥇

pjdarton:
@ccaum Well spotted ... but unfortunately it looks like the unit-tests failed as a result of the changes so we can't merge this PR as-is.
Can you make corresponding adjustments to the unit-test code as well so that they pass again?
(and, ideally, add unit tests that verify the data passed in is getting passed through to the SSH connector, i.e. unit tests that would've found this bug ages ago had they been written...)

@daxgames As you can see from the [automated tests](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fdocker-plugin/detail/PR-677/1/tests), it isn't as easy as it first appeared - "no good deed ever goes unpunished" ;-)

ccureau:
Added some sane (?) defaults if values are not passed in.  It will default to 30 tries and 2 seconds per try unless overridden.  To me that is a better solution than potentially breaking other callers. 

@pjdarton Ah, even if good deeds are punished, sometimes they just need to happen. :)

daxgames:
@pjdarton Chris and I know eaxhbother, just giving him a hard time.  We've been dealing with this for a while.  Hopefully it gets merged

pjdarton:
@ndeloof  Overall it looks like a very simple fix for a very simple problem - can you think of anything that might have been missed?  e.g. persistence / upgrade concerns?

pjdarton:
@ndeloof I'd really like a second opinion on this before it gets merged...

PR #679, opened 2018-08-22
Make it easier to run dockerNode

jglick:
By defaulting everything except the image argument to reasonable values. Should make it easier to use particularly from Evergreen. @batmat @ndeloof

### DIFF
--- a/src/main/java/io/jenkins/docker/pipeline/DockerNodeStep.java
+++ b/src/main/java/io/jenkins/docker/pipeline/DockerNodeStep.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableSet;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Util;
 import hudson.model.Computer;
 import hudson.model.Item;
 import hudson.model.Node;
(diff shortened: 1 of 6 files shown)


src/main/java/io/jenkins/docker/pipeline/DockerNodeStep.java
```diff
@@ -36,27 +37,31 @@
 
     private String image;
 
-    private String remoteFs;
+    private String remoteFs = "/tmp";
```
jglick: OTOH `DockerTemplate.provisionNode` seems to default this to container’s `WORKDIR` and, failing that, `/`. So maybe this should be left empty by default and `/tmp` being the default there? I am afraid of using `/` as a `remoteFs`; seems bound to cause permissions problems etc.

pjdarton: I'd suggest that, in general, the best idea is to make the pipeline-defined nodes work in the same way that the cloud-provisioned nodes work.

IMO using the container's workdir as a default value is much better than any other value one can come up with.  e.g. `/tmp` could be a ramdisk or other size-limited area.  `/home/jenkins` is a common choice but, strictly speaking, `/` is the only place one can rely on being present if there's no workdir defined.

Perhaps it might be better to fail the PipelineNodeStep if the specified container doesn't have a workdir and the user hasn't specified a remotefs?

jglick: At least in the case of `openjdk`, using `/` as the remote FS seems to work OK (the workspace is then `/workspace` owned by `root`), so I think I will just let it be used—no reason to aggressively fail the step.

Typz:
This seems a great improvement! Can someone please merge, and hopefully release?

PR #681, opened 2018-08-24
[JENKINS-48050] Declarative Pipeline support for dockerNode

jglick:
Implementation of something akin to [JENKINS-48050](https://issues.jenkins-ci.org/browse/JENKINS-48050) but much less ambitious in scope than https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/255 + [this patch](https://github.com/jenkinsci/docker-plugin/compare/master...ndeloof:JENKINS-48050). Rather than pretending to be a compatible variant of the existing `docker` agent type, it just introduces a new one. My reasoning is that `withDockerContainer` (for which `agent {docker …}` is sugar) is deeply flawed and its specification is essentially the implementation—for some cases it works, for other cases it does not and cannot. `dockerRun` is a totally different approach with its own set of tradeoffs. For example, `dockerRun` requires that the image contain a JVM, and uses the equivalent of `docker-run` from a Java API call made by the master, and keeps the workspace entirely private to the container which also includes the agent JVM; `withDockerContainer` could use a non-Java-related image, it uses `docker-exec` with a CLI run on some agent launched by another technique, it requires that some physical agent have access to a Docker server, it uses workspace filesystem mounts with specific Unix permission issues, it imposes special restrictions on `ENTRYPOINT`s. The list of differences is so long that I cannot imagine any practical scenario in which a user would wish to “transparently” switch from one “implementation” to another merely by installing a new plugin and perhaps flipping some global switch. Same for the `kubernetes` plugin—this is just a different world, and if you wish to use that technology, editing the `agent` line in your `Jenkinsfile` is only the first step.

No attempt yet to allow customization of server or registry credentials, etc. In fact the `dockerNode` step does not currently support a custom registry at all. (Note that the unfiled patch pretends to accept registry credentials and then pass them to the existing `dockerNode` credentials argument, but this is wrong: the existing argument is _server_ (`dockerd`) credentials, not even of the same physical type!) Anyway such additions could be done easily as compatible changes; this PR is here to get a stake in the ground. Miscellaneous requirements listed in JENKINS-48050 such as `config-file-provider` support are really just requests for tests verifying that everything “just works”, which is far more likely for `dockerRun` than for `withDockerContainer` since once the agent is started, you are doing nothing out of the ordinary, and most of that verification is not specific to Declarative either.

I just picked a symbol for the `agent` type. I do not have a strong opinion. Ideally it would be something that contrasts clearly with the existing `docker` type, which we ought to deprecate as hopeless.

@batmat @rtyler @ndeloof


src/test/java/io/jenkins/docker/pipeline/DockerAgentTest.java
```diff
@@ -0,0 +1,50 @@
...
+        WorkflowJob j = r.createProject(WorkflowJob.class, "p");
+        j.setDefinition(new CpsFlowDefinition(
+            "pipeline {\n" +
+            "  agent {\n" +
+            "    dockerContainer 'openjdk:8'\n" +
+            "  }\n" +
+            "  stages {\n" +
+            "    stage('whatever') {\n" +
+            "      steps {\n" +
+            "        sh 'java -version'\n" +
```
jglick: This is the basic usage. Pretty straightforward in basic cases: just replace `docker` with `dockerContainer` _and test_.

ndeloof:
sounds good.
I gave up with JENKINS-48050 as I can't imagine a reasonable solution as long as we have to live with a Java based remoting agent, so feel free to propose such baby-step improvements if you have some thoughts.

batmat:
CI tests run seems fishy though. I do see the new tests skipped on Windows, which is expected, but then I do not see them either under any Linux flavor. Going to double-check and run this locally too.

batmat:
OK, my bad :man_facepalming:, didn't click "Show more", and there're too many tests to make it practical to look for a specific one. 
https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/PR-681/2/testReport/io.jenkins.docker.pipeline/DockerAgentTest/ =>


image

https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/PR-681/2/testReport/io.jenkins.docker.pipeline/DockerNodeStepTest/ =>

image

So :+1:, :ship: :it: ! :-)

pjdarton:
I've been doing some housekeeping and noticed this PR was here with lots of "approved" notes but wasn't merged by @ndeloof at the time.

My knowledge of pipelines is limited; my knowledge of how to write code to support them is zero; I doubt I can tell the difference between good & bad code here ... hence this message is basically here to ask you all "WTF?" :wink: 

I did spot that there's some `TODO` comments in this PR ([DockerAgent.java](https://github.com/jenkinsci/docker-plugin/pull/681/files#diff-5caa8352d32e405e65a62feaafbd146fR19), [DockerNodeStepExecution.java](https://github.com/jenkinsci/docker-plugin/pull/681/files#diff-a684bdd4f523be8e560ebc2e3da65b19R78)), which concerns me :unamused:

I also have concerns about documentation.
The non-pipeline stuff has _lots_ of built-in help explaining what all the configuration options do, and mostly these are sufficient to get people up and running.  However, pipelines don't have linked help-text; pipelines are far more reliant on people googling for answers.
If this plugin is going to gain pipeline support then (a) it'll need to be top-tier code and (b) it'll need to be documented at least as well as the incumbents.  To do otherwise will just sow confusion. :disappointed:

Lastly, it seems to me that _lots_ of Jenkins users can't tell the difference between the different docker plugins and end up raising bugs here when they're using the docker pipeline plugin instead ... and I believe that this plugin is more "legacy" than the pipeline one.
IMO Jenkins would be a lot better off if there weren't so many docker plugins competing for attention, and I think the distinction of "pipelines are all the other plugins' responsibility" is nice and clear right now - I fear that adding pipeline support to this plugin may be taking it (and Jenkins as a whole) in the wrong direction.

Note 1: If you want to solve the "too many docker plugins" issue and have a general-purpose high-quality solution, you should _also_ check out https://github.com/KostyaSha/yet-another-docker-plugin - in an ideal world, this plugin (which KostyaSha created) and yet-another-docker-plugin would merge...

Note 2: Re skipping tests on Windows - historically, that's down to the ci.jenkins.io Windows slave VMs not supporting docker - new Windows OSs (can) have docker support, and I'm seeing issues raised by folks expecting this plugin to "just work" on Windows, so it would be nice to be able to test on both... although, personally, I have no experience of docker on windows.

jglick:
> I doubt I can tell the difference between good & bad code here

Sure. Bad code can (usually) be improved, and I am willing to put some time into polishing this up if it has a chance of being merged. I think the bigger question is whether it is a good idea to do this, whether the symbol is sufficiently self-explanatory, etc. Would appreciate any opinion from people maintaining Pipeline features on a day-to-day basis besides @abayer: @dwnusbaum, @bitwiseman, etc.

> there's some `TODO` comments

The two you linked to are RFEs, basically—additional options that could be added compatibly.

> pipelines don't have linked help-text

`Step`s do, in fact, offer in-product help, via the **Pipeline Syntax** link, and there is a subscreen for Declarative syntax which reads at least some text from the product. I suppose this could offer a `DockerAgent/help-image.html` in case the meaning of the field is not obvious, and/or `DockerAgent/config.jelly` could include an `f:description` note. But if merged, there ought to also be text on jenkins.io explaining this option and contrasting it to others.

> it'll need to be documented at least as well as the incumbents

[JENKINS-58223](https://issues.jenkins-ci.org/browse/JENKINS-58223) unfortunately means this is a low bar!

> _lots_ of Jenkins users can't tell the difference between the different docker plugins

Yes, this is a serious concern, specifically explaining the difference between this and `agent docker` (implemented in `docker-workflow` but defined, for historical reasons, in `pipeline-model-definition`; ditto the related `agent dockerfile`). Conceptually this should be much easier to explain than `withDockerContainer` ~ `agent docker`; I have come to regret writing the original, but it is out there and widely used.

> adding pipeline support to this plugin

It already _has_ Pipeline support. This PR is merely adding Declarative syntactic sugar for an existing Scripted-oriented feature: the `dockerNode` step.

> historically, that's down to the ci.jenkins.io Windows slave VMs not supporting docker

As of a few weeks ago, there is a `windock` label which can run Docker on Windows, so `DockerNodeStepTest` _might_ be able to start running on Windows too. That would require the `Jenkinsfile` to be modified to add (I think)

```groovy
platforms: ['docker', 'windock']
```

which I cannot (fruitfully) do since I lack write permissions to this repository, so someone with that permission would need to first file a PR making such a change and verifying that it does not break anything, and then merge it, before a contributor PR like this one could even attempt to take advantage of that label.

pjdarton:
> could offer a DockerAgent/help-image.html in case the meaning of the field is not obvious

Even if it's obvious, it should state it explicitly.
Users generally need to be told things twice - if they get the same information from two different routes, they'll generally follow that information.  If folks are left to do what's "obvious" then 50% will guess wrong (and then log a gazillion bugs because it didn't work the way they assumed) and the other 50% will be left uncertain they got it right.
I'm a big fan of built-in help text that removes all doubt and ambiguity :grin:

...but, before it's worth doing any of that, I think we first need to work out what strategic direction would be best for Jenkins as a whole - we don't want to replace one non-ideal solution with another one we'll regret :-)


> It already _has_ Pipeline support

My understanding is that the "pipeline support" within this plugin is minimal and buggy. e.g. the temporary templates it defines conflict with template container-limits because they're indexed by image name not the actual template (because templates don't have an id).
The core DockerCloud functionality works fairly well (because I kept fixing it until it did!), but there's a lot of stuff that's a bit "meh" and has suffered from a lack of attention.

...and, given that this plugin doesn't have any maintainers both able and willing to fix this, I don't see that changing _unless_ someone else who _does_ want that kind of thing added is prepared to drive it forwards.

FYI I'm only here reluctantly as Nick "ran off" ,leaving me in charge.  I'm not using docker pipeline functionality where I work, so I can't spend much time on it (I try to limit my time here to just basic maintenance, merging PRs that I'm 100% sure about + any urgent security fixes) rather than major development.
If _you_ want to drive this forwards then I'll happily +1 any request for a co-maintainer...


> Docker on Windows

If you can tell me exactly what changes to the (currently one-line long) Jenkinsfile are needed, I could raise one containing those changes ... however, I suspect ci.jenkins.io is unwell right now, as https://github.com/jenkinsci/docker-plugin/pull/748 has yet to be noticed and built despite a few hours elapsing.
Also, it'd be necessary to ensure that (at least initially) test failures on Windows were merely informative instead of fatal - in my experience, Microsoft rarely implement a standard without "improving" on it to the point where it no longer works with non-Microsoft software - I suspect that testing on Windows would most likely reveal that Windows docker requires Windows-specific changes ... and I can't develop that locally (at present) because my Windows dev machine is Win7 which doesn't have docker (all my docker resource is linux).

To be honest, I think the best way of getting Windows unit-tests going would be for someone from Cloudbees (with access to the ci.jenkins.co internals) to drive that forwards, as it's likely to require access that non-cloudbees folks don't have.

jglick:
> the "pipeline support" within this plugin is minimal and buggy. e.g. the temporary templates it defines

I do not know all that much about `DockerCloud`. I was referring to the `DockerNodeStep`, to which this PR is a direct follow-up.

In a sense, `DockerNodeStep` can supersede `DockerCloud`: rather than forcing a Jenkins admin to hard-code and then maintain a list of images, each `Jenkinsfile` can pick something to run. This step also bypasses the Jenkins queue (`Queue`, `NodeProvisioner`, `Cloud`, `Label`, `ExecutorStepExecution`, …) and all of its problems, since the Docker daemon handles scheduling. And compared to `withDockerContainer`, it avoids all sorts of thorny issues related to `--volume` vs. `USER` and workspace file permissions, `--volumes-from` detection when the master is containerized, `ENTRYPOINT` vs. sleeping, vagaries of `docker exec`, …

> If you can tell me exactly what changes to the (currently one-line long) Jenkinsfile are needed

Mentioned above, but more explicitly:

```diff
-buildPlugin(jenkinsVersions: [null, '2.73.3', '2.89.4', '2.107.1'])
+buildPlugin(jenkinsVersions: [null, '2.73.3', '2.89.4', '2.107.1'], platforms: ['docker', 'windock'])
```

> I suspect that testing on Windows would most likely reveal that Windows docker requires Windows-specific changes

Quite possibly, which is why you would make such a change on a PR, to see what happens.

> someone from Cloudbees (with access to the ci.jenkins.co internals) to drive that forwards, as it's likely to require access that non-cloudbees folks don't have

Well I am from CloudBees but do not have access to ci.jenkins.io internals, nor does my local Windows 10 installation have Docker (inside VirtualBox and not sure how to install it), so as in https://github.com/jenkinsci/log-cli-plugin/pull/15 I have just tried doing stuff on the server. Probably not good enough for serious development of Windows-specific code, if some is needed.

pjdarton:
> rather than forcing a Jenkins admin to hard-code and then maintain a list of images, each Jenkinsfile can pick something to run

Where I work, the former is preferred by the majority of my developers, i.e. `DockerCloud` is king.

Most of them really don't want to have to care what docker image provides the capability that they want.
Most of them tell their job to restrict where they run by labels, e.g. "unix && somecomponentsystem && somedatabase", and aren't fussed whether that'd provided by a static slave node (e.g. physical or virtual machine) or a docker container - all they care about is that it'll do shell script commands and has "some component system" & "some database" installed and running on it.
These folk generally don't _want_ to have to know the difference between a container that connects via SSH vs a container that connects via JNLP :wink:

In my experience, the folks who want to control every aspect of a docker-based build are generally ones who're running "docker run ..." commands from their build, so all they need is a Jenkins slave node that can run unix shell scripts, has the docker client installed, and has (exclusive) access to a docker daemon, as they'll be controlling the docker containers they're interested in through the docker client rather than through Jenkins.
Putting knowledge of what docker image a build needs into the build itself is a level of knowledge that `DockerCloud` allows you to abstract away and, for my devs, `DockerNodeStep` requires too much Jenkins-specific information they don't want in their source code.

i.e. *my* devs would consider `DockerNodeStep` to be step backwards, forcing them to get involved in things they don't care about.

FYI to put this in context: We also make use of the vSphere cloud plugin and the OpenStack cloud plugin to also provide other `Cloud` providers so we've got multiple sources of slave nodes (in addition to numerous docker clouds and static slave VMs).  Our users don't want to have to care what hardware resources are used to provide them with build executors; they just want "something with X and Y available" and leave it up to the Jenkins admins to find the hardware resources etc to provide that.
I think it's unfortunate that this plugin is called the "docker plugin" instead of the "docker cloud plugin".

TL;DR: This doesn't take this plugin in a direction that I need and I suspect that it's more suited to the docker-workflow plugin's area of expertise.


> but more explicitly

Thank you; that was the level of detail I needed...
So replacing the Jenkinsfile with
```
buildPlugin(
    jenkinsVersions: [
        null,
        '2.73.3',
        '2.89.4',
        '2.107.1',
    ],
    platforms: [
        'docker',
        'windock',
    ],
)
```
would probably be better in the long run, as that'll allow future PRs to add/remove individual lines instead of just replacing the only line it contains.
In fact, it may be worthwhile splitting the existing Jenkinsfile up into multiple lines first, merging that, and then doing the PR that adds the extra 4 lines for the platforms setting.
Some activity on master will at least trigger a rebuild that might clear the redness...


> I am from CloudBees but do not have access

Ah... I'd hope you'd have a better idea about who to talk to about that than I would ... but I work for IBM where "internal access" doesn't necessarily translate to "better access" either - large companies have their own rules :grin:

jglick:
> Most of them really don't want to have to care what docker image provides the capability that they want.

Whether they want to care or not, at some point they will need a newer version of some framework and have to pester the administrator to get it installed as an agent template, and then the update will suddenly break unrelated jobs that had not been prepared for incompatible changes in the new version. If the administrator is in constant contact with project developers, this is manageable, but I generally recommend making a `Jenkinsfile` be as explicit and self-contained as possible so that builds are reasonably reproducible and infrastructure changes can be done via pull request. (And where lots of projects ought to share configuration, introduce a Groovy library which can also be versioned.) Modern containerized CI systems (incl. Tekton) generally make the same choice. Jenkins of course offers an option to please everyone, and sometimes winds up pleasing no one.

> a Jenkins slave node that … has (exclusive) access to a docker daemon

If you are able to set up such infrastructure, great. But it generally means you are able to provision VMs as agents; or trust everyone and are willing to configure DinD.

> This doesn't take this plugin in a direction that I need

Fine enough, but again, this plugin _already_ defines the `dockerNode` step. This PR is merely about providing a matching `agent` syntax for Declarative. (I suppose you _could_ use `agent node` plus `dockerNode {}` inside `steps` in Declarative, but it would be awkward.)

> I suspect that it's more suited to the docker-workflow plugin's area of expertise.

Well. The `dockerNode` step is closest to `DockerCloud` in terms of what it is physically doing—connecting to a Docker daemon and launching an agent—and it in fact uses the same internal calls to do so. It is closer to `docker-workflow` functions in terms of usage patterns—allowing the author of a `Jenkinsfile` to select a Docker image in which some build steps should run—but `docker-workflow` features expect to call the `docker` CLI from an unspecified agent with the CLI installed (and, typically, a private daemon).

(In general I am in favor of deprecating most or all of `docker-workflow`: certainly the `withDockerContainer` step, perhaps also `withDockerRegistry` and `withDockerServer` that could be handled with the more generic `withCredentials`, perhaps the `docker` tool type since the whole tool system in Jenkins is no longer recommended.)

jglick:
I guess I should add that I personally work mainly with Jenkins on Kubernetes, so I would consider most Docker-based functionality to be borderline obsolete…but for those installations still using Docker directly, anything that can provide simple, sane, and convenient replacements for `docker-workflow` functionality is a step in the right direction.

pjdarton:
> administrator is in constant contact with project developers

That's our situation - our Jenkins servers are dedicated to/owned by the developers.
We're not running Jenkins as an independent service catering to general members of the public - it's _our_ Jenkins server, run by us, for us.
My aim (as a "Jenkins administrator") is to ensure that "my" devs can concentrate on writing code rather than concerning themselves with provisioning issues ... although the boundries do blur.

> VMs ... DnD

Yup; that's what we do, for builds that need to do docker operations.
We do also (for speed & efficiency) have shared docker daemons, but that means everyone has to "play nice" to avoid naming collisions etc, so that's a balancing act between execution speed vs isolation (VMs take ages to spin up compared to a container).

> Docker-based functionality ... obsolete

Yup; I agree.  Where I work, k8s is "where things are going" and our docker usage is legacy ... but it does have its uses in simpler situations, e.g. provisioning Jenkins slave nodes "on demand".

> deprecating most

Ah, so you'd like to deprecate `docker-workflow` and make _this_ plugin the main docker plugin?
Ok... that wasn't the direction I expected (I thought it was the other way around), but as long as we're not going to end up with two non-deprecated ways of doing the same thing then that's ok with me.
What would *not* be ok with me is if folks expected me to drive that development - I expect that my devs will be on k8s before we're fully on Jenkinsfile based builds (we have hundreds of non-pipeline old-style jobs defined via the job-dsl-plugin) so my appetite for the amount of investment required for re-writing pipeline support within this plugin is minimal.  My boss wants me to concentrate on doing stuff we need (and we don't need this) and I can't really offer good and well-informed opinions on issues I'm not encountering myself.
I'm not going to try to block progress (that, sadly, is how Jenkins ended up with the yet-another-docker-plugin instead of everyone concentrating efforts on this one) but it's not something I'm going to write myself.


TL;DR: If there's a real willingness (by folks other than myself) to provide top-tier, well-documented, pipeline support for docker operations within this plugin then I'm happy to cooperate ... or even grant write access so you can get on with it without bothering me at all :wink:

jglick:
> as long as we're not going to end up with two non-deprecated ways of doing the same thing

I am afraid that ship has sailed. There is not anybody in charge of deciding what tools and techniques the Jenkins project should promote, so we wind up with competing approaches.

> my appetite for the amount of investment required for re-writing pipeline support within this plugin is minimal. My boss wants me to concentrate on doing stuff we need (and we don't need this)

Quite understandable.

> If there's a real willingness (by folks other than myself) to provide top-tier, well-documented, pipeline support for docker operations within this plugin

I will ask around.

pjdarton:
> I am afraid that ship has sailed. There is not anybody in charge of deciding what tools and techniques the Jenkins project should promote, so we wind up with competing approaches.

While there's nobody with the authority to dictate what people do, the people who write and maintain the plugins and their documentation can, and effectively _do_, dictate what's promoted - users use what they find works so, by making things findable and workable, contributors do, in effect, "promote" what techniques are used.
So, while I'm not suggesting that anyone should deliver a PR that removes support for docker-workflow pipeline code, there's nothing to stop anyone providing a better option _here_ and updating the documentation to point to it.
Once a feature is undocumented and hard to find, users will stop adopting it, and the better-documented/supported feature slowly takes over.
...I just expected it'd be docker-workflow taking over from this plugin rather than the other way around...

My only real concern is that there's some nasties lurking in this code that I thought could be ignored because this code was "legacy"; if this plugin is to become "the future" then those issues really should be fixed, otherwise they'll drag the whole user-experience down.

jglick:
Right. There is a lot of stuff in this plugin that you do not necessarily want or need, especially (though not only) if you are using Pipeline exclusively, such as the `builder` package. My general inclination is towards small plugins that focus on a single well-bounded feature, but other people take the perspective that someone wanting to “use Docker in Jenkins” should only need to install “the Docker plugin for Jenkins” and get everything they might want.

pjdarton:
> builder package

Yeah, the `builder` package contains a lot of "if I were to write that today, I wouldn't do it that way" code, but doing anything to it (like punting it out into its own plugin) while maintaining backwards compatibility would be non-trivial.

> small plugins vs everything

I have some sympathy for both sides ... but I don't see why we can't satisfy both through clever code sharing.  e.g. there's a common docker plugin which provides some stuff - if more things were pushed there, a docker-pipeline plugin could contain little except the pipeline code and still have access to all the functionality it needs ... but that would mean working out how to move responsibility/ownership for some aspects of this plugin into the common one without breaking the world...

Given that this plugin already contains pipeline support _and_ you're saying that the docker-workflow pipeline code is non-ideal then I guess that adding a better implementation here is an option ... although I'd recommend that any additions are marked as "experimental" with no guarantees of continued support in future, so we're not saddled with long-term support if/when better ways of doing things become apparent.

jglick:
> I'd recommend that any additions are marked as "experimental"

Not sure if this is good enough: https://github.com/jenkinsci/docker-plugin/blob/cde48970a0ccb2260afa290d020de79dda66ae4d/src/main/java/io/jenkins/docker/pipeline/DockerNodeStep.java#L105

bverkron:
Our organization is moving to a docker based slave setup and we're very much interested in the kind of functionality the dockerNode feature provides (i.e. codifying the creation of docker containers on a specific docker host that is not the Jenkins master). Declarative pipeline support is imperative for our current model and we've been struggling to determine how to implement this. We started a PoC with dockerNode but later found out it was not usable in declarative pipelines. After spending many hours researching and fiddling we're still unclear on what the best or even 'correct' setup is for this.  The online resource are messy and almost contradictory and blurred together.

Can someone clarify or point to a resource on how to correctly setup declarative pipelines for dockers slaves on a remote host? Seemingly the most common way is doing it all through the Cloud > Docker config. If we can make use of the host specified there and define the agents and other aspects in the declarative pipeline that's fine, but we're not fans of having to define all the images and other parameters in the Jenkins config UI.

Please advise.

jglick:
@bverkron I _think_ this PR would address your use case, excepting the build from `Dockerfile` followed by use of that image in later stages of the same build which may be hard in Declarative.

bverkron:
To clarify, when you say ".... which may be hard in Declarative" do you mean currently or after this PR. Currently we're just targeting build agents / stages for the Docker implementation.

Must admit with all my research I'm still confused about how to best accomplish our goal with the current plugin versions, etc. Seemingly most people are using either A) Scripted pipelines (i.e. non-declarative) with the pre-defined Cloud > Docker > Docker Image > Docker Agent template configurations B) Declarative pipelines with the `agent { docker { image 'jenkins/jnlp-slave' } }` style syntax running against a docker daemon on the Jenkins Master, which seems to be the default (and maybe only?) way to run in a declarative pipeline? Which of course doesn't allow you to specify a remote docker host

I've seen some round about ways that might work to execute on a remote docker host in scenario B but unclear if it will work at all. We're sort of trying to piece together multiple methods through trial and error to see what might work. Some delays in getting out testing environment running so trying to dig further online for a general direction.  Apologies for hijacking this thread. Feel free to point me to a forum, etc that is better suited for this kind of discussion.

jglick:
Before this PR:

* Scripted or Declarative with a label matching an admin-configured Docker cloud; server endpoint in global settings
* Scripted `docker.image.inside` ~ `withDockerContainer`, or Declarative `agent docker` or `agent dockerfile` specifying image in `Jenkinsfile` or `Dockerfile`; agent must have Docker CLI installed, server overridable if not default
* Scripted `dockerNode` specifying image in `Jenkinsfile`; server endpoint in global settings
* whatever `yet-another-docker` plugin provides

After this PR:

* The above, plus Declarative `agent dockerNode`

bverkron:
> * Scripted or Declarative with a label matching an admin-configured Docker cloud; server endpoint in global settings
> * Scripted `docker.image.inside` ~ `withDockerContainer`, or Declarative `agent docker` or `agent dockerfile` specifying image in `Jenkinsfile` or `Dockerfile`; agent must have Docker CLI installed, server overridable if not default

These seemingly seperate config / usage scenarios you describe are where lines are blurring for me.

A label can seemingly only be defined at the Docker Agent templates level which in turn requires an image to be specified both there and in the pipeline script (as per below) if image is not included in *both* locations it results in an error complaining that `image` is required (more or less). Aside from the obvious duplication and confusion this is not ideal for us since we want to specify it via code, not in the admin UI.

```
pipeline {
    agent {
        docker {
            image 'jenkins/jnlp-slave'
            label 'testslave'
        }
    }
    stages {
        stage('Example Build') {
            steps {
                sh 'java -version'
            }
        }
    }
}
```

When this runs the container is created as expected on the remote docker host defined in Cloud > Docker. However, when it attemps to run the steps it appears to be trying to execute docker commands on the Jenkins Master rather than the Cloud Docker slave.

```
[Pipeline] {
[Pipeline] sh
+ docker inspect -f . jenkins/jnlp-slave
/home/jenkins/workspace/Ephemeral Build Agent PoC/Ephemeral Build Agent PoC@tmp/durable-24a4fc03/script.sh: 1: /home/jenkins/workspace/Ephemeral Build Agent PoC/Ephemeral Build Agent PoC@tmp/durable-24a4fc03/script.sh: docker: not found
[Pipeline] sh
+ docker pull jenkins/jnlp-slave
/home/jenkins/workspace/Ephemeral Build Agent PoC/Ephemeral Build Agent PoC@tmp/durable-becfb3df/script.sh: 1: /home/jenkins/workspace/Ephemeral Build Agent PoC/Ephemeral Build Agent PoC@tmp/durable-becfb3df/script.sh: docker: not found
[Pipeline] }
```

> * Scripted `dockerNode` specifying image in `Jenkinsfile`; server endpoint in global settings

This works and we'd like to mimic this functionality but with a declarative pipeline

```
dockerNode(
    dockerHost: 'tcp://kam1odbus134.connex.bclc.com:2375',
    image: 'jenkins/jnlp-slave', 
    remoteFs: '/home/jenkins') {
    // some block
    sh 'java -version'
}
```

jglick:
So

```groovy
    agent {
        docker {
            image 'jenkins/jnlp-slave'
            label 'testslave'
        }
    }
```

is basically sugar for

```groovy
node('testslave') { // e.g., a VM with a local Docker server
  withDockerContainer('jenkins/jnlp-slave') {
    // …
  }
}
```

and the `withDockerContainer` step effectively wraps nested `sh`s in a complicated kind of `docker run` + `docker exec`. The agent here might be anything, typically not itself running in a Docker container (though it may be possible with DinD).

`dockerNode` is totally different. It starts a Jenkins agent in a Docker container and then any `sh` steps are launched directly from that agent JVM in the usual way. This PR provides the Declarative sugar for it. Currently `dockerNode` does not support options like `remoteFs`, though it certainly could.

bverkron:
> dockerNode is totally different. It starts a Jenkins agent in a Docker container and then any sh steps are launched directly from that agent JVM in the usual way. y. This PR provides the Declarative sugar for it.

Right. So what I guess I'm looking for is the full pipeline syntax and Cloud > Docker (+ Docker agent template if required) to do this right now, without the dockerNode sugar that this PR introduces. Sounds like it will provide the exact functionality we want minus the sugar.

bverkron:
I now see where all the confusion was coming from. There are two main "paths" / approaches to communicating with Docker on a remote host...

**1) Remote host with Docker Engine API exposed to Jenkins Master (no Jenkins agent required on host)**
_Manage Jenkins > Configure System > Cloud > Docker > Docker Agent templates_
  * A label must be inclued in the agent template config (say `dockerSlave`)
  * This label must then be included in the declarative pipeline script via `agent { label "dockerSlave" }` or perhaps other syntaxes.
  * This does not allow for things like `image 'jenkins/jnlp-slave'` to be included in the declartive script as it is taken strictly from the agent template config in the UI.
  * The Docker plugin on the Jenkins master creates the container and runs the pipeline inside the container via the Docker Engine API. No jenkins save agent is required on the Docker host, but it must exists in the image.

**2) Remote host with Jenkins agent and docker engine installed**
A "Standard" Jenkins slave (Linux based) is setup and registered with the Jenkins _Master Manage Jenkins > Manage Nodes_. Docker is also installed and the Jenkins slave user is given sufficient permissions to run Docker commands locally.
  * A label must be inclued in the agent config (say `dockerEnabledSlave`)
  * This label must then be included in the declarative pipeline script via 
      ```
      agent {
        docker {
          label 'dockerEnabledSlave'
          image 'jenkins/jnlp-slave'
        }
      }
      ```
      ... or perhaps other syntaxes.
  * This allows for the declarative pipeline syntax mention here https://jenkins.io/doc/book/pipeline/syntax/#agent including specifiying the image via `image 'jenkins/jnlp-slave'` (for example) using a `dockerfile`, etc.
  * This method is totally unrleated to the Cloud > Docker method above.
  * It uses the underlying withDockerContainer, etc commands mentioned above (which can be seen in the console output of the job)
  * The Jenkins slave agents receives the instructions from the master as it would for any other job but runs it all through local Docker commands to create and run the job (or specified pieces of it) in the container. Presumably the jenkins slave agent is not required inside the image since all the master / slave interaction is happening via the agent installed directly on the remote host.

Lack of clarity in documentation, tutorials, and even stackexchange posts clamining remote hosts w/declarative pipelines were simply not possible causes us to mix the above methods generating some of our problems above (essentially it was attempting docker in docker inadvertantly, I think).

Hopefully this clarification is helpful to others. I've seen this struggle outlined many places online.

jglick:
> This does not allow for things like `image 'jenkins/jnlp-slave'` to be included in the declarative script

And that is where `dockerNode` comes in: it works basically like the first option, except

* the script, not Jenkins global configuration, specifies the agent image to use
* the Jenkins queue is bypassed, which speeds up agent allocation, reduces the incidental complexity of the system, and allows the agent launch information to be streamed to the build log

solvingj:
Not sure if this is even still an active project but an additional shortcoming of the existing docker plugin has come up and so I thought I'd point it out here for future design of this plugin. 

1. I want the docker container to run with custom user
2. I want the shell to run in a custom working directory
3. I want to opt-out of the mounting of the Jenkins volume as the jenkins user

The current plugin makes multiple things in this equation difficult. It's like... really really ugly. 

I could understand if forcing jenkins user and mounting the workspace seems like a requirement, and not doing so turns out to be impossible.  But, I believe all of the above requirements are achievable, and are numerous cloud CI's work this way (travis/appveyor/etc). 

It might be something fairly complex like... start the container as jenkins in custom working directory, install the agent files, commit the layer, setup auto registration and autostart, then stop the container, then run the container with the custom user and working directory as usual. There's probably a better idea, it's just an example.

jglick:
> shortcoming of the existing docker plugin

Unclear if this comment refers to _this_ plugin or `docker-workflow-plugin`, specifically the `withDockerContainer` step.

solvingj:
Sorry, I meant "this" to be "as you move forward with `dockerNode` " feature (whatever plugin that might live in). By the "shortcomings of the existing plugin", I meant `docker-workflow-plugin` with `docker.inside()`, `withContainer`, and `withRun`.

src/test/java/io/jenkins/docker/pipeline/DockerAgentTest.java
```diff
@@ -0,0 +1,52 @@
...
+    public void smokes() throws Exception {
+        // as in DockerNodeStepTest.defaults:
+        r.jenkins.clouds.add(new DockerCloud(
+                "whatever",
+                new DockerAPI(new DockerServerEndpoint("unix:///var/run/docker.sock", null)),
+                Collections.emptyList()));
+        WorkflowJob j = r.createProject(WorkflowJob.class, "p");
+        j.setDefinition(new CpsFlowDefinition(
+                "pipeline {\n" + "  agent {\n"
+                        + "    dockerContainer 'openjdk:11'\n"
```
basil, line 39: This image is [deprecated](https://hub.docker.com/_/openjdk). Please use a non-deprecated image.

src/main/java/io/jenkins/docker/pipeline/DockerAgent.java
```diff
@@ -0,0 +1,30 @@
+package io.jenkins.docker.pipeline;
+
+import hudson.Extension;
+import org.jenkinsci.Symbol;
+import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgent;
+import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgentDescriptor;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+@SuppressWarnings("unchecked") // TODO DeclarativeAgent.getDescriptor problem
+public class DockerAgent extends DeclarativeAgent<DockerAgent> {
```
basil, line 10: Fails SpotBugs due to lack of a serial version UID.

basil (commented):
I guess?

src/main/java/io/jenkins/docker/pipeline/DockerAgent.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final long serialVersionUID = 1;
+
+    public final String image;
+
+    @DataBoundConstructor
+    public DockerAgent(String image) {
+        this.image = image;
+    }
+
+    // TODO other properties accepted by DockerNodeStep
```
basil, line 21: This feels pretty incomplete to me without this `TODO` being implemented. Not sure I would be comfortable supporting this when it is so incomplete.

basil:
Looks fine overall; can you please either resolve or suppress as false positives the security scan warnings?

jglick:
I am not actually sure how to suppress false positives here. https://groups.google.com/g/jenkinsci-dev/c/OMe_zN8-Tkc/m/Nnqv14sbBAAJ by @daniel-beck says @yaroslavafenkin allowed `@SuppressWarnings` to work, but what would the annotation value be?

> The detailed finding descriptions on the GitHub UI explain how to use these to suppress specific findings

sounds like it applies to something visible to a repository owner, but apparently not a contributor.

Note that the flagged code is basically just copied from code already in the repository, which I suppose is also flagged somewhere but not “new”.

basil:
You can suppress them with `@SuppressWarnings("lgtm[jenkins/no-permission-check]")` and/or `@SuppressWarnings("lgtm[jenkins/csrf]")`.

basil (approved):
Thanks for the PR!

PR #685, opened 2018-09-13
Add devices flag support

kaisoz:
Hello everybody

I needed to use the devices flag in order to map devices to my container, so I implemented it. For the implementation, I followed the coding style used for other flags (volumes, etc). I'm currently using it in Jenkins 2.14.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -9,6 +9,7 @@ import com.github.dockerjava.api.model.Bind;
 import com.github.dockerjava.api.model.PortBinding;
 import com.github.dockerjava.api.model.Volume;
 import com.github.dockerjava.api.model.VolumesFrom;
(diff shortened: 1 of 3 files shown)


kaisoz:
why the CI failed? I only see a "Sending interrupt signal to process" :?

pjdarton:
I agree, it doesn't appear to be a failure in your code.  The jenkinsci process isn't perfect.
FYI if _any_ of the parallel builds fail, all the others will get cancelled, so you'll see those being interrupted - the true error is in the build that wasn't interrupted but which failed for some other reasons, thus triggering the cancellation of the others.

Try closing the PR, waiting 15 minutes, then re-opening it.  I've found that in situations where jenkinsci fails to build, it can re-trigger.
If that doesn't work, push a dummy commit.

pjdarton:
FYI you should be able to "re-open" the same PR rather than opening a new one, but what you did worked, it then built ok, and I've merged it - it'll be part of the next release.

Thank you for your contribution.

kaisoz:
I'll take it into account from now on. Thank you very much for accepting the PR :)

PR #696, opened 2018-11-05
Add capabilities and security options

elenoir:
Following features are added in this PR : 

- Capabilities : add or drop a capability
- Security options : add some security options (typically a seccomp profile)

Typically, this will allow to define a slave with less privileges than the existing "Privileged" checkbox by leveraging on Capabilities and Seccomp. You will be able to provide an agent with dedicated profile (such as node for E2E tests that can launch only Chrome Headless with associated seccomp profile).

For instance, if user adds **seccomp=file.json** as a security option, the code will take the **file.json** as an input and passes the corresponding JSON directly to the docker-java library.

Test against last Jenkins, everything seems to work.
You can check the running slave with inspect and see that CapAdd or CapDrop are filled in case of Capabilities options ; for the security options, you have look for the SecurityOpt field.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -5,11 +5,7 @@ import com.cloudbees.jenkins.plugins.sshcredentials.SSHUserListBoxModel;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.StandardUsernameCredentials;
 import com.github.dockerjava.api.command.CreateContainerCmd;
-import com.github.dockerjava.api.model.Bind;
-import com.github.dockerjava.api.model.PortBinding;
-import com.github.dockerjava.api.model.Volume;
-import com.github.dockerjava.api.model.VolumesFrom;
-import com.github.dockerjava.api.model.Device;
+import com.github.dockerjava.api.model.*;
 import com.github.dockerjava.core.NameParser;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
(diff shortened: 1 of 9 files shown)


pjdarton:
I've merged the CapAdd and CapDrop functionality.
By the time I merged this PR, the code already had securityOpts functionality ... albeit without the ability to take a file from the local filesystem.
I'm not keen on adding functionality which hits the Jenkins master's filesystem directly, especially without any policing/verification of where it's looking, so I left that part of this PR out.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -596,9 +681,62 @@ public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig
...
+    private List<String> toSecurityOpts(List<String> securityOpts) {
+        List<String> res = new ArrayList<>();
+        for (String securityOpt : securityOpts) {
+            // Seccomp : file or direct input
+            if (securityOpt.startsWith("seccomp")) {
+                String attr = securityOpt.split("=")[1];
+                Path pathAttr = Paths.get(attr);
+                if (Files.exists(pathAttr)) {
+                    try {
+                        String content = new String(Files.readAllBytes(pathAttr), StandardCharsets.UTF_8);
```
pjdarton, line 723: This bit concerns me - it's reading _any_ file from the Jenkins master's filesystem, and that's going to open up a can of worms w.r.t. security.
This existing plugin code already has support for securityOpts that does not include such dangerous functionality so I'm going to keep the existing (more limited) code and not include this (IMO dangerous) code.

IMO, if we want functionality like this, it should be provided purely via the configuration WebUI and/or build pipeline code; we shouldn't be hitting the Jenkins filesystem or the docker host's filesystem directly.
i.e. users should be entering in json, not a filename of a file containing json.

PR #713, opened 2019-01-07
[wip] fix racing condition when using JNLP Slaves

sodre:
Hi,
I am trying to use your plugin for launching Compute Nodes using docker images and more often than not I end up getting an error of this type:
```
[JNLP4-connect connection from 192.168.128.77/192.168.128.77:58220] Refusing headers from remote: Unknown client name: docker-0001g98ue3gzg
```
I have a hunch that the problem is happening because my containers are coming online and connecting back to the Jenkins Master L553 before the actual JenkinsNode has been created/commited L559-574.

Would it be possible to help me get a build of this plugin with those lines switched?

https://github.com/jenkinsci/docker-plugin/blob/907a3144f98931a5e61fc2ec9bcca865f407c955/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java#L541-L588

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -25,6 +25,7 @@ import hudson.slaves.NodeProvisioner;
 import hudson.util.FormValidation;
 import io.jenkins.docker.DockerTransientNode;
 import io.jenkins.docker.client.DockerAPI;
+import io.jenkins.docker.connector.DockerComputerConnector;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
 import jenkins.model.Jenkins;
 import org.apache.commons.codec.binary.Base64;
@@ -363,11 +364,19 @@ public class DockerCloud extends Cloud {
                             final DockerAPI api = DockerCloud.this.getDockerApi();
                             slave = t.provisionNode(api, TaskListener.NULL);
                             slave.setCloudId(DockerCloud.this.name);
-                            plannedNode.complete(slave);
 
                             // On provisioning completion, let's trigger NodeProvisioner
                             robustlyAddNodeToJenkins(slave);
 
+                            final DockerComputerConnector connector=t.getConnector();
+                            final String containerId = slave.getContainerId();
+                            final String remoteFs = slave.getRemoteFS();
+
+                            connector.beforeContainerStarted(api, remoteFs, containerId);
+                            api.getClient().startContainerCmd(containerId).exec();
+                            connector.afterContainerStarted(api, remoteFs, containerId);
+
+                            plannedNode.complete(slave);
                         } catch (Exception ex) {
                             LOGGER.error("Error in provisioning; template='{}' for cloud='{}'",
                                     t, getDisplayName(), ex);
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -559,12 +559,7 @@ public class DockerTemplate implements Describable<DockerTemplate> {
         LOGGER.info("Started container ID {} for node {} from image: {}", containerId, nodeName, getImage());
 
         try {
-            connector.beforeContainerStarted(api, remoteFs, containerId);
-            client.startContainerCmd(containerId).exec();
-            connector.afterContainerStarted(api, remoteFs, containerId);
-
             final ComputerLauncher launcher = connector.createLauncher(api, containerId, remoteFs, listener);
-    
             final DockerTransientNode node = new DockerTransientNode(nodeName, containerId, remoteFs, launcher);
             node.setNodeDescription("Docker Agent [" + getImage() + " on "+ api.getDockerHost().getUri() + " ID " + containerId + "]");
             node.setMode(mode);
--- a/src/main/java/io/jenkins/docker/connector/DockerComputerConnector.java
+++ b/src/main/java/io/jenkins/docker/connector/DockerComputerConnector.java
@@ -91,9 +91,9 @@ public abstract class DockerComputerConnector extends AbstractDescribableImpl<Do
         final ComputerLauncher launcher = createLauncher(api, workdir, inspect, listener);
 
         final Boolean running = inspect.getState().getRunning();
-        if (Boolean.FALSE.equals(running)) {
-            listener.error("Container {} is not running. {}", containerId, inspect.getState().getStatus());
-            throw new IOException("Container is not running.");
+        if (Boolean.TRUE.equals(running)) {
+            listener.error("Container {} is already running. {}", containerId, inspect.getState().getStatus());
+            throw new IOException("Container is already running.");
         }
 
         return new DockerDelegatingComputerLauncher(launcher, api, containerId);


sodre:
@pjdarton, thanks for a quick response. 

I was able to switch the logic around, although I have to admit it took a couple more tries. I think the order I am proposing is still correct and there is no need to rely on a --retry flag (I have to look for it).

I changed the issue to a PR so it easier to talk over code. If you think there is interest in merging this back into master then I can spend more time cleaning up code if required.

Alternatively, if the --retry option is the favored approach we should make that setting be available by default if launching using JNLP. Do you have any pointers to where in the code that should happen?

sodre:
@pjdarton, thanks for a quick response. 

I changed the issue to a PR so it easier to talk over code. If you think there is interest in merging this back into master then I can spend more time cleaning up code. Essentially, what I did is enforce the following order:
  1. create the container but do not start it, this gives us the containerId which is needed for step 2.
  2. create the Jenkins Slave/Node and robustly add it to Jenkins. This has racing conditions itself, and you have the 10 retries in there to get things to work.
  3. Start the container.

The order I proposed above aligns more naturally to how I have worked with Jenkins pre-containers, and there is no need to rely on a --retry flag (I have to look for that option as well!).

Alternatively, if the --retry option is the favored approach we should make that setting be available by default if launching using JNLP. Do you have any pointers to where in the code that should happen?

pjdarton:
The problem here is that different connection methods require things to be in a different order.
 - JNLP connections require Jenkins to have the slave node "robustly added" before the container starts (if the container starts very quickly and doesn't cope with Jenkins telling it that the slave isn't known yet).
 - SSH connections require the container to be started and running before Jenkins has the slave node "robustly added" otherwise Jenkins will do the first SSH connection before the container is running.
 - ...and "Docker attach" connections need the container running but idle as the actual slave connection is done via a `docker exec`, but the existing code handles that (see [DockerAttachLauncher.launch](https://github.com/jenkinsci/docker-plugin/blob/907a3144f98931a5e61fc2ec9bcca865f407c955/src/main/java/io/jenkins/docker/connector/DockerComputerAttachConnector.java#L113)).

At present, the order things are done in the existing code favors SSH.
If we switch it around so that the slave node is created in Jenkins before the container is running (to make it easier for JNLP) then that's likely to break things for existing users who are using SSH because, by default, the SSH connection method isn't configured to retry, and so that would break things for existing users, plus there's quite a bit of SSH-connection code that requires the container to be alive when the Jenkins slave node is created.

I suspect that, in order to get JNLP-based slaves working the way you want _without_ breaking SSH-based slaves for others, it'd be necessary to push the "start container" and "robustly add node" calls into the connection-specific [DockerComputerXXXConnector classes](https://github.com/jenkinsci/docker-plugin/tree/907a3144f98931a5e61fc2ec9bcca865f407c955/src/main/java/io/jenkins/docker/connector) so that each connection method does things in the order it needs.
... and then have to take a deep look at the error handling to ensure we don't "leak" resources in cases where we've created one thing (slave node or container) before failing on the other.

Note: According to `java -jar slave.jar --help`, allowing the JNLP process to retry connection failures is as simple as _not_ specifying a `-noReconnect` argument.  i.e. retries are enabled by default but can be disabled.

sodre:
@pjdarton 
> The problem here is that different connection methods require things to be in a different order.
> ... 
   
I was afraid that was going to be the case when I made the change and pushed the code for test.

> 
> At present, the order things are done in the existing code favors SSH.

I think the current state is not usable with JNLP and the default jenkins/jnlp-slave image. Do you know anyone else using JNLP with the latest version?

> 
> I suspect that, in order to get JNLP-based slaves working the way you want _without_ breaking SSH-based slaves for others, it'd be necessary to push the "start container" and "robustly add node" calls into the connection-specific [DockerComputerXXXConnector classes](https://github.com/jenkinsci/docker-plugin/tree/907a3144f98931a5e61fc2ec9bcca865f407c955/src/main/java/io/jenkins/docker/connector) so that each connection method does things in the order it needs.

Thank you for pointing me to the classes that we could use to change the way things are launched. 

> ... and then have to take a deep look at the error handling to ensure we don't "leak" resources in cases where we've created one thing (slave node or container) before failing on the other.
> 

Do we have unit-tests that check for any potential leaks? It would make it easier for me to write the code above and know I did not break something else or use any existing similar unit-tests to test for the problem.

>
> Note: According to `java -jar slave.jar --help`, allowing the JNLP process to retry connection failures is as simple as _not_ specifying a `-noReconnect` argument. i.e. retries are enabled by default but can be disabled.

I saw that in the documentation as well. I was not specifying any additional parameters to the container, and I checked that when it is launched. So it should have retried, but like I said before it is not retrying, in fact the docker container dies immediately after launching.

pjdarton:
Yup, nothing is ever quite as simple as one might hope ;-)

While I mostly use SSH in my environment, I've used JNLP myself and not encountered any problems.  I even fixed a JNLP-specific bug in PR #654 so JNLP is in active use by other folks.
You're the first to report this - my guess is that this issue is local to your environment.

I'm not aware of there being much in the way of unit-tests for error handling scenarios.  You're very welcome to write some :grin: 
The plugin does, however, have unit tests that verify that all 3 connection methods work with the 3 official Jenkins slave docker images.  Those tests all work fine (at least, they do when the jenkins-ci infrastructure doesn't run out of storage space mid-build like it did with the automated test of your code changes - if that issue persists then it might be an idea to mention it to the jenkins-ci infrastructure folks).

I think that, if your container dies immediately after launching, you might well have other issues that are causing it to fail.  e.g. if you're using Java8 and your containers are limited in the amount of memory they're allowed to use, Java is likely to try to consume too much (as Java likes to use a %ge of the host's ram, not a %ge of the container's allowance) and get killed.
Try using the "advanced" options in the JNLP connection method to specify a custom JNLP command and include some form of delay in it before it ends (e.g. sleep) so that you can manually `docker exec` a shell on the container and check logs, processes etc.  Hell, you could even create a slave in Jenkins and then manually spawn a new slave container (giving it the _exact_ same arguments that the docker plugin would give it) and see what that does.

sodre:
> While I mostly use SSH in my environment, I've used JNLP myself and not encountered any problems. I even fixed a JNLP-specific bug in PR #654 so JNLP is in active use by other folks.
> You're the first to report this - my guess is that this issue is local to your environment.
>
It could be something specific to my side. However, I was able to get the logs with docker logs, and I can reliably show that this will always fail if the Jenkins node does not exist because there is no retry logic after the slave exhausts all possible protocols. Maybe an easier PR is an additional option to hudson.remoting.Engine to continue retrying even after not finding a common protocol.

Here is the result of docker logs.
<details>

```
± docker logs 02e3be151ac6 
Warning: JnlpProtocol3 is disabled by default, use JNLP_PROTOCOL_OPTS to alter the behavior
Jan 08, 2019 10:01:54 PM hudson.remoting.jnlp.Main createEngine
INFO: Setting up agent: docker-0002yr0lks5tm
Jan 08, 2019 10:01:54 PM hudson.remoting.jnlp.Main$CuiListener <init>
INFO: Jenkins agent is running in headless mode.
Jan 08, 2019 10:01:54 PM hudson.remoting.Engine startEngine
INFO: Using Remoting version: 3.27
Jan 08, 2019 10:01:54 PM hudson.remoting.Engine startEngine
WARNING: No Working Directory. Using the legacy JAR Cache location: /home/jenkins/.jenkins/cache/jars
Jan 08, 2019 10:01:54 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Locating server among [https://****/]
Jan 08, 2019 10:01:55 PM org.jenkinsci.remoting.engine.JnlpAgentEndpointResolver resolve
INFO: Remoting server accepts the following protocols: [JNLP4-connect, Ping]
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Agent discovery successful
  Agent address: https://****
  Agent port:    ****
  Identity:      ****
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Handshaking
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connecting to ci.conda-forge.0ae.us:50000
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Trying protocol: JNLP4-connect
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Remote identity confirmed: ****
Jan 08, 2019 10:01:55 PM org.jenkinsci.remoting.protocol.impl.ConnectionHeadersFilterLayer onRecv
INFO: [JNLP4-connect connection to ****/****:****] Local headers refused by remote: Unknown client name: docker-0002yr0lks5tm
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Protocol JNLP4-connect encountered an unexpected exception
java.util.concurrent.ExecutionException: org.jenkinsci.remoting.protocol.impl.ConnectionRefusalException: Unknown client name: docker-0002yr0lks5tm
        at org.jenkinsci.remoting.util.SettableFuture.get(SettableFuture.java:223)
        at hudson.remoting.Engine.innerRun(Engine.java:614)
        at hudson.remoting.Engine.run(Engine.java:474)
Caused by: org.jenkinsci.remoting.protocol.impl.ConnectionRefusalException: Unknown client name: docker-0002yr0lks5tm
        at org.jenkinsci.remoting.protocol.impl.ConnectionHeadersFilterLayer.newAbortCause(ConnectionHeadersFilterLayer.java:378)
        at org.jenkinsci.remoting.protocol.impl.ConnectionHeadersFilterLayer.onRecvClosed(ConnectionHeadersFilterLayer.java:433)
        at org.jenkinsci.remoting.protocol.ProtocolStack$Ptr.onRecvClosed(ProtocolStack.java:816)
        at org.jenkinsci.remoting.protocol.FilterLayer.onRecvClosed(FilterLayer.java:287)
        at org.jenkinsci.remoting.protocol.impl.SSLEngineFilterLayer.onRecvClosed(SSLEngineFilterLayer.java:172)
        at org.jenkinsci.remoting.protocol.ProtocolStack$Ptr.onRecvClosed(ProtocolStack.java:816)
        at org.jenkinsci.remoting.protocol.NetworkLayer.onRecvClosed(NetworkLayer.java:154)
        at org.jenkinsci.remoting.protocol.impl.BIONetworkLayer.access$1500(BIONetworkLayer.java:48)
        at org.jenkinsci.remoting.protocol.impl.BIONetworkLayer$Reader.run(BIONetworkLayer.java:247)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at hudson.remoting.Engine$1.lambda$newThread$0(Engine.java:93)
        at java.lang.Thread.run(Thread.java:748)
        Suppressed: java.nio.channels.ClosedChannelException
                ... 7 more

Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Connecting to ci.conda-forge.0ae.us:50000
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Server reports protocol JNLP4-plaintext not supported, skipping
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Protocol JNLP3-connect is not enabled, skipping
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Server reports protocol JNLP2-connect not supported, skipping
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener status
INFO: Server reports protocol JNLP-connect not supported, skipping
Jan 08, 2019 10:01:55 PM hudson.remoting.jnlp.Main$CuiListener error
SEVERE: The server rejected the connection: None of the protocols were accepted
java.lang.Exception: The server rejected the connection: None of the protocols were accepted
        at hudson.remoting.Engine.onConnectionRejected(Engine.java:675)
        at hudson.remoting.Engine.innerRun(Engine.java:639)
        at hudson.remoting.Engine.run(Engine.java:474)
```
</details>
 

> The plugin does, however, have unit tests that verify that all 3 connection methods work with the 3 official Jenkins slave docker images. Those tests all work fine (at least, they do when the jenkins-ci infrastructure doesn't run out of storage space mid-build like it did with the automated test of your code changes - if that issue persists then it might be an idea to mention it to the jenkins-ci infrastructure folks).
> 
Okay, I will try to run these again and see if I have better luck with the infrastructure.

> I think that, if your container dies immediately after launching, you might well have other issues that are causing it to fail. e.g. if you're using Java8 and your containers are limited in the amount of memory they're allowed to use, Java is likely to try to consume too much (as Java likes to use a %ge of the host's ram, not a %ge of the container's allowance) and get killed.

I don't think that is the issue, because if I install my patched version of the docker-plugin then everything works fine as is.

Issue #712, opened 2019-01-03
Unable to specify private docker registry in Configure Jenkins -> Cloud


AshwinPai1981:
- [1.1.5 ] docker-plugin version you use

 - [2.138 ] jenkins version you use

 - [ ] docker engine version you use
Version = 1.12.6, API Version = 1.24

 - [ ] stack trace / logs / any technical details that could help diagnose this issue

We host a private docker registry from which i want to pull images. Unfortunately providing the Docker Image as <registry>/imagename:latest and have also provided the credentials id to <registry>. However I get the below error when the plugin tries to pull down the image. I can also see that the registry URL seems to be null in the 

`Error in provisioning; template='DockerTemplate{configVersion=2, labelString='docker-iib-build', connector=io.jenkins.docker.connector.DockerComputerSSHConnector@1596d10b, remoteFs='/home/jenkins', instanceCap=2147483647, mode=EXCLUSIVE, retentionStrategy=com.nirima.jenkins.plugins.docker.strategy.DockerOnceRetentionStrategy@62fe5c4, dockerTemplateBase=DockerTemplateBase{image='**registry.xxxxxxx/integration_automation/int-docker-jenkins-slave-iib-build**', pullCredentialsId='xxxxxxx', registry=### **DockerRegistryEndpoint[null;credentialsId=xxxxxxx]**, dockerCommand='', hostname='', dnsHosts=[], network='', volumes=[/home/cloud-user/jenkins:/home/jenkins], volumesFrom2=[], environment=[], bindPorts='', bindAllPorts=true, memoryLimit=null, memorySwap=null, cpuShares=null, shmSize=null, privileged=false, tty=false, macAddress='null', extraHosts=[]}, removeVolumes=false, pullStrategy=PULL_LATEST, nodeProperties=[], disabled=BySystem,0 ms,9.9 sec,Template provisioning failed.}' for cloud='docker-iib-build'
com.github.dockerjava.api.exception.NotFoundException: {"message":"No such image: registry.gitlab.mycompany.com/gitlab_group/int-docker-jenkins-slave-iib-build:latest"}

	at com.github.dockerjava.netty.handler.HttpResponseHandler.channelRead0(HttpResponseHandler.java:103)
	at com.github.dockerjava.netty.handler.HttpResponseHandler.channelRead0(HttpResponseHandler.java:33)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
[...]

PR #717, opened 2019-02-01
Fix NPE in DockerComputerSSHConnector

pjdarton:
Code was using Integer fields as int and failing to check for nulls.
This caused the exception listed here: https://github.com/jenkinsci/docker-plugin/issues/712#issuecomment-459468409


pjdarton:
Code has been tested by @InTheCloudDan and seems to have fixed the bug.

PR #720, opened 2019-02-12
Fix/Debug Docker Slave SSH Connection not possible due to NoRouteToHostException when connecting to 0.0.0.0

appidman:
Lately i was trying out jenkins docker-slaves with the docker-plugin. 

Jenkins spawned the docker containers as configured but it could never connect to them. 
I tried to find out why for mutiple days, but i couldn't find aynthing(since the related Exception was silenced) . Then i decided to go ahead and add some log lines to the plugin and build it from source.

That's why i added some logging see : https://github.com/jenkinsci/docker-plugin/commit/a0da3d28b4c40666bc6952d0068e16d7b715c7aa

After adding this log lines i could finally see the reason why it couldn't connect:

> Testing connectivity to 0.0.0.0 port 33425
> This is a message from AppIDMan the socket didn't work because of: 
> Feb 11, 2019 10:47:57 AM WARNING com.nirima.jenkins.plugins.docker.utils.PortUtils$ConnectionCheck executeOnce
java.net.NoRouteToHostException: No route to host (Host unreachable)
>	at java.net.PlainSocketImpl.socketConnect(Native Method)
>	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
>	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:204)
>	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
>	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
>	at java.net.Socket.connect(Socket.java:589)
>	at java.net.Socket.connect(Socket.java:538)
>	at java.net.Socket.<init>(Socket.java:434)
>	at java.net.Socket.<init>(Socket.java:211)
>	at com.nirima.jenkins.plugins.docker.utils.PortUtils$ConnectionCheck.executeOnce(PortUtils.java:67)
>	at com.nirima.jenkins.plugins.docker.utils.PortUtils$ConnectionCheck.execute(PortUtils.java:84)
>	at io.jenkins.docker.connector.DockerComputerSSHConnector.createLauncher(DockerComputerSSHConnector.java:252)
>	at io.jenkins.docker.connector.DockerComputerConnector.createLauncher(DockerComputerConnector.java:91)
>	at com.nirima.jenkins.plugins.docker.DockerTemplate.doProvisionNode(DockerTemplate.java:566)
>	at com.nirima.jenkins.plugins.docker.DockerTemplate.provisionNode(DockerTemplate.java:528)
>	at com.nirima.jenkins.plugins.docker.DockerCloud$1.run(DockerCloud.java:364)
>	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
>	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:59)
>	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
>	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
>	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
>	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
>	at java.lang.Thread.run(Thread.java:748)

The root cause of this issue is that when a Socket to ```0.0.0.0``` is opened in Java it **can** raise an NoRouteToHostException . I opened a StackOverflow question [here](https://stackoverflow.com/questions/54636989/java-socket-fails-to-connect-to-0-0-0-0-with-noroutetohostexception-instead-of) regarding investigation of the root cause. 

In order to fix/workaround https://github.com/jenkinsci/docker-plugin/commit/30a85b086ef6df57bd1bf776e88cbbf68eab7ab7 this problem i hardcoded host to ```127.0.0.1``` on ```PortUtils.java``` and ```DockerComputerSSHConnector.java``` . With this "fix" jenkins can connect to the docker slaves. 
I'm quite certain that this dirty fix I used is **not** the correct way to do it, i just included it so i can demonstrate to you what the issue is and how it can be resolved. 

I would really appreciate if you can make a real fix out of it if.
**Or** if you think this is a Environment Issue with my Server tell me how to fix it. 

Also I'd be really happy if you could include the logging of the exception this might help other people as well if there are some socket errors. 

Thank you.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
@@ -8,6 +8,8 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.util.concurrent.TimeUnit;
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static com.google.common.base.Preconditions.checkState;
@@ -65,6 +67,11 @@ public class PortUtils {
             try (Socket ignored = new Socket(host, port)) {
                 return true;
             } catch (IOException e) {
+                ByteArrayOutputStream out = new ByteArrayOutputStream(); 
+                e.printStackTrace(new PrintStream(out));
+                String str = new String(out.toByteArray());
+                LOGGER.warn("This is a message from AppIDMan opening the socket didn't work because of: ");
+                LOGGER.warn(str);
                 return false;
             }
         }
--- a/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
@@ -41,7 +41,7 @@ public class PortUtils {
         private long retryDelay = (long) SECONDS.toMillis(2);
 
         private ConnectionCheck(String host, int port) {
-            this.host = host;
+            this.host = "127.0.0.1";//host;
             this.port = port;
         }
 
--- a/src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
+++ b/src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
@@ -279,7 +279,7 @@ public class DockerComputerSSHConnector extends DockerComputerConnector {
         }
         String host = getExternalIP(api, ir, networkSettings, sshBindings);
 
-        return new InetSocketAddress(host, port);
+        return new InetSocketAddress("127.0.0.1", port);
     }
 
     private String getExternalIP(DockerAPI api, InspectContainerResponse ir, NetworkSettings networkSettings,


src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
```diff
@@ -279,7 +279,7 @@ private InetSocketAddress getBindingForPort(DockerAPI api, InspectContainerRespo
         }
         String host = getExternalIP(api, ir, networkSettings, sshBindings);
 
-        return new InetSocketAddress(host, port);
+        return new InetSocketAddress("127.0.0.1", port);
```
pjdarton: This will only work if Jenkins is running on the same host as the docker daemon.  It won't work in the general case where Jenkins is on a separate host from the docker hosts/swarms that it's talking to.

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
```diff
@@ -65,6 +67,11 @@ public boolean executeOnce() {
             try (Socket ignored = new Socket(host, port)) {
                 return true;
             } catch (IOException e) {
+                ByteArrayOutputStream out = new ByteArrayOutputStream(); 
+                e.printStackTrace(new PrintStream(out));
+                String str = new String(out.toByteArray());
+                LOGGER.warn("This is a message from AppIDMan opening the socket didn't work because of: ");
+                LOGGER.warn(str);
```
pjdarton: If you want to log an exception, pass it to the logger as an exception; that way the logger can decide whether or not to log anything at all and only bother calculating the stacktrace if the severity threshold is sufficient.

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
```diff
@@ -39,7 +41,7 @@ public static ConnectionCheck connectionCheck(InetSocketAddress address) {
         private long retryDelay = (long) SECONDS.toMillis(2);
 
         private ConnectionCheck(String host, int port) {
-            this.host = host;
+            this.host = "127.0.0.1";//host;
```
pjdarton: This will only work if Jenkins is running on the same host as the docker daemon. It won't work in the general case where Jenkins is on a separate host from the docker hosts/swarms that it's talking to.

appidman:
Hi, thank you for your answers.

> This is a nasty hack that would break the plugin for the majority of use-cases.

I totally aggree with you that this is a nasty hack :)
I just wanted to demonstrate that the 0.0.0.0 is the issue and that it would work with 127.0.0.1 . 

> If your environment is providing an address of 0.0.0.0 to the plugin then that's what you need to fix - > hard coding the answer within the plugin isn't the right answer.

Do you have an idea where this address 0.0.0.0 could be coming from ? I don't know where this is coming from. I'd love to fix it, but nowhere in my config i specified 0.0.0.0 . 

The only thing close to a IP would be the ```Docker Host URI``` which is set to ```unix:///var/run/docker.sock```. 

Other than that i just specify the ```Connect method``` with ```Connect with SSH``` and SSH Key ```Inject SSH key``` . So no IP Address configured anywhere.

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
appidman: Sorry i don't have any experience with the Logger in Java, so i don't really know how. 
Would this work ````LOGGER.warn(e)``` ?

pjdarton:
The IP address is coming from the docker daemon.
The plugin does the equivalent of `docker inspect` and looks at the network information to find out the IP address and port which the container's SSH daemon is being exposed as.
I guess it's possible that 0.0.0.0 is what the plugin sees if the docker daemon doesn't provide any answers at all (you'd have to check the code to see if that's the case).

You *are* exporting/exposing port 22 from your container, aren't you?  If you're not, it would explain why the IP/port isn't being found.

PS. Have you tried the "direct attach" method?  It's probably the most efficient of all of them...

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
pjdarton: See https://github.com/jenkinsci/docker-plugin/blob/30a85b086ef6df57bd1bf776e88cbbf68eab7ab7/src/main/java/com/nirima/jenkins/plugins/docker/DockerJobProperty.java#L182 for an example.

appidman: Thank you

appidman:
Hi thank you for your response it helped me in tracking down the issue and i think I can now provide a clean fix.

The situation is as follows (as far as i understand it): 

A docker inspect is being done in ```DockerComputerConnector.java:89``` 

```java
public final ComputerLauncher createLauncher(final DockerAPI api, @Nonnull final String containerId, String workdir, TaskListener listener) throws IOException, InterruptedException {
        final InspectContainerResponse inspect;
        try(final DockerClient client = api.getClient()) {
            inspect = client.inspectContainerCmd(containerId).exec();
        }
        final ComputerLauncher launcher = createLauncher(api, workdir, inspect, listener);
```

The inspect response is then passed down into ```DockerComputerSSHConnector.createLauncher```. 
Where ```getBindingForPort(api, inspect, port) located on  DockerComputerSSHConnector.createLauncher:238``` is called. 

getBindingForPort does the following call 
```String host = getExternalIP(api, ir, networkSettings, sshBindings) located on DockerComputerSSHConnector.createLauncher:280;```
)  .
The final method which is then deciding which IPAddress to return is **getExternalIP**
```java
private String getExternalIP(DockerAPI api, InspectContainerResponse ir, NetworkSettings networkSettings,
                                 Ports.Binding[] sshBindings) {
        // If an explicit IP/hostname has been defined, always prefer this one
        String dockerHostname = api.getHostname();
        if (dockerHostname != null && !dockerHostname.trim().isEmpty()) {
            return dockerHostname;
        }

        // for (standalone) swarm, need to get the address of the actual host
        if (api.isSwarm()) {
            for (Ports.Binding b : sshBindings) {
                String ipAddress = b.getHostIp();
                if (ipAddress != null && !"0.0.0.0".equals(ipAddress)) {
                    return ipAddress;
                }
            }
        }

        // see https://github.com/joyent/sdc-docker/issues/132
        final String driver = ir.getExecDriver();
        if (driver != null && driver.startsWith("sdc")) {
            // We run on Joyent's Triton
            // see https://docs.joyent.com/public-cloud/instances/docker/how/inspect-containers
            return networkSettings.getIpAddress();
        }

        final URI uri = URI.create(api.getDockerHost().getUri());
        if(uri.getScheme().equals("unix")) {
            // Communicating with unix domain socket. so we assume localhost
            return "0.0.0.0";
        }
        return uri.getHost();
    }
```
As one can see here the docker inspect response here is only used to check whether there is a driver present and to check whether it starts with sdc. 

If this isn't the case it's checking if the uri scheme is a **unix socket**,  if so it returns "0.0.0.0" and the docker inspect value isn't actually used in this case. 

So I think it would be fine if return ```return 0.0.0.0``` get's replaced with ```return 127.0.0.1``` since ```0.0.0.0``` doesn't work reliably on all hosts but ```127.0.0.1``` should.  There is even a comment explaining the intention, behind it:  
```java
// Communicating with unix domain socket. so we assume localhost
            return "0.0.0.0";
```
What do you think ? Let me know if you think there is an issue with this approach . If you see no issue with this, I would start testing the fix on our Jenkins and if there is no issue I would update the PR with a new commit, for a clean fix. 

I'd like to test my pfirst

appidman:
I tested my change and i can now successfully connect to the Docker Slave on my Jenkins.
I also updated the Pull Request: It should now be a clean fix, also the exception logging is now done as you suggested.

pjdarton:
The more I think about this, the more I think it shouldn't be necessary, as there's a field titled "Docker Hostname or IP address" in the UI (Manage Jenkins -> Configure System -> Cloud -> Docker -> Docker Cloud Details ... -> Advanced ...) which I _believe_ is meant to solve this kind of problem for you.

i.e. in situations Jenkins accesses the docker API via a unix socket (meaning that the docker plugin can't work out the IP address for the SSH endpoints), I think it uses the "Docker Hostname or IP address" field.

So, in your case, if you set that field to _either_ 127.0.0.1 _or_ your Jenkins server's actual IP address _or_ your Jenkins server's own hostname _then_ there should be no need for this patch.

If that's the case then I think the nature of this bug changes to "the built-in help isn't good enough to explain this" rather than requiring code changes...

appidman:
Hi, I tried your suggestion, so when i set **"Docker Hostname or IP address"** in the _UI (Manage Jenkins -> Configure System -> Cloud -> Docker -> Docker Cloud Details ... -> Advanced ...)_ to 127.0.0.1 it works as well. 

Returning 0.0.0.0 for localhost is still an issue in my opinion . It might work on some machines but it might not work on others (even if OS and Java version are the same). Just as an example if you try to open a socket to 0.0.0.0 on Windows you'll get an Address Invalid Exception. 

So I still think the fix would make sense, but I'm happy that i have now another way to get my container integration to work. 

Thank you

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
```diff
@@ -65,6 +65,7 @@ public boolean executeOnce() {
             try (Socket ignored = new Socket(host, port)) {
                 return true;
             } catch (IOException e) {
+                LOGGER.warn("Socket to host: " + host + " with port: " + String.valueOf(port) + " couldn't be opened", e);
```
pjdarton: This is going to spam the log with nasty exceptions even in normal operating conditions.
While I can understand why you might want to have added this during dev work, it can't go into the core code as-is.

FYI _normally_ the docker plugin will run `executeOnce` multiple times after starting the docker container, waiting for the port to open.  If it doesn't open then we get a _single_ log message about it.
I don't want to fill the log full of exception traces just because the socket wasn't open before we started looking.

TL;DR: Please remove this line,

src/main/java/io/jenkins/docker/connector/DockerComputerSSHConnector.java
```diff
@@ -109,7 +109,7 @@ public void setJvmOptions(String jvmOptions) {
     public String getJavaPath() {
         return javaPath;
     }
-
+
```
pjdarton: You've added a trailing whitespace to this line.
Please remove the whitespace and put the line back how it was.

pjdarton:
As there's been no movement on this for over a year, I'm going to close it.
If you still want this, feel free to re-open this (or a new PR), resolve the conflicts and requested changes and I can take another look at merging it.

Issue #731, opened 2019-04-20
New option for container stop timeout


sithmein:
It would be nice if the stop timeout of a container could be configured. Currently it's hard-coded to 10s. I have a container that needs to perform some clean after the job finishes and this may take longer than 10s. Therefore this values should be configurable similar to the pull timeout.

PR #732, opened 2019-04-20
Issue 731: Add new docker template option for container stop timeout

sithmein:
Addresses #731 .

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -56,6 +56,12 @@ import java.util.Set;
 
 
 public class DockerTemplate implements Describable<DockerTemplate> {
+	/**
+	 * The default timeout in seconds ({@value)s} to wait during container shutdown
+	 * until it will be forcefully terminated.
+	 */
+	public static final int DEFAULT_STOP_TIMEOUT = 10;
+	
     private static final Logger LOGGER = LoggerFactory.getLogger(DockerTemplate.class.getName());
 
     private static final UniqueIdGenerator ID_GENERATOR = new UniqueIdGenerator(36);
@@ -84,6 +90,8 @@ public class DockerTemplate implements Describable<DockerTemplate> {
     private DockerTemplateBase dockerTemplateBase;
 
     private boolean removeVolumes;
+    
+    private int stopTimeout = DEFAULT_STOP_TIMEOUT;
 
     private transient /*almost final*/ Set<LabelAtom> labelSet;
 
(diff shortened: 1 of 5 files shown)


sithmein:
The test failures are very likely not caused by the PR since the same tests have also been failing for the latest commit on master.
I cannot run them locally, too, especially `DockerComputerSSHConnectorTest` always times out.

pjdarton:
Thanks for the PR.
I'm a bit snowed-under at work to deal with this right now (and going on vacation tomorrow for a week) so this isn't going to get merged immediately, but once I'm "back in the game" I'll take a detailed look and take it from there.
In the meantime, if you could "persuade" the JenkinsCI system to do a non-broken build, that would be good.  FYI if you close the PR, wait 10-15 minutes (for the JenkinsCI system to notice it's gone), and then re-open the PR (i.e. re-open the old PR; don't create a new one) then the JenkinsCI system should have another try at building it.
If it's consistently borked, bring it to the attention of Jenkins folks on IRC e.g. http://webchat.freenode.net/?channels=jenkins-infra

Re: Can't run unit-tests locally
I used to have that problem too - until I updated docker on my test system.  There's nothing special about the build environment that the JenkinsCI systems have so you should be able to run the build & tests locally.

pjdarton:
I'm going to close this PR, wait a bit, and re-open it to try to get the Jenkins CI build to rebuild it - hopefully it'll build cleanly...

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -84,6 +84,8 @@
     private DockerTemplateBase dockerTemplateBase;
 
     private boolean removeVolumes;
+    
+    private int stopTimeout = 100;
```
pjdarton: This is not consistent with the rest of the code, which has a default of 10.

src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplate/help-stopTimeout.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    Timeout, in seconds, to apply when stopping the container after a job has finished. The default of 10s is fine for
+    most cases. However, if you container needs to do some cleanup after the slave has stopped you can increase the
+    timeout. After the timeout the container will be killed.
```
pjdarton: Please rephrase the last sentence.
"If the container fails to stop gracefully before the timeout ends, the container will be killed."

I'd also suggest that you use a new line for every sentence rather than wrapping the text manually (it'll look the same to the end user, but it'll means that any subsequent edits to this file will result in smaller & simpler git commits).

Issue #735, opened 2019-06-15
New option for using the build slave name as container name


pcavezzan:
This feature is pretty similar to the issue #404 

With my team, we use docker to run containers as build slave thanks to this awesome docker plugin.

It runs perfectly fine but as we would like to monitor those building containers slave, we think it would be nicer to lookup directly to the slave thanks to their names as container name.

Right now, we surely can monitor them but to get to the right running container, we need to:
1) get the slave name generated 
2) go to Docker admin page to look at the lookup table displaying the running containers
3) search and find for the container ID associated to the running slave
4) go to our monitor dashboard to look at the differents stats for the docker container thanks to the ID

In our situation, it would be much simplier to :
- directly get the slave name generated
- go to our monitor dashboard to look at the differents stats for the docker container thanks to the name

I wrote this new optional feature (and will open a PR) with a simple checkbox (by default, unchecked to keep the actual docker plugin behavior).
If the checkbox is checked, then we run the container with the same name as the slave name generated. 
This should not be a problem as the docker slave name is generated with a unique name.

pjdarton:
To be honest, I'm not keen on having Jenkins specify container names :-/

FYI a number of users (including me) previously hit problems with this plugin precisely because any name generated by Jenkins cannot be guaranteed to be unique within a given docker daemon.
Jenkins only knows its own slave names; when you have a docker daemon (or docker swarm) that's shared between multiple Jenkins instances (as numerous folks do) then you can get multiple Jenkins instances all choosing to use the same name, causing a naming-collision and then you get failures.
To ensure unique container names within a docker daemon, you _have_ to leave it up to the docker daemon to choose them.

However, because users (such as yourself) _also_ want to be able to go poking inside containers, the plugin sets labels that contain lots of useful meta-data; this includes the Jenkins instance's ID, its URL, and (as you want) the name of the Jenkins slave within that Jenkins instance.

Personally, I think it'd be easier to make the container name easily available from the slave node page.
The slave node page already tells you "Docker container <container-ID> on cloud <docker-cloud-name>" - I _suspect_ that it would be relatively simple (and non-invasive) to make cosmetic enhancements to include the container name (as well as the ID) on that page.

pcavezzan:
Hi @pjdarton ,

I hope you're doing well !
Sorry for my late answer, quite busy this couple of days...

Yes, your idea to add the container name in the sentence is great :

<ID_CONTAINER> on Cloud xxxx
like 
[...]

PR #736, opened 2019-06-15
Issue 735: New option for using the build slave name as container name

pcavezzan:
New feature added according to description at #735


pcavezzan:
As @pjdarton mentionned, there is no need for settings the container name on the slave.
To get to this information, you should query on containers labels to filter out the ones with com.nirima.jenkins.plugins.docker.JenkinsNodeName.

PR #742, opened 2019-07-18
Log exception if connecting to container is failing

appidman:
Log exception if connecting to container is failing, otherwise user might have no feedback if ssh connection to the container fails

**Background info**
I just spent a few days trying to figure out why the docker plugin stopped working:
In jenkins i could only see: ```Jenkins does not have label "mycontainer" ``` and the docker plugin debug log didn't show anything useful. Which caused me to spend a lot of time troubleshooting. In order to save people from stepping in the dark i created this PR. 

It was caused by a firewall change on the CENTOS machine I'm running jenkins on. Unfortunately there was no trace in the log of the docker plugin that the ssh connection failed , so i didn't really know where to look.


appidman:
Can anyone merge this ?

pjdarton:
Yes, but not right now; probably about 2 weeks - I got married a week ago and I'm out of the office at present.

In the meantime, you can get a plugin built from master+your PR from the "all checks have passed" bit above, if you drill down and find the build artifacts from the automated build.

Normal service will resume in about two weeks.

appidman:
cool thx and congratulations 👍 😃

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -575,11 +573,17 @@ private DockerTransientNode doProvisionNode(DockerAPI api, DockerClient client,
             node.setDockerAPI(api);
             finallyRemoveTheContainer = false;
             return node;
-        } finally {
+        }
+        catch(Throwable e) {
```
pjdarton: I don't like the idea of catching `Throwable`.
I'd suggest you just catch whatever checked exceptions that code can throw, plus `RuntimeException`.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -575,11 +573,17 @@ private DockerTransientNode doProvisionNode(DockerAPI api, DockerClient client,
             node.setDockerAPI(api);
             finallyRemoveTheContainer = false;
             return node;
-        } finally {
```
pjdarton: Code style:  The rest of this code uses
```
} catch(...) {
```
whereas you've changed this to
```
}
catch(...) {
```
making this bit inconsistent with the rest of the file.
Same applies to `} finally {` too.

While I try not to let my opinions on "which style is better" get in the way of progress :grin: but I do care about consistency.
...also, if you change the code style of your new code to be consistent with the existing style, it'll make for less changes overall, which is no bad thing.

pjdarton:
Gentle-nudge... if you sort the coding-style issues I highlighted, I'd be happy to merge.

basil:
This is solving the problem at the wrong layer. A stack trace is now properly printed as of jenkinsci/jenkins#7284, which implemented the solution at the right layer.

PR #744, opened 2019-07-24
Add a configuration for securityOptions

gamma:
Setting the security options are required to add a windows docker container to an Active Directory. The Docker-Java plugin supports this options.

AFAIK this has been in discussion about two years ago, but was discarded due to the docker-java plugin not being able to set the option either. That has changed last December.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -104,6 +104,9 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
     @CheckForNull
     private List<String> extraHosts;
 
+    @CheckForNull
+    private List<String> securityOpts;
+
     @DataBoundConstructor
     public DockerTemplateBase(String image) {
         if (image == null) {
@@ -405,6 +408,16 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
         return Joiner.on("\n").join(extraHosts);
     }
 
+    public String getSecurityOptsString() {
+        if (securityOpts == null) return null;
+        return Joiner.on("\n").join(securityOpts);
+    }
+
+    @DataBoundSetter
+    public void setSecurityOptsString(String securityOpts) {
+        this.securityOpts = splitAndFilterEmptyList(securityOpts, "\n");
+    }
+
     // -- UI binding End
 
 
@@ -642,6 +655,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
         sb.append(", cpuShares=").append(cpuShares);
         sb.append(", shmSize=").append(shmSize);
         sb.append(", privileged=").append(privileged);
+        sb.append(", securityOpts=").append(securityOpts);
         sb.append(", tty=").append(tty);
         sb.append(", macAddress='").append(macAddress).append('\'');
         sb.append(", extraHosts=").append(extraHosts);
@@ -677,6 +691,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
         if (cpuShares != null ? !cpuShares.equals(that.cpuShares) : that.cpuShares != null) return false;
         if (shmSize != null ? !shmSize.equals(that.shmSize) : that.shmSize != null) return false;
         if (macAddress != null ? !macAddress.equals(that.macAddress) : that.macAddress != null) return false;
+        if (securityOpts != null ? !securityOpts.equals(that.securityOpts) : that.securityOpts != null) return false;
         return extraHosts != null ? extraHosts.equals(that.extraHosts) : that.extraHosts == null;
     }
 
@@ -699,6 +714,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
         result = 31 * result + (cpuShares != null ? cpuShares.hashCode() : 0);
         result = 31 * result + (shmSize != null ? shmSize.hashCode() : 0);
         result = 31 * result + (privileged ? 1 : 0);
+        result = 31 * result + (securityOpts != null ? securityOpts.hashCode() : 0);
         result = 31 * result + (tty ? 1 : 0);
         result = 31 * result + (macAddress != null ? macAddress.hashCode() : 0);
         result = 31 * result + (extraHosts != null ? extraHosts.hashCode() : 0);
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -426,7 +426,7 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
 
     @CheckForNull
     public List<String> getSecurityOpts() {
-        return this.securityOpts;
+        return securityOpts;
     }
 
     @CheckForNull
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -414,7 +414,9 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
     }
 
     public String getSecurityOptsString() {
-        if (securityOpts == null) return null;
+        if (securityOpts == null) {
+            return "";
+        }
         return Joiner.on("\n").join(securityOpts);
     }
 
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
@@ -427,21 +427,23 @@ public class DockerTemplateBase implements Describable<DockerTemplateBase>, Seri
         return Joiner.on("\n").join(extraHosts);
     }
 
-    @CheckForNull
+    @Nullable
     public List<String> getSecurityOpts() {
         return this.securityOpts;
     }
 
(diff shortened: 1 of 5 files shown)


src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplateBase/help-securityOpts.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    A list of new line separated security options.
```
pjdarton: Please expand this help.
Unless I've misunderstood, I think that the syntax expected here is that each line specified here is equivalent to passing one `--security-opt` argument to `docker run` - if that's true, please expand the help text to make this clear (and if not, please explain what it really is).
Maybe even include an example and/or a URL link to https://docs.docker.com/engine/reference/run/#security-configuration too.
If it's clearly documented then we won't end up getting mistaken "bug reports" from people who didn't understand it :grin:

i.e. please use this file to remove _any_ excuses for people to raise a bug report about this...

gamma:
Note the failing check is something internal in the tests and has nothing to do with my changes as far as I can see ... can you please check and re-run the CI Job?

pjdarton:
The trick I use to re-run the CI job for a PR is to close the PR, wait 10 minutes (so that the Jenkins-CI server has time to notice), and then re-open the PR again (Note: don't create a new PR; just re-open the old one).
I think it's only Cloudbees employees that can re-trigger a CI build just by prodding the CI Jenkins server directly - I don't have any more access to that server than you do.

...however, I think that the unit-test failure _is_ caused by your changes - the "expected" config had `securityOpts=null` and the "actual" config (after the round-trip_ had `securityOpts=[]` which causes `DockerTemplateBase.equals(...)` to return false (as null isn't considered to be the same as empty).
I'd guess that's because [DockerTemplateBase.setSecurityOptsString(String)](https://github.com/jenkinsci/docker-plugin/pull/744/files#diff-42be2c48b9d97e78cfad41d09befa720R423) does not distinguish between being given a null string and being given an empty one.
Right now, you've got a mix of "treat null and empty differently" and "turn one into the other" which is what's causing the unit-test failure as what's being put in isn't the same as what's coming out again.

I can think of two options here:
1. Having all getter/setter methods always turn "empty" into "null" (so the rest of the code never encounters an empty list).  Note: You may need to change hashCode and equals to call the getters instead of hitting the fields directly.
1. Ensure that the getter/setters always distinguish between null and empty so that this difference is preserved when the class is serialized to XML and deserialised back to an instance.

gamma:
OK, will check. I just noticed that the `disabled` field in the error was `null` was `no`

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -752,7 +780,17 @@ public FormValidation doCheckExtraHostsString(@QueryParameter String extraHostsS
...
+                }
+            }
+            return FormValidation.ok();
+        }
+
+        public FormValidation doCheckSecurityOptsString(@QueryParameter String securityOptsString) {
+            final List<String> securityOpts = splitAndFilterEmptyList(securityOptsString, "\n");
+            for (String securityOpt : securityOpts) {
+                if (securityOpt.trim().split("=").length < 2) {
+                    return FormValidation.error("Wrong security option format: '%s'", securityOpt);
```
pjdarton: According to https://docs.docker.com/engine/reference/run/#security-configuration, while _most_ of the `--security-opt` arguments are of the form "name=value", there's at least one that isn't:
> --security-opt="no-new-privileges:true|false"   : Disable/enable container processes from gaining new privileges

...and this is then further confused by the example:
> docker run --security-opt no-new-privileges -it centos bash

which doesn't have a `:true` or `:false` at the end.

As-is, this code would complain to the user if they specified that option in either form.

pjdarton:
`DockerTemplate.getDisabled()` turns nulls into an instance _and_ `DockerTemplate`'s `hashCode` & `equals` methods both call `.getDisabled()` instead of directly accessing `.disabled`, which is why this isn't a problem.  FYI `pullStrategy` is handled much the same way.

i.e. this coerces nulls into non-nulls and also ensures that this isn't a problem.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
gamma: The documentation does not state whether there could be more of these options. I could either add a value checker for `key:label` values in general or specifically check for `no-new-privileges`. I think that last option is safer towards the current documentation which refers to a kernel documentation. 

Which would you prefer?

pjdarton: Well, we don't want this plugin code to have to be updated every time docker gets enhanced, so we want to decouple our code from the docker implementation as much as possible...

Suggestion: if there's something that isn't "name=value" and isn't `no-new-privileges`, `no-new-privileges:true` or `no-new-privileges:false` then we raise a warning (not an error) indicating that this "does not look like a valid security option".

My thinking here is that:
 - as we can't be 100% certain that it _can't_ be valid then we shouldn't indicate that it's an error, which is why we use a warning.
 - it looks like options are _usually_ of the form "name=value" which suggests (to me) that any new options added in the future are likely to be of this form too.
 - we shouldn't state as fact anything we can't be 100% sure of, which is why we say "may be invalid" rather than "is invalid".

FYI: If you would like the UI to include localized strings for these messages (e.g. so you can have the error in German when the browser locale says that's preferable) then you can do that - you'd add a line (e.g. `DockerTemplateBase.unrecognizedSecurityOpt={0} may not be a valid security option.`) to `src/main/resources/com/nirima/jenkins/plugins/docker/Messages*.properties`, which would then create a new pair of methods in `com.nirima.jenkins.plugins.docker.Messages`, which you'd then call from the validator.  However, personally, I'd suggest that adding localization support is a much bigger, and separate, issue that'd be best tackled as a separate PR.

pjdarton:
Review in progress but I ran out of time today.  FYI it's looking good aside from some trivialities.

However, I'd like to push out a new release ASAP so *please let me know* if you think there's any work outstanding on this.
Similarly, if you'd be happy if this was merged as-is then please let me know too.

If you don't tell me *not* to merge, I may decide deal with any trivial review issues myself, merge it all in and push out a release.  If you'd prefer this PR to wait until the next release then please let me know ASAP, preferably within the next 19 hours (i.e. by 1pm UK/London time, aka 12 noon GMT), as I'd like to push out a new release (in my) tomorrow afternoon (because #752 is a problem that needs an urgent fix).

gamma:
I'm good with a merge as is. I'm currently not waiting urgently for the changes but if you think it is clear for merging please feel free.

pjdarton:
<grumble>
Well my aim to get this working quickly was foiled - https://issues.jenkins-ci.org/browse/JENKINS-26407 bit me.  The `doCheck...` method wasn't getting called and it took me ages to figure out why I wasn't able to test it.
I do not understand Jenkins "jelly" & its associated bindings to Descriptor code - it's all magic to me - so when it doesn't "just work", I struggle...
</grumble>

The upshot is that I didn't do a release today; maybe tomorrow; maybe next week.
If you can get changes sorted in time then I'll merge this before I release; if not, it'll go into the next one.

pjdarton:
FYI that build problem looks like a Jenkins infrastructure problem.
Previously I've found that it helps to mention this sort of thing on the [jenkins-infra IRC channel](http://webchat.freenode.net/?channels=jenkins-infra) as that way someone from Cloudbees can give the system a kicking until it works.

gamma:
Yeah, I thought so because it was in the clean-up phase. I tried the trick nevertheless but it did not trigger a new build. Could you take a look anyway? I hope I didn't miss something *fingers crossed*

gamma:
Github says there is one more change requested ... I can not find out which one is meant. And the build system still is stuck. Can you advice?

pjdarton:
I don't think the build system is stuck - I think that's just a very poorly-coded unit test.
There are other PR builds, e.g. https://ci.jenkins.io/job/Plugins/job/docker-plugin/view/change-requests/job/PR-720/16/consoleFull , which report:
```
12:54:19  [WARNING] Flakes: 
12:54:19  [WARNING] io.jenkins.docker.pipeline.DockerNodeStepTest.toolInstall(io.jenkins.docker.pipeline.DockerNodeStepTest)
12:54:19  [ERROR]   Run 1: DockerNodeStepTest.toolInstall » TestTimedOut test timed out after 180 seconds
12:54:19  [ERROR]   Run 2: DockerNodeStepTest.toolInstall » IndexOutOfBounds Index: 0
12:54:19  [INFO]   Run 3: PASS
```
so I don't think the fact that it failed for this PR is because of problems introduced in this PR.
The fact that the [code](https://github.com/jenkinsci/docker-plugin/blob/f77a14e1e527d6f4e5e2eef3f85f43b7052b7964/src/test/java/io/jenkins/docker/pipeline/DockerNodeStepTest.java#L125) says:
```
                // I feel like there's a way to do this without downloading the JSON, but at the moment I can't figure it out.
                DownloadService.Downloadable mvnDl = DownloadService.Downloadable.get("hudson.tasks.Maven.MavenInstaller");
                mvnDl.updateNow();
                DownloadFromUrlInstaller.Installable ins = story.j.get(Maven.MavenInstaller.DescriptorImpl.class).getInstallables().get(0);
```
implies that the author didn't think it was very good even when they wrote it ... and now it's in there, causing problems for everyone else 😞 

...and I think the "one remaining change" is because GitHub isn't reacting to you marking conversations as resolved - it's probably waiting for me (who "requested changes") to mark it as approved.
FYI I've looked through the code and it certainly _looks_ nice and clean now.

I'll try closing the PR, waiting a bit, and re-opening it to see if we can get a nice clean build...

Issue #751, opened 2019-10-01
Retries in ConnectionCheckSSH


php50350:
In the execute method of ConnectionCheckSSH, there is a line of code as follow:

`final int retries = Math.min(0, parent.retries);`

That causes the number of retries to be always 0 (unless the parent.retries is negative). The result is that when the plugin starts a Docker container and then attempts to connect to it, the error "java.io.IOException: SSH service hadn't started after 0 seconds." will often be encountered.

pjdarton:
"oops"
Yes, that looks like a bit of an oversight on my part.
...and that code has no unit-tests so this one slipped through.

I'll fix it.

pjdarton:
OK, I've created PR #752 that should fix this.  It's building on ci.jenkins.io at present (right now it's successfully run the build and tests and is waiting for an executor to save the final results).

Once it's built, you should be able to get a `.hpi` file from the "Last Successful Artifacts" listed under https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/PR-752/
Once that's available, please download it, install it on your Jenkins server(s) [manage jenkins -> manage plugins -> advanced], test that it fixes the bug for you and let me know the results (here).
If it's all confirmed ok then I'll release the fix officially.

pjdarton:
...in the meantime, if you could check the code changes (in #752) then that would be helpful - it would be rather embarrassing if the fix for this was also buggy... :wink:

php50350:
Thank you for taking a quick response to fix the issue. I did a quick test using the updated plugin in my system and it works fine :)

pjdarton:
OK, thanks for confirming the fix.
I've merged the code changes into master and this will be fixed in the next release.

pjdarton:
Code changes went into released of docker-plugin version 1.1.9

PR #752, opened 2019-10-01
Fix retries=0 bug in PortUtils.ConnectionCheckSSH

pjdarton:
* Issue #751 - Math.min should've been Math.max
* Removed unused imports from unit test
* Fixed failure-case unit test for ConnectionCheckSSH

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
@@ -157,7 +157,7 @@ public class PortUtils {
         public boolean execute() throws InterruptedException {
             checkState(parent.execute(), "Port %s is not opened to connect to", parent.port);
 
-            final int retries = Math.min(0, parent.retries);
+            final int retries = Math.max(0, parent.retries);
             final long retryDelay = parent.retryDelay;
             final int totalTriesIntended = retries + 1;
             int thisTryNumber;
@@ -172,16 +172,16 @@ public class PortUtils {
         }
 
         private boolean executeOnce(final int thisTryNumber, final int totalTriesIntended) {
-            final Connection connection = new Connection(parent.host, parent.port);
+            final Connection sshConnection = new Connection(parent.host, parent.port);
             try {
-                connection.connect(null, 0, sshTimeoutMillis, sshTimeoutMillis);
+                sshConnection.connect(null, 0, sshTimeoutMillis, sshTimeoutMillis);
                 LOGGER.info("SSH port is open on {}:{}", parent.host, parent.port);
                 return true;
             } catch (IOException e) {
                 LOGGER.error("Failed to connect to {}:{} (try {}/{}) - {}", parent.host, parent.port, thisTryNumber, totalTriesIntended, e.getMessage());
                 return false;
             } finally {
-                connection.close();
+                sshConnection.close();
             }
         }
     }
--- a/src/test/java/com/nirima/jenkins/plugins/docker/utils/PortUtilsTest.java
+++ b/src/test/java/com/nirima/jenkins/plugins/docker/utils/PortUtilsTest.java
@@ -76,19 +76,24 @@ public class PortUtilsTest {
     @Test
     public void shouldWaitIfPortAvailableButNotSshUntilTimeout() throws Exception {
         // Given
-        // e.g. try, delay, try, delay, try = 3 tries, 2 delays.
-        final long minExpectedTime = RETRY_COUNT * DELAY;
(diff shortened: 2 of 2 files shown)


src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
```diff
@@ -157,7 +157,7 @@ public ConnectionCheckSSH withSSHTimeout(int time, TimeUnit units) {
         public boolean execute() throws InterruptedException {
             checkState(parent.execute(), "Port %s is not opened to connect to", parent.port);
 
-            final int retries = Math.min(0, parent.retries);
+            final int retries = Math.max(0, parent.retries);
```
pjdarton: This is the bugfix.

src/main/java/com/nirima/jenkins/plugins/docker/utils/PortUtils.java
```diff
@@ -172,16 +172,16 @@ public boolean execute() throws InterruptedException {
         }
 
         private boolean executeOnce(final int thisTryNumber, final int totalTriesIntended) {
-            final Connection connection = new Connection(parent.host, parent.port);
+            final Connection sshConnection = new Connection(parent.host, parent.port);
```
pjdarton: I renamed the variable to make it clear that this was a SSH connection not merely a TCP connection that was being made.

src/test/java/com/nirima/jenkins/plugins/docker/utils/PortUtilsTest.java
```diff
@@ -7,18 +7,15 @@
 
 import java.io.IOException;
 import java.net.ServerSocket;
-import java.util.Date;
```
pjdarton: Removed some unused imports

src/test/java/com/nirima/jenkins/plugins/docker/utils/PortUtilsTest.java
```diff
@@ -7,18 +7,15 @@
...
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import static java.lang.System.currentTimeMillis;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.both;
+import static org.hamcrest.Matchers.equalTo;
```
pjdarton: Replaced use of `is` with `equalTo` as "is" is deprecated.

src/test/java/com/nirima/jenkins/plugins/docker/utils/PortUtilsTest.java
```diff
@@ -79,19 +76,24 @@ public void shouldThrowIllegalStateExOnNotAvailPort() throws Exception {
     @Test
     public void shouldWaitIfPortAvailableButNotSshUntilTimeout() throws Exception {
```
pjdarton: This was the buggy unit test that allowed the bug to go undetected.

pjdarton:
@php50350 @biozshock This is the proposed fix for issue #751.
Please download the `.hpi` file from the "Last Successful Artifacts" section in https://ci.jenkins.io/job/Plugins/job/docker-plugin/job/PR-752/, install it on your Jenkins server(s) (via manage jenkins -> manage plugins -> advanced), test that it fixes the bug, and let me know the results (either here or in #751).

Once the fix is confirmed I'll merge the changes.

biozshock:
@pjdarton the URL does not work for me:
$ telnet ci.jenkins.io 443
Trying 104.208.238.39...
telnet: Unable to connect to remote host: Connection refused

same with http

pjdarton:
@biozshock That's weird.  Maybe there's some internet firewall preventing you accessing that server? That server should be open to everyone on the internet...

Anyway, I've attached (a zip containing) the .hpi file here, [docker-plugin-1.1.9-rc895.e262f2331080.hpi.zip](https://github.com/jenkinsci/docker-plugin/files/3676335/docker-plugin-1.1.9-rc895.e262f2331080.hpi.zip).  Hopefully you'll be able to download it from here.
(I had to wrap it in a zip file because github says it doesn't allow uploads of .hpi files, even though these are really zip files by another name)

biozshock:
Seems to be working correctly here. Thanks for quick fix @pjdarton

dschnabel:
We ran into the same issue (#751) with 1.1.8 and the 1.1.9 RC as provided above fixed it for us. Thank you.

pjdarton:

Issue #612, opened 2018-02-01
RFE: "--no-cache" option for build / publish docker image build step


jbnance:
When building Docker images from the CLI one has the ability to pass the `--no-cache` option which instructs Docker to pull fresh copies of source images (the `FROM`) during the build.  Please add an equivalent option to the "Build / Publish Docker Image" build step.

curi0s:
Any news on this?

pjdarton:
Nope; nobody seems to want this enough to implement it.

If you want it and are willing to implement it, please create a PR - I'd be happy to review it and merge it in once it's complete.

curi0s:
I'm not a java developer sorry. Therefor I have to remove it and do it oldschool

PR #758, opened 2019-10-24
Add --pull and --no-cache options to Build/Publish Docker Image-Step

ybroeker:
This pull request add's the --no-cache and --pull options requested in #612


src/main/java/com/nirima/jenkins/plugins/docker/action/DockerBuildImageAction.java
```diff
@@ -29,27 +29,38 @@
...
+    public final boolean pull;
+
     @Deprecated
     public DockerBuildImageAction(String containerHost,
                                   String containerId,
                                   String taggedId,
                                   boolean cleanupWithJenkinsJobDelete,
-                                  boolean pushOnSuccess) {
+                                  boolean pushOnSuccess,
+                                  final boolean noCache,
```
pjdarton: Code style: "Try to fit in" - don't use `final` here.
While, personally, I'm a big fan of `finally` and use it every time I can, that hasn't been done in the rest of this class so, unless you want to first file a PR that reformats/re-styles this code to use `final` everywhere, please ensure that your new additions use the same code-style as is prevalent in the rest of this code.
See the "Coding style" section in the [contribution guidelines](https://github.com/jenkinsci/docker-plugin/blob/master/CONTRIBUTING.md) for more details.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -127,6 +127,9 @@
 
     public final String cloud;
```
pjdarton: Unnecessary blank line.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -288,7 +295,7 @@ boolean run() throws IOException, InterruptedException {
             log("Docker Build Response : " + imageId);
 
             // Add an action to the build
-            run.addAction(new DockerBuildImageAction(dockerApi.getDockerHost().getUri(), imageId, tagsToUse, cleanupWithJenkinsJobDelete, pushOnSuccess));
+            run.addAction(new DockerBuildImageAction(dockerApi.getDockerHost().getUri(), imageId, tagsToUse, cleanupWithJenkinsJobDelete, pushOnSuccess, noCache, pull));
```
pjdarton: This line is intolerably long now - please construct the new DockerBuildImageAction and store it in a local variable `action` and then call `run.addAction(action);`
That'll also make life easier if we ever add more fields in the future.

src/main/resources/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher/help-noCache.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Builds the image with --no-cache. Disables caching of layers.
```
pjdarton: Please surround `--no-cache` with `<code>` ... `</code>` to make it 100% clear that we're referring to a docker argument here.
I'd suggest that you also add the words "If set" to the start, e.g. "If set, builds the image with --no-cache which disables caching of layers." so that nobody can misunderstand things.  I know this and you know this, and I expect that most users will understand this, but I'm sure that some folks won't...

A link to the https://docs.docker.com/engine/reference/commandline/build/ docs would also help, as that way folks can see what specific "no-cache" argument you're referring to here (not all users of this plugin are familiar with the docker command line).

src/main/resources/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher/help-pull.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Builds the image with --pull. Pulls the latest version of the base image, instead of using the local one.
```
pjdarton: Same as with the noCache help - `<code>`, "If set..." and link to the docker docs.

src/main/java/com/nirima/jenkins/plugins/docker/action/DockerBuildImageAction.java
```diff
@@ -29,27 +29,38 @@
     public final boolean cleanupWithJenkinsJobDelete;
     public final boolean pushOnSuccess;
```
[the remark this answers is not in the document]
ybroeker: There were some blank lines between the existing fields, I just removed the ones added by me, should I also remove the other ones?

src/main/java/com/nirima/jenkins/plugins/docker/action/DockerBuildImageAction.java
```diff
@@ -72,6 +92,15 @@ public DescriptorImpl getDescriptor() {
         return (DescriptorImpl) Jenkins.getInstance().getDescriptorOrDie(getClass());
     }
 
+    public void setPull(boolean pull) {
```
pjdarton: Minor point: I'd prefer field getters/setters to immediately follow the constructor(s).
(I'm not going to refuse to merge the code if you don't do this, but if it's no hassle for you to move it, please do so)

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -346,6 +367,8 @@ public void onNext(BuildResponseItem item) {
             final String imageId;
             try(final DockerClient client = getClientWithNoTimeout()) {
                 imageId = client.buildImageCmd(tar)
+                        .withNoCache(noCache)
```
pjdarton: FYI _usually_ I'd want to guarantee that "not ticking the new checkbox" would mean that no new code got called at all (i.e. the `withNoCache` method & `withPull` method wouldn't get called at all); that way, there would be no chance that existing users who didn't ask for new stuff could be affected by any bugs in the new stuff.
However, I've (just now) checked the `dockerjava` code and it looks like its `AbstrDockerCmdExec.bool` method treats both `null` and `false` as equivalent so, in this particular case, we can get away with _always_ calling `withNoCache` and `withPull`.

src/main/resources/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher/help-noCache.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    If set, builds the image with <code>--no-cache</code> which disables caching of layers.
+    See docs.docker.com/engine/reference/commandline/build for documentation.
```
pjdarton: This should be a hyperlink to the docs. e.g. `See the docker <a href="https://docs.docker.com/engine/reference/commandline/build/">build command</a> for more information.`, which should end up looking something like
> See the docker [build command](https://docs.docker.com/engine/reference/commandline/build/) for more information.

src/main/resources/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher/help-pull.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    If set, builds the image with <code>--pull</code> to pull the latest version of the base image, instead of using the local one.
+    See docs.docker.com/engine/reference/commandline/build for documentation.
```
pjdarton: Should be a link.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher.java
```diff
@@ -21,9 +21,7 @@
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractProject;
-import hudson.model.Item;
-import hudson.model.TaskListener;
+import hudson.model.*;
```
pjdarton: Please undo these changes to imports (it helps to keep code churn to a minimum).
(I guess this was your IDE "tidying things up" - it seems that no two people's IDEs have the same settings :wink:)

src/main/resources/com/nirima/jenkins/plugins/docker/builder/DockerBuilderPublisher/config.jelly
```diff
@@ -39,4 +39,11 @@
         <f:checkbox default="false"/>
     </f:entry>
 
+    <f:entry title="${%Disable caching}" field="noCache">
+        <f:checkbox default="false"/>
+    </f:entry>
```
pjdarton: While I normally dislike "unnecessary" blank lines ... in this file (and in most of the jelly files in this plugin) the trend is to have a blank line between entries.
i.e. please add a blank line between line 44 and 45.

pjdarton:
My understanding is that it's only necessary to add `@DataBoundSetter` when extending a `@DataBoundConstructor` so, as there's no data-bound-anything in `DockerBuildImageAction` there isn't any need for the annotation on its setters.
i.e. the approach you've taken is correct.

pjdarton:

PR #761, opened 2019-11-17
Fix Docker Jenkins JNLP Agent image name

alecharp:
The repository is jenkinsci/docker-jnlp-slave but the Docker image name is jenkins/jnlp-slave.
Fixing this will prevent incorrect copy/paste for users who want to try it.

### DIFF
--- a/src/main/resources/io/jenkins/docker/connector/DockerComputerAttachConnector/config.jelly
+++ b/src/main/resources/io/jenkins/docker/connector/DockerComputerAttachConnector/config.jelly
@@ -10,7 +10,7 @@
             <li>
                 Entrypoint must be able to accept jenkins slave connection parameters.
                 See 
-                <a href="https://github.com/jenkinsci/docker-jnlp-slave">jenkins/docker-jnlp-slave</a>
+                <a href="https://github.com/jenkinsci/docker-jnlp-slave">jenkins/jnlp-slave</a>
                 as an example.
             </li>
         </ul>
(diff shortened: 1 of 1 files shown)


src/main/resources/io/jenkins/docker/connector/DockerComputerAttachConnector/config.jelly
```diff
@@ -10,7 +10,7 @@
             <li>
                 Entrypoint must be able to accept jenkins slave connection parameters.
                 See 
-                <a href="https://github.com/jenkinsci/docker-jnlp-slave">jenkins/docker-jnlp-slave</a>
+                <a href="https://github.com/jenkinsci/docker-jnlp-slave">jenkins/jnlp-slave</a>
```
pjdarton: I think that, if we're going to update this text, it'd be best if the link-text matched the link-URL so, given that the source code uses a different name to the docker container, it'd be better if we linked to both (that also means that, if things change again, users have twice the chance of tracking down the new location).

So I think we should say something like:
> See docker container [jenkins/jnlp-slave](https://hub.docker.com/r/jenkins/jnlp-slave/) and/or source [jenkinsci/docker-jnlp-slave](https://github.com/jenkinsci/docker-jnlp-slave).

alecharp:
@pjdarton I'm really really sorry for the force-push.. I hate to do that on PRs but I mis-wrote your handle on the first try.

pjdarton:
Sometimes a re-base and force-push is the right thing to do - in this case, the changes were small enough that it was not an inconvenience.

It all looks good now.

pjdarton:

PR #762, opened 2019-11-20
Fix SSH-connector unit-tests

pjdarton:
* The official ssh-slave (now) requires additional configuration for the
old unit test to work, as java is no longer found on the PATH when the
container is invoked as it is here.  That broke the injected-key test.
* There never was a unit-test for the specified-credentials method (only
the injected-key one) so that's been added too.
* The unit-test's cleanup code was a TODO statement, so it left things
in a mess that tended to prevent other tests from running properly.

### DIFF
--- a/src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
+++ b/src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
@@ -2,57 +2,136 @@ package io.jenkins.docker.connector;
 
 import com.nirima.jenkins.plugins.docker.DockerCloud;
 import com.nirima.jenkins.plugins.docker.DockerTemplate;
+
 import hudson.model.FreeStyleBuild;
 import hudson.model.FreeStyleProject;
 import hudson.model.Label;
+import hudson.model.Node;
 import hudson.model.Result;
+import hudson.model.TaskListener;
+import hudson.model.queue.QueueTaskFuture;
+import hudson.slaves.Cloud;
 import hudson.tasks.Shell;
+import hudson.util.StreamTaskListener;
+import io.jenkins.docker.DockerTransientNode;
 import io.jenkins.docker.client.DockerAPI;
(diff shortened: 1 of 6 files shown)


src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
```diff
@@ -2,57 +2,136 @@
...
+
+    @Before
+    public void setUp() {
+        testNumber++;
+        cloudName = "DockerCloudFor" + this.getClass().getSimpleName() + testNumber;
+    }
 
     @After
-    public void cleanup() {
-        // TODO rm orphaned container
```
pjdarton: That's the key thing being dealt with within this file ;-)

src/test/java/io/jenkins/docker/connector/DockerComputerSSHConnectorTest.java
```diff
@@ -1,32 +1,67 @@
...
-    @Test
-    public void should_connect_agent() throws InterruptedException, ExecutionException, IOException {
+    private static final String SSH_SLAVE_IMAGE_IMAGENAME = "jenkins/ssh-slave";
+    private static final String SSH_SLAVE_IMAGE_JAVAPATH = "/usr/local/openjdk-8/bin/java";
 
+    @Test
+    public void connectAgentViaSSHUsingInjectSshKey() throws Exception {
+        final DockerComputerSSHConnector.SSHKeyStrategy sshKeyStrategy = new DockerComputerSSHConnector.InjectSSHKey(COMMON_IMAGE_USERNAME);
+        final DockerComputerSSHConnector connector = new DockerComputerSSHConnector(sshKeyStrategy);
+        connector.setJavaPath(SSH_SLAVE_IMAGE_JAVAPATH);
```
pjdarton: The lack of this is what broke the test - the way this plugin invokes the container, the PATH doesn't have java on it within the SSH sessions (arguably a bug in the container image).

PR #763, opened 2019-11-21
[WIP] Improve support for default ssh docker container

pjdarton:
The [official Jenkins ssh container](https://hub.docker.com/r/jenkinsci/ssh-slave/) says it takes the SSH certificate (that it should accept) as the first (and only) argument.
The default docker-plugin behaviour is to explicitly specify the command (`/usr/sbin/sshd`...) to run sshd but (a) this doesn't work on Windows and (b) a [bug](https://github.com/jenkinsci/docker-ssh-slave/issues/33) in the container's entry-point script means that environment variables in the container don't get into the ssh daemon's environment, with the ultimate result that Java isn't on the $PATH, causing containers to fail to come online (unless you also set the "Java Path" to `/usr/local/openjdk-8/bin/java`).

This PR adds a new SSH connection method that provides the SSH certificate as the first argument to the container, which is exactly what the official ssh container expects.

Change summary:
- the `DockerComputerSSHConnector` class has been refactored to reduce commonality between its key-strategy inner classes.
- the `ManuallyConfiguredSSHKey` key-strategy inner class has been refactored but should be functionally unaltered.
- the `InjectSSHKey` key-strategy inner class now shows up on the WebUI as "Inject SSH key *using SSH AuthorizedKeysCommand option*" instead of just "Inject SSH key".
- the `InjectSSHKey` key-strategy inner class now _also_ sets the `JENKINS_SLAVE_SSH_PUBKEY` environment variable (in the container) to the injected key ... although that's unlikely to be much use to anyone using the [Java-8 Linux docker container](https://hub.docker.com/r/jenkinsci/ssh-slave/) because of a [bug](https://github.com/jenkinsci/docker-ssh-slave/issues/33).
- there's a new `InjectSSHKeyAsContainerArgument` key-strategy inner class that calls itself "Inject SSH key *as 1st container argument*"; this _uses no Linux-specific commands_ and therefore may well work for Windows users.

TODO:
- add help text that explains what each option does
- test it.


basil (commented):
A promising start! It is never too late to finish this draft PR.

basil:
@pjdarton Is this PR still needed? I was able to successfully create a new Docker cloud template whose image was `jenkins/ssh-agent:latest`, whose connection method was **Connect with SSH**, whose SSH key was **Inject SSH key**, and whose user was `jenkins`.

basil:
@Peter-Darton-i2 Looks like at least some of this PR is still needed as explained in #966.

pjdarton:
It's been a while since I looked at this but IIRC the official docker image wanted to be invoked with specific arguments and then it'd "just work" whereas this plugin kinda overrode everything to force the connection to work on its terms instead of the way the official docker image intended.
... and I think this PR was to support using the official image the way that image wanted.

... but I'm no longer in a position where I can test things out "in anger" the way I used to (I used to use this plugin for work at scale; I don't use it at all anymore) so it could be abandoned without any loss to me ... unless someone else wants this enough to drive this forward.

Issue #724, opened 2019-03-13
Add options to customize user, group, uid, gid in Docker Agent templates


devster31:
- [1.1.6] docker-plugin version
 - [2.150.3] jenkins version
 - [18.09.2] docker engine version

The current template options don't allow to customize user, group, uid or gid of the container being created.
This may cause issues when using `jenkins` to build docker images.
My current setup, for example, uses the default `jenkins/jnlp-agent-docker` which encounters the following error:
```
docker build -t <user>/<project> .
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.37/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&session=<sessionid>&shmsize=0&t=<user>%2F<project>&target=&ulimits=null: dial unix /var/run/docker.sock: connect: permission denied
```
which I believe could be avoided using `--user jenkins:<gid>` where `<gid>` corresponds to the group ID of the docker user on the host machine.

idueppe:
I would prefer if you could configure the `group-add` argument of `docker run`.

For instance, in the config are two group ids defined: <gid_1> and <gid_2> then the run command should look like:

```
docker run --user jenkins --group-add <GID1> --group-add <GID2> ...
 ```

I guess this would be more flexible.

ericcitaire:
Incremental builds are available :
* [docker-java-api](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/docker-java-api/3.1.5-rc11.59b91388afe6/) built from https://github.com/jenkinsci/docker-java-api-plugin/pull/5
* [docker-plugin](https://repo.jenkins-ci.org/incrementals/io/jenkins/docker/docker-plugin/1.1.10-rc915.f7c9b174bfb4/) built from https://github.com/jenkinsci/docker-plugin/pull/777

Please give it a try and let me know. Feel free to up-vote and comment the 2 PRs! :)

pjdarton:
I think this is complete, as the linked PRs are all merged.

PR #777, opened 2020-02-25
Docker java api 3.1.5

ericcitaire:


ericcitaire:
Relates to issue #724

ericcitaire:
Hi @pjdarton 

Did you have a chance to take a look at this?

Thanks.

pjdarton:
Not yet; out of the office at present and have a high priority task waiting for my return ... If I've not responded by the end of the month, feel free to nag me again.

pom.xml
```diff
@@ -56,6 +56,7 @@
         <java.level>8</java.level>
         <groovy.version>2.4.7</groovy.version>
         <jenkins.version>2.60.3</jenkins.version>
+        <spotbugs.failOnError>false</spotbugs.failOnError>
```
pjdarton: If we're now using spotbugs, we shouldn't retain the config for findbugs.
We should be using one or the other; we shouldn't be configuring both.

pom.xml
```diff
@@ -68,7 +69,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>docker-java-api</artifactId>
-            <version>3.0.14</version>
+            <version>3.1.5-rc11.59b91388afe6</version>
```
pjdarton: We shouldn't be depending on a rc build.
We should only depend on a "proper" release.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```diff
@@ -502,6 +502,7 @@ InspectImageResponse pullImage(DockerAPI api, TaskListener listener) throws IOEx
                 cmd.exec(new PullImageResultCallback() {
                     @Override
                     public void onNext(PullResponseItem item) {
+                        super.onNext(item);
```
pjdarton: This _looks_ like you're fixing an old bug here while doing everything else... Can you explain your reasoning here?
Was this line "always missing but should've been there since the outset" or has this only become an issue after updating to a later version of the docker-java code?

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -1,11 +1,7 @@
 package com.nirima.jenkins.plugins.docker;
 
 import com.github.dockerjava.api.command.CreateContainerCmd;
-import com.github.dockerjava.api.model.Bind;
-import com.github.dockerjava.api.model.PortBinding;
-import com.github.dockerjava.api.model.Volume;
-import com.github.dockerjava.api.model.VolumesFrom;
-import com.github.dockerjava.api.model.Device;
+import com.github.dockerjava.api.model.*;
```
pjdarton: Opinions vary on this subject but, personally, I'm not a fan of `.*` in imports, and I'm even less of a fan of unnecessary changes to source code.
Please just add the imports you need, one per line.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -663,14 +690,14 @@ public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig
             containerConfig.withExtraHosts(extraHosts.toArray(new String[extraHosts.size()]));
         }
 
-        if (shmSize != null && shmSize > 0) {
+        if (hostConfig != null && shmSize != null && shmSize > 0) {
             final Long shmSizeInByte = shmSize * 1024L * 1024L;
-            containerConfig.getHostConfig().withShmSize(shmSizeInByte);
+            hostConfig.withShmSize(shmSizeInByte);
```
pjdarton: Please undo these changes; they are unnecessary.
The established pattern in this code (and in the docker-java code) is that "everyone calls getHostConfig() every time they need it" rather than calling it once and reusing it.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -663,14 +690,14 @@ public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig
...
         }
 
         final List<String> securityOptions = getSecurityOpts();
-        if (CollectionUtils.isNotEmpty(securityOptions)) {
-            containerConfig.getHostConfig().withSecurityOpts( securityOptions );
+        if (hostConfig != null && CollectionUtils.isNotEmpty(securityOptions)) {
+            hostConfig.withSecurityOpts( securityOptions );
```
pjdarton: Please undo these changes; they are unnecessary.
The established pattern in this code (and in the docker-java code) is that "everyone calls getHostConfig() every time they need it" rather than calling it once and reusing it.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -552,6 +573,8 @@ public PortBinding apply(String s) {
     }
 
     public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig) {
+        HostConfig hostConfig = containerConfig.getHostConfig();
+
```
pjdarton: Please remove these lines.
The established pattern in this code (and in the docker-java code) is that "everyone calls getHostConfig() every time they need it" rather than calling it once and reusing it.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -560,6 +583,10 @@ public CreateContainerCmd fillContainerConfig(CreateContainerCmd containerConfig
             containerConfig.withUser(user);
         }
 
+        if (hostConfig != null && CollectionUtils.isNotEmpty(extraGroups)) {
+            hostConfig.withGroupAdd(extraGroups);
+        }
+
```
pjdarton: The established pattern in this code (and in the docker-java code) is that "everyone calls getHostConfig() every time they need it" rather than calling it once and reusing it.
Please follow the existing pattern.

FYI there is some logic to this "extra code" - if there's a problem with the getHostConfig() method then, if we only ever call getHostConfig() when we need it, anyone who doesn't use the affected functionality will be able to continue working; if we always call getHostConfig() regardless of whether or not we need it _and_ there's a problem with it then everyone is screwed.
The principal of "retain existing functionality" is key here - we don't want all existing users to be affected by this new functionality unless they decide to use it.

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java
```diff
@@ -781,6 +810,7 @@ public int hashCode() {
         result = 31 * result + (dockerCommand != null ? dockerCommand.hashCode() : 0);
         result = 31 * result + (hostname != null ? hostname.hashCode() : 0);
         result = 31 * result + (user != null ? user.hashCode() : 0);
+        result = 31 * result + (extraGroups != null ? extraGroups.hashCode() : 0);
```
pjdarton: This isn't a comment about _this_ line but github doesn't let me comment on lines that haven't been changed and this is the closest we get...

There's an inner class called `DescriptorImpl` and that needs to have a `public FormValidation doCheckExtraGroupsString(@QueryParameter String extraGroupsString)` method added to it.
The purpose of that code is to check that we can turn the user's text into something we can persist and use, and let the user know if they've typed rubbish into the field.
The method `doCheckVolumesFromString` might be a good one to use as a basis - it does a "can we parse this? if not, catch the exception and let the user know" test.

src/main/java/io/jenkins/docker/client/DelegatingDockerClient.java
```diff
@@ -5,60 +5,9 @@
...
-import com.github.dockerjava.api.command.TopContainerCmd;
-import com.github.dockerjava.api.command.UnpauseContainerCmd;
-import com.github.dockerjava.api.command.UpdateContainerCmd;
-import com.github.dockerjava.api.command.VersionCmd;
-import com.github.dockerjava.api.command.WaitContainerCmd;
+import com.github.dockerjava.api.command.*;
 import com.github.dockerjava.api.exception.DockerException;
-import com.github.dockerjava.api.model.AuthConfig;
-import com.github.dockerjava.api.model.Identifier;
+import com.github.dockerjava.api.model.*;
```
pjdarton: As previous: Please don't use `.*` in imports; just add the imports you need, one per line.

src/main/resources/com/nirima/jenkins/plugins/docker/DockerTemplateBase/help-extraGroupsString.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    Run a container with the specified additional groups (e.g. "docker run --group-add ...").
+    <br/>
+    Refer to the <a href="https://docs.docker.com/engine/reference/run/">docker run command reference</a> for the correct syntax.
+</div>
```
pjdarton: Please add a LF at the end of the file.

src/main/java/com/nirima/jenkins/plugins/docker/builder/DockerBuilderControlOptionRun.java
```diff
@@ -90,13 +94,15 @@ public DockerBuilderControlOptionRun(
...
         this.user = user;
+        this.extraGroupsString = extraGroupsString;
         this.bindPorts = bindPorts;
         this.memoryLimit = memoryLimit;
         this.memorySwap = memorySwap;
         this.cpuShares = cpuShares;
         this.shmSize = shmSize;
         this.bindAllPorts = bindAllPorts;
         this.macAddress = macAddress;
+        this.extraHostsString = extraHostsString;
```
pjdarton: This looks like feature creep to me - looks like you're adding support for extraHosts here, and that wasn't part of this PR's description.
We either need to enhance the description or split this change into its own PR.

src/test/java/com/nirima/jenkins/plugins/docker/DockerTemplateTest.java
```diff
@@ -30,6 +30,7 @@
     String environmentsString = "environmentString";
     String hostname = "hostname";
     String user = "user1";
+    String extraGroupsString = "foo\nbar";
```
pjdarton: This isn't a comment about this line, it's just close-by...

We should also have unit tests added to `DockerTemplateBaseTest` that verify that
1) fillContainerConfig does _not_ set any "extra groups", not even setting things to null/empty, if the user hasn't given any extra groups.
2) fillContainerConfig _does_ set the extra groups that the user specified.
See `testFillContainerShmSize` and the methods that call it for an example.

pom.xml
ericcitaire: I had a lot of spotbugs errors after upgrading the parent pom. Fixing these errors has nothing to do with this PR. And I don't have the time to do it. That's why I just configured it to not fail the build.

I had to upgrade the parent pom because some dependencies are now using Multi-Release JARs, which was not supported with the older version of the parent pom.

pom.xml
ericcitaire: I know. I'm waiting for [this other PR](https://github.com/jenkinsci/docker-java-api-plugin/pull/5) to be merged. But it seems that maintainers of docker-java-api-plugin are all retired...

pom.xml
pjdarton: I agree that fixing all the spotbugs errors isn't part of this PR; that's not what I'm asking for here.

What I was complaining about was that, after switching _from_ findbugs _to_ spotbugs, we need to _remove_ the old "tell findbugs not to fail on error" config and replace it with a "tell spotbugs not to fail on error" config.
At present, we've got two "tell ... not to fail on error" configs in the pom.xml - if we're not using findbugs anymore then we can (and should) remove the findbugs-specific configuration.
(but we should probably retain the TODO comment, as fixing all of findbugs/spotbugs' complaints is something that ought to be done at some point, just not as part of this PR)

i.e. remove the findbugs config (line 226 in the original pom) and add a `<!-- FIXME need to review spotbugs errors -->` comment to the spotbugs config.

pom.xml
pjdarton: FYI Nicolas De loof backed away from all-things-Jenkins some time ago.  We'd worked together on this plugin (after he'd added me as a co-maintainer) but it was a bit of a surprise for me to find I was "suddenly" the sole maintainer rather than co-maintainer of _this_ plugin.

Looking at the commit history for the `docker-java-api-plugin`, it looks like `olivergondza` has commit access; maybe he can help.  If not, there's a Jenkins process you can follow to get control of an abandoned plugin; you can become its maintainer and push changes through yourself.

(FYI this "maintainers are all retired" scenario is not unusual; of all the bits of Jenkins I've worked on, I've only avoided being given commit-access twice - "no good deed goes unpunished" :wink:)

src/main/java/io/jenkins/docker/client/ListContainersCmdExec.java
```diff
@@ -6,9 +6,9 @@
 import com.github.dockerjava.api.command.ListContainersCmd;
 import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.core.DockerClientConfig;
-import com.github.dockerjava.netty.MediaType;
-import com.github.dockerjava.netty.WebTarget;
-import com.github.dockerjava.netty.exec.AbstrSyncDockerCmdExec;
+import com.github.dockerjava.core.MediaType;
+import com.github.dockerjava.core.WebTarget;
+import com.github.dockerjava.core.exec.AbstrSyncDockerCmdExec;
```
pjdarton: I _suspect_ that, now we're using the official `NettyDockerCmdExecFactory` instead of our own version, there's nothing that references this class either.
The reaon we needed this class was because there was a bug in the `docker-java` code where the Netty implementation (accidentally) depended on a Jersey `FilterEncoder` class which broke it for Jenkins usage.
(see the comments in this file for details)

So we might well be able to delete this class too.

Note: We would need to test that we didn't get nasty exceptions when trying to ask "so, how many containers are we running then?", as that was the bug this class was here to fix.  If that functionality works without this class being used then that'd tell us we don't need this class anymore.

ericcitaire: Ok, I remove it since it's not used anymore.
Can you give me a simple way to reproduce the bug it was supposed to fix, so I can test it?

src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
ericcitaire: It was already missing before the docker-java update. But the underlying code has change a lot in docker-java 3.1.5 and this missing line is now causing a *Could not pull image* error (see https://github.com/docker-java/docker-java/blob/3.1.5/src/main/java/com/github/dockerjava/core/command/PullImageResultCallback.java#L94)

pom.xml
ericcitaire: According to [this comment](https://github.com/jenkinsci/docker-java-api-plugin/pull/4#issuecomment-589619984), `olivergondza` is not a maintainer of docker-java-api-plugin.

I don't want to become a maintainer, I just want to contribute this feature.

It is scary that maintainers are giving up on Jenkins. Is Jenkins dead??

src/main/java/io/jenkins/docker/client/ListContainersCmdExec.java
pjdarton: The bug caused an inability to do a "list instances with these labels" query.  The code uses labels to distinguish between containers it started and anything else running within the docker host.

To test:
1. Configure a single cloud with a single template.
1. Set a cloud instance limit of 2, template instance max of 1.
1. Have a build require that template and trigger the build.

With the bug, when the `DockerCloud` code tries to ask the docker host "how many instances do we have running in total" (method `countContainersInDocker`), there would be a big ClassNotFound exception.
...and when it tries to ask the docker host "how many instances do we have of this particular template", it'd get the same exception.
If you don't set any cloud-instance limits or template instance limits then the code doesn't get called and you're immune to the bug (but can't test for it); with instance-count limits set _and_ the bug present, it'll throw and be unable to provision a new container, and hence be unable to start a slave and execute the build.

If the code works, the build should run ok and you'll see a message (from the `canAddProvisionedSlave` method) logged along the lines of `"Provisioning '{}' number {} (of {}) on '{}'; Total containers: {} (of {})"`, where the numbers "in use" came from the list-containers command result.
If we get that log message, it means that the list-containers command worked.

pom.xml
pjdarton: Jenkins isn't dead, no; Jenkins is an open-source project and its life is independent of any single individual ... but not everyone who becomes involved in its development enough to become a contributor _stays_ involved forever, e.g. people change jobs.
Progress is mostly driven by folks who have a need to make changes and are prepared to put a bit of work in to make that happen; for some components, there's sometimes _years_ that go by without anyone needing to change anything and, during that time, whoever maintained it last may disappear off the radar.
That seems to be the case with the docker-java-api-plugin - it's been stable for so long it didn't need an active maintainer until you stirred things up 😛 

e.g. the vSphere plugin was started by one person, continued by another and, when I wanted to add to it, neither of those two people wanted any involvement with it anymore and basically told me "it's yours now!", which is how I became the maintainer of the vSphere plugin :wink:
When I wanted to add to the docker plugin, Nickolas was actively developing it and was happy to have a second contributor ... but then he left, leaving only me.
When I wanted to contribute to the OpenStack plugin, I was lucky - Oliver was happy to stay sole maintainer of that one.

FYI this is why the Jenkins project _overall_ has a "I want to contribute to this plugin but nobody's actively maintaining it anymore" process whereby an individual can get control of an abandoned plugin - it allows any given component to outlive the person who started it _without_ that person actively handing over controls to a successor.

As for becoming a maintainer, well, you don't have to do it forever (as Nicolas has demonstrated) - you take it over, get up to speed on it, drive this change to completion and then, once we're all happy that it's working and no further changes are needed, you can leave it without a maintainer - it's not exactly a high-traffic plugin.

ericcitaire:
Hi @pjdarton.

The build succeeded, but the incremental build was not uploaded to https://repo.jenkins-ci.org/incrementals/io/jenkins/docker/docker-plugin/.

According to the build log, it failed with a `403 Forbidden`, without marking the step as failed (`-f` option is missing from `curl` command).

Can you help me with this? Thanks.

pjdarton:
One trick I've picked up is that if you close a PR, wait 10 minutes and then re-open the (same) PR (using the re-open button - don't make a new one!) then the Jenkins CI build will re-run.
This lets you retry a build which failed for infrastructure reasons.

docwhat:
@ericcitaire Should I file a bug with docker-java for not following [semantic versioning](https://semver.org)? Or add a  comment someplace if you think it is too late to worry about that, since it was in 3.1.0 over a year ago?

Because when they merged docker-java/docker-java#980 they didn't bump the major number from 3 to 4 when they renamed `WebTarget` to `NettyWebTarget` they broke the API.

pjdarton:
@docwhat I don't think there's any point raising a bug w.r.t. versioning - it's too late to worry - there's nothing (reasonable) that can be done about it now.

src/main/java/io/jenkins/docker/client/ListContainersCmdExec.java
ericcitaire: I tested it, the build runs OK and I see the appropriate message in the log.

pjdarton: As you've probably noticed, I've merged this PR "as is" because we need the "work with docker-java-api-plugin version 3.1.5" changes _now_ so I'm hoping that it'll all work.
I did some basic smoke-tests locally before merging, and it all seemed ok.
...but then the automated Jenkins-CI build failed for seemingly-unrelated reasons ☹️.  Twice 😠 

Given that the release of docker-java-api-plugin has broken the world, we've got little to lose by doing a release, so that's what I'm doing right now...

pjdarton:
Merged and released as version 1.2.0
I decided to bump the middle version number because of the somewhat-breaky-changes we're introducing/coping-with here.
That was uploaded a few minutes ago - it should start appearing in the main Jenkins plugins as an available upgrade in the next few hours.

PR #770, opened 2020-01-20
Feature/thread start container jnlp

mat1e:
See #757 and [JENKINS-59790](https://issues.jenkins-ci.org/browse/JENKINS-59790).

With JNLP connection method, adding a Jenkins node is asynchronous with when the container starts.

It is a problem when Jenkins is slow and the docker container start very quickly. The node is not yet created on Jenkins when the container ask to connect.

This solution delays the container start until the Jenkins node is created for JNLP.

PR #789, opened 2020-04-15
Fix #757 and JENKINS-59790

### LINKED ISSUE #757: Container cannot connect to node because it doesn't exist
We recently updated our version of Jenkins to 2.176.3. And now a connection error with docker-agent randomly block the queue of jobs :

`Refusing headers from remote: Unknown client name: docker-00026wu6nor9w`

The docker container is ready and try to connect to the Jenkins master but the node doesn't exist yet.

I saw in the code of docker-plugin that the container is created and started before the Jenkins node. While the connection method is JNLP, the commands to download and run the remoting.jar are executed at the start of the container. But at this moment, the node wasn't added to Jenkins master.

Have you ever encountered this error? Is there a solution?

Is it possible to modify provision methods and create the Jenkins node before instanciate the container to fix this issue?

Jenkins version : 2.176.3
docker-plugin version : 1.1.7
docker engine version : 1.13.1

pjdarton:
Fix #757 and JENKINS-59790.
Alternative implementation to #770.

Refactor DockerTransientNode constructor so the launcher is provided post-construction.
Refactor DockerComputerConnector.beforeContainerStarted(...) so it takes a node (that holds a containerId) instead of a containerId.
Refactor DockerTemplate.doProvision(...) so that the DockerTransientNode is created without a launcher before we start the container.
Enhance DockerComputerJNLPConnector so that it adds the node to Jenkins immediately before the container is created, ensuring that it'll be known to Jenkins (just) before the JNLP process starts.

SSH and DirectAttach connectors continue as normal.
JNLP ones call `Jenkins.getInstance().addNode(dockerNode)` before the container starts, ensuring that the container is known to Jenkins before the JNLP process can start and attempt to connect.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
@@ -716,12 +716,7 @@ public class DockerTemplate implements Describable<DockerTemplate> {
         LOGGER.info("Started container ID {} for node {} from image: {}", containerId, nodeName, ourImage);
 
         try {
-            ourConnector.beforeContainerStarted(api, effectiveRemoteFsDir, containerId);
-            client.startContainerCmd(containerId).exec();
-            ourConnector.afterContainerStarted(api, effectiveRemoteFsDir, containerId);
-
-            final ComputerLauncher nodeLauncher = ourConnector.createLauncher(api, containerId, effectiveRemoteFsDir, listener);
-            final DockerTransientNode node = new DockerTransientNode(nodeName, containerId, effectiveRemoteFsDir, nodeLauncher);
+            final DockerTransientNode node = new DockerTransientNode(nodeName, containerId, effectiveRemoteFsDir);
             node.setNodeDescription("Docker Agent [" + ourImage + " on "+ api.getDockerHost().getUri() + " ID " + containerId + "]");
             node.setMode(getMode());
             node.setLabelString(getLabelString());
@@ -730,6 +725,11 @@ public class DockerTemplate implements Describable<DockerTemplate> {
             node.setRemoveVolumes(isRemoveVolumes());
             node.setStopTimeout(getStopTimeout());
             node.setDockerAPI(api);
+            ourConnector.beforeContainerStarted(api, effectiveRemoteFsDir, node);
+            client.startContainerCmd(containerId).exec();
+            ourConnector.afterContainerStarted(api, effectiveRemoteFsDir, node);
+            final ComputerLauncher nodeLauncher = ourConnector.createLauncher(api, containerId, effectiveRemoteFsDir, listener);
+            node.setLauncher(nodeLauncher);
             finallyRemoveTheContainer = false;
             return node;
         } finally {
--- a/src/main/java/io/jenkins/docker/DockerTransientNode.java
+++ b/src/main/java/io/jenkins/docker/DockerTransientNode.java
@@ -35,7 +35,6 @@ public class DockerTransientNode extends Slave {
     private static final long serialVersionUID = 1349729340506926183L;
     private static final Logger LOGGER = LoggerFactory.getLogger(DockerTransientNode.class.getName());
 
-    //Keeping real containerId information, but using containerName as containerId
     private final String containerId;
 
     private transient DockerAPI dockerAPI;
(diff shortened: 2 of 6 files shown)


pjdarton:
@mat1e FYI I'm trying to figure out why the unit tests keep failing... and not just for this PR; seems to be an issue with the DockerNodeStepTest and/or with the unit-tests running docker containers that consume excessive amounts of memory and then fail with an out-of-memory issue.
Hopefully I'll fix both; if not, might just have to merge this one in "as is".

While I'm trying to sort that ... could you try out this PR and let me know if it fixes the issue you were encountering "in the real world"?
I don't hit the problem "for real" in my own environment so all I know is that this PR doesn't break the unit tests (at least, no more than they are already); I don't know if it really fixes the problem that you hit...

mat1e:
@pjdarton I'm working on scheduling tests on this solution. It will take time to validate that it corrects the problem (few weeks I think). And it is difficult to do any change on ours environments actually.

mat1e:
@pjdarton,
Sorry for the delay, our infra team don't want to take risks installing a "custom build" of the plugin. For this reason it is going to pass on all of ours environnments. I'm trying to speed things up.

pjdarton:
I fully understand why infra teams want to be cautious (I have my own infrastructure and I'm cautious with that too!) but, in this case, by deciding to avoid testing it before it becomes mainstream, they're putting themselves (and others) at more risk, not less.
What I'm doing here is letting you folk get early access to stuff before you've got no choice about it; it's more of an opportunity to _reduce_ risk than take risk.

FYI what I do, if I've written a bunch of new code for this plugin, is to try it out on my "least important" Jenkins server first, and progressively push it onto the others until it's eventually on my most-important server(s), being prepared to yank it out and revert to the previous version if there's any big problems.
You should tell 'em that, with a "custom build", they have the option of going back to a mainstream one _and ensuring the code doesn't go mainstream if they don't like it._  If they aren't prepared to test it before it becomes mainstream their ability to go back is somewhat more constrained as the functionality will be in the mainstream version of the plugin (and all future ones).

You might also want to remind your infra team that the only difference between a "custom build" and a "mainstream" one is the version number - there's no _extra_ QA for a mainstream release as copared with a custom one ... which is why any decent infra team will know how to revert a plugin version in case of problems; using a custom build carries no more risks than any other plugin update.

TL;DR: This plugin software is FOSS; I'm not being paid to help you.  If you (and your colleagues) want something (e,g, this PR), you will (collectively) need to help with dev _and test._

cpfeiffer:
I can also recommend having a dedicated test-instance of Jenkins for trying out new releases, new plugins, ... It's really easy to do if it's all Docker containers anyway.

pjdarton:
The problem, IME, with a "dedicated test-instance" is that it doesn't get "real world" usage but only "test usage", and murphy's law dictates that not only will there will be a difference between "real world usage" and "test usage", but it'll also mean the difference between "bug showing up" and "bug remaining hidden".

One can test in an isolated environment as much as one likes (and FYI that's what the unit tests do) but the only "real proof" is "does it work in the real world".
That's why IMO it's important that the folks pushing for a change test that the proposed changes do actually fix something "in the real world" ... especially when the unit tests don't notice any difference...

mat1e:
We have many test instances but, it used only by us or testers. Exactly like @pjdarton said, we got this bug on the instances used by many "real users" for "real projects". The usage and the load of the server is important in this case. For this reason, we want to validate the fix on our finals environments. But infra team have to keep the integrity and the security of it and have to respect somes process. Actually, the custom build is going to pass all tests and have to be validated on "tests instances". The only thing I can do is to try to reproduce the bug on a "ISO Prod" environment like pre-production and verify that the bug is fixed by the new version. For now, I have the pipeline we used to analyse the bug, but I need the old docker image used by the agent (because we updated it and we put the retry in the entry point). I keep you informed.

mat1e:
I think we misunderstood each other. The process for testing the plugin is ongoing since you asked me to do it. It is just very long and it was delayed by covid-19... I'm just trying to accelerate by doing the validation myself.

pjdarton:
Folks, is there any progress to report?
I'm just starting to emerge from a multi-month long [very tedious and time consuming] bit of work and now have a chance to do maintenance stuff on this plugin, so if we have a "yes" result on this, now's the time to say so...

pjdarton:
So, FYI, I've installed this on a couple of Jenkins servers and it doesn't seem to have broken anything.
I can't comment on whether or not it fixes this problem, but it doesn't seem to break my environment either, which is cool.

mat1e:
The plugin with this fix is installed on our production environment since a month now and it is working well. I consider that the fix works.

mat1e (approved):

pjdarton:
@mat1e Thanks for testing this - that's excellent news.
I'll merge it.

PR #799, opened 2020-04-28
Fixing groovy configuration script for latest DockerTemplateBase

adamjwsuch:
Changes to DockerTemplateBase prevent the example groovy script working.

For reference/those searching for the error, like me, it gives this error:
`Could not find matching constructor for: com.nirima.jenkins.plugins.docker.DockerTemplateBase`

I have made the updates to the groovy script to match the latest DockerTemplateBase


pjdarton:
This script contains tech-debt - it's calling a deprecated DockerTemplateBase constructor.  I would prefer the script changed so that it doesn't use that deprecated constructor at all; that'll then ensure that it doesn't have to be kept in step every time someone introduces anything new.

The "official" constructor for DockerTemplateBase takes just the String image; everything else is set by setter methods (which you can call implicitly from groovy).
You may even be able to use `as DockerTemplateBase` on the map and let groovy figure out what constructor to use.

TL;DR: If you're updating this script, make it use the non-deprecated constructor instead of perpetuating the tech-dept so that the issue is fixed once and for all.

docs/attachments/docker-plugin-configuration-script.groovy
```diff
@@ -56,11 +58,13 @@ DockerTemplateBase dockerTemplateBase = new DockerTemplateBase(
   dockerTemplateBaseParameters.volumesFromString,
   dockerTemplateBaseParameters.environmentsString,
   dockerTemplateBaseParameters.hostname,
+  dockerTemplateBaseParameters.user,
```
pjdarton, line 61: The constructor used here was deprecated in 2017; we need to stop using it.
Call `new DockerTemplateBase(dockerTemplateBaseParameters.image)` and then set all the remaining parameters.
...or try something like `dockerTemplateBaseParameters as DockerTemplateBase` or similar groovy "magic" to set everything.

...and either remove or update the `// https://github.com/jenkinsci/docker-plugin/blob/docker-plugin-1.1.2/src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplateBase.java` comment that's pointing to an ancient version of the plugin.

adamjwsuch: I agree if the DockerTemplateBase constructor is deprecated it makes sense to update the script properly. My groovy skills do not go far enough for that.
I hit upon the script from the official docs but it just errored out.

maxim-nazarenko: hey @adamjwsuch, I have created a [PR](https://github.com/adamjwsuch/docker-plugin/pull/1) in your fork which addresses deprecation stuff.

Would you be able to review and merge it, please?

PR #800, opened 2020-05-08
Cloud stats support

theshadow27:
Add support for the Cloud-Stats jenkins plugin (https://plugins.jenkins.io/cloud-stats/) which enables monitoring of deployments. The usage pattern was copied directly from https://github.com/KostyaSha/yet-another-docker-plugin/blob/8552d5aef2c52554c0652087b75cb056a477d754/yet-another-docker-plugin/src/main/java/com/github/kostyasha/yad/DockerCloud.java , although I have limited the use of TrackedItem and CloudStatistics references into a new package accessed through the Factory pattern such that the dependency is optional, and gracefully falls back to the original implementation if the Cloud Stats plugin is not available.


src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -28,6 +28,7 @@
 import jenkins.authentication.tokens.api.AuthenticationTokens;
 import jenkins.model.Jenkins;
 import org.apache.commons.codec.binary.Base64;
+import org.jenkinsci.plugins.cloudstats.ProvisioningActivity;
```
pjdarton, line 31: In the PR description, you state that the dependency on the cloudstats plugin is optional.
Here, we've got an import that'll cause this class (which is core to this plugin) to fail to classload if the cloudstats plugin is not available.
i.e. either the description needs work, or this code does.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -67,6 +69,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DockerCloud.class);
 
+    static final DockerNodeFactory NODE_FACTORY = DockerNodeFactory.getInstance();
```
pjdarton, line 72: If this doesn't need to be accessed from outside this class, it should be private.
If it does need to be accesserd from outside this class, it should be `@Restricted(NoExternalUse.class)` to prevent its usage from outside the plugin.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -122,6 +125,7 @@
     /** Length of time, in seconds, that {@link #disabled} should auto-disable for if we encounter an error. */
     private @CheckForNull Integer errorDuration;
 
+
```
pjdarton, line 128: I'm not keen on unnecessary blank lines, or unnecessary changes to whitespace.
Please undo this line.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
                         t.getImage(), label, getDisplayName());
 
                 final CompletableFuture<Node> plannedNode = new CompletableFuture<>();
-                r.add(new NodeProvisioner.PlannedNode(t.getDisplayName(), plannedNode, t.getNumExecutors()));
+                DockerNodeFactory.DockerPlannedNode pn =
```
pjdarton, line 368: `pn` should be declared `final` unless it needs to be mutable ... and it doesn't look like it needs to be mutable.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
...
+                        NODE_FACTORY.createPlannedNode(t.getDisplayName(), plannedNode, t.getNumExecutors(),
+                                getDisplayName(), t.getImage(), label.getName());
 
+                r.add(pn);
```
pjdarton, line 372: Please move this line up on so `r.add` is next to the call to `NODE_FACTORY`.
The "add" belongs more to the node's creation than it does to the creation of the new runnable.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
...
                 final Runnable taskToCreateNewSlave = new Runnable() {
                     @Override
                     public void run() {
                         DockerTransientNode slave = null;
                         try {
-                            // TODO where can we log provisioning progress ?
+                            // TODO where can we log provisioning progress ? // In the CloudMonitor API!
```
pjdarton, line 378: If you've resolved the TODO comment, please remove the TODO comment and replace it with a comment explaining where we log provisioning progress.
i.e. once it's done, it's no longer TODO :wink:

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
...
+                            pn.notifyStarted(); // get().onStarted(id);
```
pjdarton, line 379: Why the commented out code?  What does that signify?
(I generally don't like commented out code - it smells of tech-debt to me - by all means add comments, but preferably not in the form of commented out code that has no context or explanation)

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
...
                             final DockerAPI api = DockerCloud.this.getDockerApi();
                             slave = t.provisionNode(api, TaskListener.NULL);
                             slave.setDockerAPI(api);
                             slave.setCloudId(DockerCloud.this.name);
-                            plannedNode.complete(slave);
+                            pn.notifySuccess(slave); //get().onComplete(id, slave);
```
pjdarton, line 384: As above; either have a human comment or no comment; commented-out code looks like it's a work in progress that needs fixing.

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -361,26 +365,32 @@ int countContainersInProgress() {
...
 
+                            plannedNode.complete(slave);
                             // On provisioning completion, let's trigger NodeProvisioner
                             robustlyAddNodeToJenkins(slave);
 
                         } catch (Exception ex) {
                             LOGGER.error("Error in provisioning; template='{}' for cloud='{}'",
                                     t, getDisplayName(), ex);
+                            pn.notifyFailure(ex); // get().onFailure(id, ex);
```
pjdarton, line 393: As above; either have a human comment or no comment; commented-out code looks like it's a work in progress that needs fixing.

src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+import hudson.slaves.NodeProvisioner;
+import io.jenkins.docker.DockerTransientNode;
+import org.jenkinsci.plugins.cloudstats.ProvisioningActivity;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.concurrent.Future;
+
+public interface DockerNodeFactory extends Serializable  {
```
pjdarton, line 18: Whitespace: should be one (and only one) space before the `{`

Later lines have similar issues w.r.t. inconsistent whitespace.

src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+    public DockerTransientNode createTransientNode(
+            @Nonnull String nodeName,
+            @Nonnull String containerId,
+            @Nonnull String effectiveRemoteFsDir,
+            @Nonnull ComputerLauncher launcher)
+            throws Descriptor.FormException, IOException;
+
+
+    public class DockerPlannedNode extends NodeProvisioner.PlannedNode {
+
```
pjdarton, line 86: We don't need this blank line.

src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+        public void notifyFailure(Throwable e){
+            // does nothing
+        }
+
+        public void notifySuccess(Node node){
+            // does nothing
+        }
+
+
```
pjdarton, line 103: We don't need these two blank lines.

src/main/java/com/nirima/jenkins/plugins/docker/cloudstat/TrackedDockerComputer.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public void recordTermination() {
+        try {
+            super.recordTermination();
+        } finally {
+            _terminate();
+        }
+    }
+
+    private void _terminate(){
```
pjdarton, line 42: Whitespace: should be one (and only one) space before the `{`

src/main/java/com/nirima/jenkins/plugins/docker/cloudstat/TrackedDockerPlannedNode.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final ProvisioningActivity.Id id;
+
+    @Restricted(NoExternalUse.class)
+    public TrackedDockerPlannedNode(String displayName, Future<Node> future, int numExecutors,
+                                    String cloud, String template, String node) {
+        super(displayName, future, numExecutors);
+        this.id = new ProvisioningActivity.Id(cloud,template,node);
+    }
+
```
pjdarton, line 26: Please limit to just one blank line between methods.

src/test/java/com/nirima/jenkins/plugins/docker/DockerNodeFactoryTest.java
```diff
@@ -0,0 +1,15 @@
...
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+
+public class DockerNodeFactoryTest {
+    @Test
+    public void testGetInstance() {
+        DockerNodeFactory dnf = DockerNodeFactory.getInstance();
+        assertThat(dnf, is(instanceOf(CloudStatsFactory.class)));
+    }
+}
```
pjdarton, line 15: Please add a EOL to this final line.

src/test/java/com/nirima/jenkins/plugins/docker/cloudstat/CloudStatsFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+public class CloudStatsFactoryTest {
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @BeforeClass
+    public static void setUpClass() {
+        JenkinsUtils.setTestInstanceId(CloudStatsFactoryTest.class.getSimpleName());
+
```
pjdarton, line 28: Remove blank line.

src/test/java/com/nirima/jenkins/plugins/docker/cloudstat/CloudStatsFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+        CloudStatsFactory uut = new CloudStatsFactory();
+        ComputerLauncher launcher = mock(ComputerLauncher.class);
+        DockerTransientNode result = uut.createTransientNode("nodeName","containerId","/foo/bar", launcher);
+
+        assertNotNull(result);
+        verifyNoMoreInteractions(launcher);
+        assertThat(result, is(instanceOf(TrackedItem.class)));
+
+    }
+}
```
pjdarton, line 56: Please all an EOL to this final line.

src/test/java/com/nirima/jenkins/plugins/docker/cloudstat/CloudStatsFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+    @Test
+    public void createTransientNode() throws Throwable{
+        CloudStatsFactory uut = new CloudStatsFactory();
+        ComputerLauncher launcher = mock(ComputerLauncher.class);
+        DockerTransientNode result = uut.createTransientNode("nodeName","containerId","/foo/bar", launcher);
+
+        assertNotNull(result);
+        verifyNoMoreInteractions(launcher);
+        assertThat(result, is(instanceOf(TrackedItem.class)));
+
```
pjdarton, line 54: Please remove this blank line.

pjdarton:
PS. Sorry for the delay in looking at this; other work issues got right in the way of this.

pjdarton:
PS. How does this compare with #797 ?
I presume that we only need one of these two implementations to be merged - one obsoletes the other...
I do like the idea of an optional dependency over a mandatory one ... but I don't like the idea of an _untested_ optional dependency - unless we can figure out a way to test how the code handles without the cloud stats plugin (so we can be sure it's as optional as you say), it'll have to be mandatory (I don't like untested promises).

src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+public interface DockerNodeFactory extends Serializable  {
+
+    /**
+     * Safely load the instance of the DockerNodeFactory
+     * @return a new instance of a implementation of DockerNodeFactory
+     */
+    static DockerNodeFactory getInstance() {
+        try{
+            return (DockerNodeFactory)Class.forName("com.nirima.jenkins.plugins.docker.cloudstat.CloudStatsFactory").newInstance();
```
Typz, line 26: can't we use `[@Extension(optional = true)](https://www.jenkins.io/doc/developer/plugin-development/optional-dependencies/)` annotation on CloudStatsFactory instead, and check for available descriptors?

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
theshadow27: The JLS requires classloading to be deferred until the class is referenced by another class (though most classloaders are lazy, and defer loading even longer until a need for class initialization occurs). 

In this proposal, there are no *direct* references to `com.nirima.jenkins.plugins.docker.cloudstat.*` package from the entry point - the only reference made is indirect (reflection) loading in [`src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java`](https://github.com/jenkinsci/docker-plugin/compare/master...theshadow27:cloud-stats-support#diff-5d3d42111eb185179cd4f284f9f18318R26), on line 26:
```
return (DockerNodeFactory)Class.forName("com.nirima.jenkins.plugins.docker.cloudstat.CloudStatsFactory").newInstance();
```
This is the first (and only) time that the `CloudStatsFactory` class is initialized. You are correct that this class references the optional dependencies which may be missing, and will fail to class load if they are not there - but the `try` block around this line ensures that a default stub implementation is used instead in that case.

src/main/java/com/nirima/jenkins/plugins/docker/DockerNodeFactory.java
theshadow27: Yeah that looks like the correct approach. This was my first time working in a Jenkins plugin so please forgive the lack of Jenkins-ey-ness :) 

Regarding testing... I'm not sure that it is possible to unit test "without" a dependency marked optional, unless an external driver re-writes the POM from optional -> compile

theshadow27: Also, I don't think our DevOps guys got my branch to work, and I didn't have time to debug... This might be why...

src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
theshadow27: This is the code that the other docker plugin used, to make sure I didn't miss any. I can take it out if it is annoying

pom.xml
```diff
@@ -235,41 +240,41 @@
...
-        </activation>
-        <dependencies>
-            <dependency>
-                <groupId>org.jenkins-ci.plugins</groupId>
-                <artifactId>command-launcher</artifactId>
-                <version>1.2</version>
-                <scope>test</scope>
-            </dependency>
-        </dependencies>
-      </profile>
```
theshadow27, line 272: This was my bad - reflexive twitch. I'll revert when I get a chance to come back to this, and try to clean up the other complaints as well. Not sure when that will be...

theshadow27:
Builds are failing due to:
 - You must update Jenkins from v2.60.3 to v2.73.3 or later to run this plugin.

basil:
Trying to make Cloud Statistics optional made the code very complicated. I checked other Jenkins cloud plugins that supported Cloud Statistics and it was not optional for any of them. Therefore, I am closing this PR in favor of #797.

PR #826, opened 2021-01-28
Replace volumes with mounts, allowing to use tmpfs type

### LINKED ISSUE #714: Enh: ability to specify tmpdir when starting container
First off, very nice plugin!   It seems to support most of the Docker run functionality, and seems very stable.

One feature which would be very useful is to be able to specify the --tmpfs (or --mount) options (0..n times) for a particular path.  

A tmpfs option to docker run is needed in order to run a container w/systemd in non-privileged mode, as described at 
 https://developers.redhat.com/blog/2016/09/13/running-systemd-in-a-non-privileged-container/
which suggests starting a container w/systemd like this:
   docker run -d --tmpfs /tmp --tmpfs /run -v /sys/fs/cgroup:/sys/fs/cgroup:ro httpd

and which according to this page: https://docs.docker.com/storage/tmpfs/
I believe is equivalent to this (--mount allows for additional tmpfs config options such as size and mode):
   docker run -d --mount type=tmpfs,destination=/tmp  --mount type=tmpfs,destination=/run  -v /sys/fs/cgroup:/sys/fs/cgroup:ro httpd

Today, it does not seem possible to specify this when starting up the build agent containers.   Adding either the --tmpfs or --mount options would allow this to be specified.

This would allow systemd-based containers to run in non-privileged mode, which would be very useful in a Jenkins-based testing environment.

A workaround (the only one I know of right now) for this issue is to run the docker template/container in privileged mode.

bguerin:
Fix #714


bguerin:
Hello @pjdarton, do you think you will have some time soon to look at it ? Thanks !

bguerin:
@pjdarton rebased on master, successfully tested on my installation

pjdarton:
After a few (unrelated) hiccups (https://issues.jenkins.io/browse/JENKINS-65142), I was able to test this on my local setup too, in that I've got old-style volume mounts specified and they kept working after I'd installed this.

I'll merge it now.

Issue #825, opened 2021-01-27
"bad_certificate" on TLS connection to remote dockerd


Osipion:
## Summary:

I am unable to connect to a remote dockerd instance due to TLS issues.

## Versions & Error Messages:

* docker-plugin version: 1.2.2
* jenkins version: 2.263.3
* docker engine version: 20.10.2
* java version: OpenJDK 1.11.0 (Java 11)
* OS: ubuntu-sever-20.04
* issue occurs during cloud setup, before any containers are in use
* stack trace:
```
javax.net.ssl.SSLHandshakeException: Received fatal alert: bad_certificate

javax.net.ssl.SSLHandshakeException: Received fatal alert: bad_certificate
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:117)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:337)
	at java.base/sun.security.ssl.Alert$AlertConsumer.consume(Alert.java:293)
	at java.base/sun.security.ssl.TransportContext.dispatch(TransportContext.java:186)
	at java.base/sun.security.ssl.SSLTransport.decode(SSLTransport.java:171)
	at java.base/sun.security.ssl.SSLEngineImpl.decode(SSLEngineImpl.java:681)
	at java.base/sun.security.ssl.SSLEngineImpl.readRecord(SSLEngineImpl.java:636)
	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:454)
	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:433)
	at java.base/javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:637)
	at io.netty.handler.ssl.SslHandler$SslEngineType$3.unwrap(SslHandler.java:294)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1297)
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1199)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1243)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:502)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:441)
[...]

PR #829, opened 2021-02-04
upgrade to docker-java-api 3.2.2

Osipion:
please see #825

notes:

* relies on docker-java-api being updated to use docker-java 3.2.2 (e.g. DO NOT MERGE :-) )
* requires docker-java-api remove the maven exclusion on bouncycastle
* Amazon ECR Credentials plugin no longer works with this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,20 +55,20 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.level>8</java.level>
         <groovy.version>2.4.7</groovy.version>
-        <jenkins.version>2.73.3</jenkins.version>
+        <jenkins.version>2.235.1</jenkins.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>bouncycastle-api</artifactId>
-            <version>2.16.2</version>
+            <version>2.18</version>
         </dependency>
 
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>docker-java-api</artifactId>
-            <version>3.1.5</version>
+            <version>3.2.2</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.apache.httpcomponents</groupId>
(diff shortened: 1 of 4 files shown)


pjdarton:
I think we need to break this up a bit and fix the nasty "the docker-plugin needs to immediately release a new version every time the docker-java-api-plugin changes the docker-java version" issue so we have a wee bit more freedom to upgrade the docker-java-api-plugin without it killing this plugin.
...so I'm working on a PR, https://github.com/jenkinsci/docker-java-api-plugin/pull/9 , which moves the most fragile code in the docker-plugin (fragile because it needs to perfectly match the docker-java code) into the docker-java-api-plugin (where it can be updated each time docker-java is updated).

If/when that's merged, I'll do a "now make use of it" PR to docker-plugin (meaning the the DelegatingDockerClient disappears from this plugin, but then changes to it have to be a part of any "change the version of docker-java" PR on docker-java-api-plugin) and then we merge the current state of play into this PR (which would be a bit smaller as DelegatingDockerClient will've gone from here).

I remain "concerned/unhappy" about the dependencies fallout from this PR, e.g. the Amazon ECR stopping to work sounds like a nasty unwanted side effect; sadly, maven dependencies aren't something I fully understand so it's not something I can help with on a technical front (I'm kinda floundering around in the dark the moment it gets non-trivial, and any "excludes" are non-trivial IMO).

Osipion:
Ok, this seems like a good approach, thanks @pjdarton . As a very temporary measure, I ended up forking Yet Another Docker and patching the changes into that. The benefit YAD has is that it uses the Maven shader plugin on docker-java to directly include docker-java in the project, removing the dependency on docker-java-api. I basically followed a simillar approach to enable ECR authentication - directly copying the relevant code from [amazon-ecr-plugin](https://github.com/jenkinsci/amazon-ecr-plugin) into the project, and refactoring the custom registry credentials YAD uses to make them compatible with the ECR format. In this way I was able to get something that seemed to work fine and minimized the dependencies on the rest of the ecosystem. Definitely not saying that this is the right way to do things though

pjdarton:
FYI I have (#836) refactored the pom.xml file so it now uses the Jenkins "BOM" plugin to help with dependency management, bumped the minimum Jenkins version forward, and resolved many of the fixups you needed to address in this PR "all at once".
So, _hopefully_ if you update the pom.xml to be based off the current (as-yet-unreleased) master then you should see the need for a whole load of changes disappear.
Furthermore, if my [proposed change](https://github.com/jenkinsci/docker-java-api-plugin/pull/9) to the docker-java-api-plugin go in (which would then result in a new version 3.1.5.3) then that'd allow us to remove the DelegatingDockerClient code from this plugin and thus remove the need to make the changes you needed to do within this PR too.
i.e. that'd make this PR all about changing versions and not having to address old tech-debt at the same time 😉

pjdarton:
This was resolved by #882 and #884 which allowed this plugin to use later versions of the docker-java-api-plugin.

PR #831, opened 2021-02-10
Cpusetcpus

andrei-barna:
Adding new option --cpuset-cpus


andrei-barna:
Hi everyone. I am not a Java developer, however it didn't seem too difficult to add it.

Can someone confirm please if it's correct, and if I should expand on the help? Thank you.

andrei-barna:
Hi, I am sure I'm not the only one who needs this addition, so anyone with more knowledge than I have, can validate my changes? Thank you.

pjdarton:
I don't know if you're the only person who wanted this, but I can tell you that you're the first person to provide this enhancement.
...and, despite your claim that you're "not a Java developer", you've done pretty well 👍 

There's a couple of things I'd like to see:

1) Validation of the String contents when the user is editing the configuration.  That requires the addition of a new "doCheck..." method in the Descriptor class that's inside the class where the new field was added.
In this case, that'll be a new `doCheckCpusetCpus` method within DockerTemplateBase.java's `DescriptorImpl" sub-class.
FYI Jenkins calls the "doCheck..." methods when the WebUI's focus moves out of the field, and this allows the code to tell the user e.g. "No, typing in the string 'flibble' isn't a valid CPUSet specification"
This isn't _always_ feasible/reasonable, but if the syntax of this field is always numbers, commas & minus-signs then some form of validation ought to be feasible.
It's always a good idea to provide as much validation as possible in the configuration WebUI section because it's difficult for users to see the errors resulting from bad configuration (it only appears in the Jenkins logs & the "enabled" flag in the config WebUI, but not all users look there)

2) A unit-test that proves it works.
That's easist done by adding a new method (or new methods) to DockerTemplateBaseTest.java
It's always a good idea to add unit-tests for new functionality because that way it means that any subsequent change that breaks this functionality will get detected by the automated tests.
If we don't have a unit test then nobody else would notice if a subsequent PR breaks your enhancement, so you run the risk that one day you update to the latest plugin, find it's not working anymore, and then have to figure out why.

andrei-barna:
Thank you very much @pjdarton for your feedback, it's very valuable and I appreciate it. :)
Unfortunately I know very well, what you mean, about the not so obvious errors (either visibility or location), you're absolutely right. In the next few days I will try to implement your suggestions, and probably will post again a comment. :) Thank you again.

pjdarton:
As there's been no movement on this for over a year, I'm going to close it.
If you still want this, feel free to re-open this (or a new PR), resolve the conflicts and requested changes and we can take another look at merging it.

PR #843, opened 2021-07-28
Fix Jenkins 66025

realitix:
This pull request switch Netty implementation by HTTP implementation.

CI tests should be passed on your CI with your configuration to validate the good behavior.

https://issues.jenkins.io/browse/JENKINS-66025

Thanks

### DIFF
--- a/src/main/java/io/jenkins/docker/client/DockerAPI.java
+++ b/src/main/java/io/jenkins/docker/client/DockerAPI.java
@@ -1,22 +1,21 @@
 package io.jenkins.docker.client;
 
-import com.cloudbees.plugins.credentials.domains.DomainRequirement;
-import com.github.dockerjava.api.DockerClient;
-import com.github.dockerjava.api.command.VersionCmd;
-import com.github.dockerjava.api.model.Version;
-import com.github.dockerjava.core.DefaultDockerClientConfig;
-import com.github.dockerjava.core.DockerClientBuilder;
-import com.github.dockerjava.core.SSLConfig;
-import com.github.dockerjava.netty.NettyDockerCmdExecFactory;
+import static com.cloudbees.plugins.credentials.CredentialsMatchers.firstOrNull;
+import static com.cloudbees.plugins.credentials.CredentialsMatchers.withId;
+import static com.cloudbees.plugins.credentials.CredentialsProvider.lookupCredentials;
+import static com.nirima.jenkins.plugins.docker.utils.JenkinsUtils.bldToString;
+import static com.nirima.jenkins.plugins.docker.utils.JenkinsUtils.endToString;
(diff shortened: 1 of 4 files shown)


realitix:
Hello, could you look at this PR ?

Thanks

pjdarton:
It has merge issues that need fixing.

src/main/java/io/jenkins/docker/client/DockerAPI.java
```diff
@@ -236,34 +240,25 @@ public void reallyClose() throws IOException {
...
-                    .withConnectTimeout(connectTimeoutInMillisecondsOrNull);
-            final DefaultDockerClientConfig.Builder configBuilder = new DefaultDockerClientConfig.Builder()
-                    .withDockerHost(dockerUri)
-                    .withCustomSslConfig(toSSlConfig(credentialsId));
-            actualClient = DockerClientBuilder.getInstance(configBuilder)
-                .withDockerCmdExecFactory(cmdExecFactory)
-                .build();
+            DockerHttpClient httpClient = new ApacheDockerHttpClient.Builder().dockerHost(URI.create(dockerUri))
+                    .sslConfig(toSSlConfig(credentialsId)).build();
+            actualClient = DockerClientBuilder.getInstance().withDockerHttpClient(httpClient).build();
```
pjdarton, line 253: Please retail the same "code style" as the previous code.

FYI the reason why it was putting each bit on its own line is so that, in future, if we need to add/remove/change any calls to the builder, all we're doing is adding/removing/changing one line - all the other lines will remain unchanged, and this makes viewing the history in git (e.g. using "git blame") _so_ much easier as you can see where each builder parameter came from.
If we allow multiple calls per line, it means that to add an extra "with..." the code change will show that we've rewritten the entire builder clause rather than just added one extra thing.

TL;DR: Multi-line builder code makes for more maintainable code.

pjdarton (changes requested):
The main codebase has moved on since this PR was first done, and it needs to be rebased on the current latest code before it'll make sense again.

Also, there's a _lot_ of unnecessary changes to files in here.
This PR is a change to the plugin's functionality, so we want cosmetic/layout changes kept to an absolute minimum so that this PR's changes consist of _only_ the functionality changes ... and does so in a manner that "fits in" with the existing code style (however much you might dislike it 😉).
If you find the code intolerably ugly and need to make cosmetic changes to the code's layout etc to make it "prettier", do that in a separate PR that _only_ makes cosmetic code changes.

pjdarton:
@mat1e Has this PR also been superseded by #899 ?

mat1e:
> been

Yes.

mat1e:
> @mat1e Has this PR also been superseded by #899 ?

Yes.

PR #846, opened 2021-08-08
JENKINS-66320 Forwards compatible guava

timja:
See https://issues.jenkins.io/browse/JENKINS-66320

cc @pjdarton


basil:
Thanks, @pjdarton! Would it be possible for this to be released? This would be of assistance to the Jenkins core effort to upgrade Guava.

pjdarton:
Released an hour ago.

PR #848, opened 2021-08-22
Migrate from `com.google.common.base.Objects` to `java.util.Objects`

basil:


pjdarton:
While I can see that the commit succinctly describes what this PR is doing, I am not sure why it is necessary (or even desireable).

What's the rationale/justification for this change?

basil:
> What's the rationale/justification for this change?

See [the Guava documentation](https://guava.dev/releases/snapshot/api/docs/com/google/common/base/Objects.html):

> **Note for Java 7 and later:** This method should be treated as deprecated; use `Objects.[equals|hash]` instead.

PR #855, opened 2021-09-11
Stop using deprecated `Util#join`

basil:
This method was deprecated in jenkinsci/jenkins#5467, so migrate away from it in favor of native Java Platform functionality.

### DIFF
--- a/src/main/java/io/jenkins/docker/pipeline/DockerNodeStepExecution.java
+++ b/src/main/java/io/jenkins/docker/pipeline/DockerNodeStepExecution.java
@@ -5,7 +5,6 @@ import com.nirima.jenkins.plugins.docker.DockerTemplate;
 import com.nirima.jenkins.plugins.docker.DockerTemplateBase;
 import hudson.EnvVars;
 import hudson.FilePath;
-import hudson.Util;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.model.TaskListener;
@@ -31,6 +30,7 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 /**
  * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/docker/pipeline/DockerNodeStepExecution.java
```diff
@@ -193,7 +193,7 @@ private void invokeBody(DockerTransientNode node, TaskListener listener) {
             env.overrideExpandingAll(computer.buildEnvironment(listener));
             env.put("NODE_NAME", computer.getName());
             env.put("EXECUTOR_NUMBER", "0");
-            env.put("NODE_LABELS", Util.join(node.getAssignedLabels(), " "));
+            env.put("NODE_LABELS", node.getAssignedLabels().stream().map(Object::toString).collect(Collectors.joining(" ")));
```
pjdarton, line 196: While I'm unsure of the cause, it looks like this change breaks the unit-tests.

...also, I'm not a big fan of having so many method calls daisy-chained on the same line - it makes debugging a NPE nigh-on impossible.
I'd suggest implementing a `join` method (possibly in this plugin's utils package, maybe as a new class unless it fits well in another), and adding a unit-test for it (to prove it'll cope with a null input, or an input containing nulls, as well as normal non-null stuff), and then call that instead of `Util.join`.
I _suspect_ that such a unit-test would then reveal that `Util.join` was more null-safe than this new code and fixing that would then fix the failing unit-tests.

pjdarton (changes requested):
Can't merge as-is as it seems to break the unit-tests.

pjdarton:
Thank you for bringing that to my attention.
Unfortunately, this is old code (long since abandoned by its original authors) and (as you've demonstrated) it seems to be somewhat fragile. It's not an ideal situation.
I would, of course, welcome a PR that makes it work again (or even educated guesses as to where the problem(s) are as I'd be temped to have a go myself) but, until someone cares enough to fix it (usually when someone wants some non-trivial enhancement merged), it'll remain broken.

pjdarton:
FYI I found one issue which I fixed in #858 (the official docker image had changed, which broke the unit tests that used it).
The unit tests now pass ... most of the time.
Sometimes they fail because they weren't able to download a Windows JRE, sometimes because of a ClassNotFound exception, sometimes because a cloud resource ran out of memory etc - the reasons seem to vary from one run to the next but, if one spams the "re-run" button, it's possible to get a clear run. I don't have access to the jenkinsci server internals so I can't investigate why this happens (it works when I run things locally and the jenkinsci service has never been 100% reliable IME)

@basil I've (re)implemented your suggested changes in #862 (which is now merged), so you can cross this plugin off the TODO list in  jenkinsci/jenkins#5467

Issue #884, opened 2022-04-19
Docker plugin Version 1.2.8  with Docker API Plugin Version 3.2.13-35.vb82bff5c8bd0 


dnwe:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.332.1
OS: Linux - 4.15.0-173-generic
---
ace-editor:1.1
analysis-model-api:10.10.0
ansicolor:1.0.1
ant:1.13
antisamy-markup-formatter:2.7
apache-httpcomponents-client-4-api:4.5.13-1.0
artifactory:3.16.1
authentication-tokens:1.4
authorize-project:1.4.0
badge:1.9.1
bootstrap4-api:4.6.0-3
bootstrap5-api:5.1.3-6
bouncycastle-api:2.25
branch-api:2.1044.v2c007e51b_87f
build-keeper-plugin:1.3
build-user-vars-plugin:1.8
build-with-parameters:1.6
caffeine-api:2.9.2-29.v717aac953ff3
checks-api:1.7.2
cloudbees-disk-usage-simple:0.10
cloudbees-folder:6.714.v79e858ef76a_2
collapsing-console-sections:1.8.0
command-launcher:1.6
conditional-buildstep:1.4.2
config-file-provider:3.9.0
configuration-as-code:1414.v878271fc496f
copyartifact:1.46.3
credentials:1087.1089.v2f1b_9a_b_040e4
credentials-binding:1.27.1
cvs:2.19.1
dark-theme:156.v6cf16af6f9ef
data-tables-api:1.11.4-4
display-url-api:2.3.6
docker-commons:1.19
docker-java-api:3.2.13-35.vb82bff5c8bd0
docker-plugin:1.2.8
docker-workflow:1.28
durable-task:495.v29cd95ec10f2
echarts-api:5.3.2-1
email-ext:2.87
embeddable-build-status:2.0.3
envinject:2.854.vfa_1657078c97
envinject-api:1.199.v3ce31253ed13
extended-read-permission:3.2
external-monitor-job:191.v363d0d1efdf8
font-awesome-api:6.0.0-1
forensics-api:1.12.0
ghprb:1.42.2
git:4.11.0
git-client:3.11.0
git-parameter:0.9.16
git-server:1.10
github:1.34.3
github-api:1.301-378.v9807bd746da5
github-branch-source:1598.v91207e9f9b_4a_
google-oauth-plugin:1.0.6
[...]

PR #885, opened 2022-04-19
Cope with docker-java API change.

pjdarton:
Attempt to resolve issue #884

Looks like the docker-java code moved a method into a parent class and changed its API while doing so.
Switching to use the parent class ought to be version agnostic...


pjdarton:
It compiled & tests pass locally; seems unreliable on JenkinsCI ... but that's nothing new 😞

PR #901, opened 2022-06-15
Improve unit tests on Windows

Peter-Darton-i2:
Running the unit-tests on Windows has always been problematic.
 - Timers behave differently
 - socket tests take a non-trivial amount of time and throw timing-based unit tests off
 - ... and we still assume that there's no docker daemon on Windows.

This improves the docker-connector unit tests, fixes a bug in PortUtilsTest that made it unreliable on Windows, and gets us closer to enabling docker-dependent tests on Windows (but doesn't because the Windows VMs on the Jenkins CI still don't include docker)

### DIFF
--- a/src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
+++ b/src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
@@ -13,6 +13,8 @@ import hudson.model.Result;
 import hudson.model.TaskListener;
 import hudson.model.queue.QueueTaskFuture;
 import hudson.slaves.Cloud;
+import hudson.tasks.BatchFile;
+import hudson.tasks.Builder;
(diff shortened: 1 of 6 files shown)


src/test/java/io/jenkins/docker/connector/DockerComputerConnectorTest.java
```diff
@@ -159,23 +165,41 @@ protected void should_connect_agent(DockerTemplate template) throws IOException,
         // FIXME on CI windows nodes don't have Docker4Windows
         Assume.assumeFalse(SystemUtils.IS_OS_WINDOWS);
```
Peter-Darton-i2, line 166: I tried removing this line but the tests failed saying "connection refused" to docker on localhost port 2375.
That implies that the Windows VMs we test with don't include docker.

PR #907, opened 2022-07-29
Gitpodify repository

jmMeessen:
PR to add Gitpod configuration. Accommodates Java 17 requirement for VSC Language server

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,8 @@
+FROM gitpod/workspace-full
+
+USER gitpod
+
+RUN bash -c ". /home/gitpod/.sdkman/bin/sdkman-init.sh && \
+    sdk install java 17.0.3-tem && \
+    sdk install java 11.0.15-tem  && \
+    sdk default java 17.0.3-tem"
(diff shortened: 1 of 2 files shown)


Peter-Darton-i2:
I know nothing of gitpod ... but presumably this will result in builds (of this git repo) being made.
Are they going to appear anywhere visible to the general public, or is this purely for private use?

MarkEWaite:
@Peter-Darton-i2 thanks for asking.  [Gitpod](https://www.gitpod.io/) allows a user to run a development environment with Visual Studio Code (or IntelliJ IDEA) from inside their web browser.  It is an easier way to configure a development environment for new contributors.

We saw on the [plugins site](https://plugins.jenkins.io/docker-plugin/) that this plugin is listed as "up for adoption", so we started the process to prepare to use it as one of several plugins in the "Contributing to Open Source" workshop that we'll be doing at DevOps World 2022 in September 2022.  

However, since you responded promptly to the pull request, that may indicate that the plugin really is not up for adoption.  Do you consider the plugin still up for adoption?

Peter-Darton-i2:
I'm only here because I'm the "last one standing" - I don't use this plugin anymore and so while I do still have maintainer access, I don't have the time/motivation to push this forwards anymore.
So, yes, it is "up for adoption" but I'll do what I can to keep things ticking along until a new owner is found ... and if adding gitpod files will help get a new maintainer on board then I'm all in favour.

MarkEWaite:
Thanks so much for the clarification @Peter-Darton-i2 and for your willingness to help even when you're not using the plugin

PR #911, opened 2022-08-29
Update Docker logo to be a Jenkins symbol

janfaracik:
Small PR to update the Docker logo to be a [Jenkins Symbol](https://weekly.ci.jenkins.io/design-library/Symbols/) rather than a `.png`. This brings the advantages that the icon now scales and can change color depending on the user's theme.

**New icon**
image

I've updated the min Jenkins version to support Symbols + updated an if condition which was now failing due to an unnecessary null check.

If there's any issues let me know 👍 

---


pom.xml
```diff
@@ -53,7 +53,7 @@
         <changelist>-SNAPSHOT</changelist>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <groovy.version>2.4.7</groovy.version>
-        <jenkins.version>2.303.3</jenkins.version>
+        <jenkins.version>2.346</jenkins.version>
```
NotMyFault, line 56:
```suggestion
        <jenkins.version>2.346.1</jenkins.version>
```

NotMyFault (commented):
The docker logo is shipped by the ionicons api plugin too, we don't necessarily need to ship the symbol with the docker plugin.

janfaracik:
> The docker logo is shipped by the ionicons api plugin too, we don't necessarily need to ship the symbol with the docker plugin.

Happy to change to that if you want 👍

NotMyFault:
> > The docker logo is shipped by the ionicons api plugin too, we don't necessarily need to ship the symbol with the docker plugin.
> 
> Happy to change to that if you want 👍

I don't mind either way, I'm not a maintainer of this plugin, but I thought of using the library to remove the need to ship a possible obsolete icon, given you may already have a plugin, which depends on the ionicons api plugin.

basil:
@janfaracik I am sorry that the previous maintainer left you hanging on this PR without so much of an acknowledgment. On behalf of the project, I want to thank you for your contribution. I love this kind of thoroughness to adapt all parts of the ecosystem for a consistent user experience.

basil:
> updated an if condition which was now failing due to an unnecessary null check.

I couldn't convince myself through code inspection that the null check was unnecessary in all cases, and it does not seem related to the core purpose of this PR, so I have reverted this hunk. If this is indeed necessary for some reason, please file a separate PR where this change can be tested separately.

basil (changes requested):
Causes a regression in the Built on Docker page for Freestyle projects. Before, the icon is a normal size. After, the icon is too huge. Screenshots attached.

docker-logo-before

docker-logo-after

basil:
CC @jenkinsci/sig-ux

timja:
> CC @jenkinsci/sig-ux

Steps to reproduce? I spent some time and I don't know how to get that page setup, I was able to get to the `DockerBuildImageAction` but not the `DockerBuildAction`. I've never used this plugin before

The patch is likely to be:

```diff
diff --git a/src/main/resources/com/nirima/jenkins/plugins/docker/action/DockerBuildAction/index.jelly b/src/main/resources/com/nirima/jenkins/plugins/docker/action/DockerBuildAction/index.jelly
index 518285e..ee1dc7f 100644
--- a/src/main/resources/com/nirima/jenkins/plugins/docker/action/DockerBuildAction/index.jelly
+++ b/src/main/resources/com/nirima/jenkins/plugins/docker/action/DockerBuildAction/index.jelly
@@ -3,7 +3,22 @@
     <l:layout title="Docker">
     <l:main-panel>

-       <st:include page="summary.jelly" />
+        <h1>Docker Build Data</h1>
+
+        <ul>
+            <j:choose>
+                <j:when test="${it.cloudId != null}">
+                    <li><b>Cloud:</b> ${it.cloudId}</li>
+                </j:when>
+                <j:otherwise>
+                    <li><b>Host:</b> ${it.containerHost}</li>
+                </j:otherwise>
+            </j:choose>
+            <li><b>Container Id:</b> ${it.containerId}</li>
+            <j:if test="${it.taggedId != null}">
+                <li><b>Committed Container Id:</b> ${it.taggedId}</li>
+            </j:if>
+        </ul>

       <j:if test="${it.inspect != null}">
          <h2>Container details (docker inspect)</h2>
```

basil:
That patch seems to merely work around the problem by removing the icon rather than solving the problem by rendering the icon at the correct size. To reproduce the problem, you can create a Docker cloud with an agent template whose name is `docker-agent`, whose image is `jenkins/agent:latest`, and whose connect method is **Attach Docker Container**; followed by creating a Freestyle project that runs on agents labeled `docker-agent`. Building the Freestyle project should run the build in a Docker agent, after which there should be a **Built on Docker** link in the sidebar when viewing the build. Clicking on that link takes you to the page with the huge icon.

timja:
> That patch seems to merely work around the problem by removing the icon rather than solving the problem by rendering the icon at the correct size

It removes an icon that shouldn't be there:

https://weekly.ci.jenkins.io/design-library/Symbols/

> Don't add symbols to headings, they just create visual clutter

timja:
I had tried that I debugged through the code and found the Cloud was disabled.
After clicking 'Docker cloud details' -> 'Advanced" I saw an enabled flag which is off by default =/.

image

After enabling that it worked

----

Tested and pushed:
image

timja (approved):

PR #919, opened 2022-10-21
Unit-test robustness

pjdarton:
The unit-tests are a bit flakey in places, leading to false-negative builds on the official Jenkins CI.
This PR makes some of the code less flakey.

 - `DockerMultiplexedInputStreamTest` wasn't waiting for threads to finish before testing results
 - `DockerComputerConnectorTest` sometimes takes more than 90 seconds if docker is slow
 - The code sometimes called `Jenkins.get().addNode(...)` directly instead of from within a retry loop

Only the latter point affects non-test behaviour, which _might_ have contributed to unreliability on agents not using SSH (i.e. JNLP & "direct") ... although that'd be hard to prove because, in "real world" usage, Jenkins would retry and a failure should merely cause a provisioning delay rather than a hard failure.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -379,7 +379,7 @@ public class DockerCloud extends Cloud {
                             plannedNode.complete(agent);
 
                             // On provisioning completion, let's trigger NodeProvisioner
-                            robustlyAddNodeToJenkins(agent);
+                            agent.robustlyAddToJenkins();
 
                         } catch (Exception ex) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
```diff
@@ -421,44 +421,6 @@ public void run() {
...
-                final long delayInMilliseconds = 10L * attempt;
-                try {
-                    Thread.sleep(delayInMilliseconds);
-                } catch (InterruptedException e) {
-                    throw new IOException(e);
-                }
-            }
-        }
-    }
-
```
pjdarton, line 461: This method moved to DockerTransientNode

PR #926, opened 2022-11-16
Suppressing common info-level messages from `DockerContainerWatchdog`

jglick:
These get printed every few seconds, even if you are not actively using Docker agents.

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerContainerWatchdog.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerContainerWatchdog.java
@@ -121,7 +121,7 @@ public class DockerContainerWatchdog extends AsyncPeriodicWork {
             return;
         }
 
-        LOGGER.info("Docker Container Watchdog has been triggered");
+        LOGGER.debug("Docker Container Watchdog has been triggered");
 
         executionStatistics.writeStatisticsToLog();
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/nirima/jenkins/plugins/docker/DockerContainerWatchdog.java
```diff
@@ -121,7 +121,7 @@ protected void execute(TaskListener listener) throws IOException, InterruptedExc
             return;
         }
 
-        LOGGER.info("Docker Container Watchdog has been triggered");
+        LOGGER.debug("Docker Container Watchdog has been triggered");
```
jglick, line 124: For some reason this plugin uses SLF4J, unlike the rest of the Jenkins ecosystem which uses JUL.

MarkEWaite (approved):

Peter-Darton-i2:
These should be every 5 _minutes_ unless something's gone tragically wrong.

DockerContainerWatchdog.java defaults [RECURRENCE_PERIOD_IN_MS](https://github.com/jenkinsci/docker-plugin/pull/926/files/d2b044d7e44ab02c182b20dd44ca8990eb28f911#diff-cdeb363c70a425b216a3e0f803017ddb0733d7cf645c1df21255c4f70393b4adR64) to `5*60*1000` which is returned by [getRecurrencePeriod()](https://github.com/jenkinsci/docker-plugin/pull/926/files/d2b044d7e44ab02c182b20dd44ca8990eb28f911#diff-cdeb363c70a425b216a3e0f803017ddb0733d7cf645c1df21255c4f70393b4adR76) which should result in a callback every 5 minutes.

If this is actually triggering "every few seconds" then there's more serious problems than just a message at the wrong severity level.

jglick:
Indeed looks like every five minutes, but the point is that it is noise printed even when the controller is completely idle.

PR #942, opened 2023-04-18
Docker plugin does not work correctly with agent Images built using BuildKit

### LINKED ISSUE #867: Docker plugin does not work correctly with Agent Images built using buildkit
I believe the docker-plugin uses information from the docker agent image specifically
```
"ContainerConfig" {
  "WorkingDir": "/home/jenkins"
}
```
to work out where to install the remoting jar and where to log to when using the "Attach Docker container" connection method.
When agent images are built using buildkit this information is not present in ContainerConfig. It should be retrieved from Config instead.
I think this error is in getEffectiveRemoteFs() method of 
```
src/main/java/com/nirima/jenkins/plugins/docker/DockerTemplate.java
```
The result is that the remoting can not start due to permission issues on the log files.

nilsmahlstaedt:
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue - manual integration test

## What did I do:
Incorporated xabolcs' suggested change (#867) to enable usage of docker based jenkins agent's built with buildkit (the new default docker build backend on linux, mac and win).

Change was nexessary, as buildkit no longer fills the `.containerConfig` section of a docker inspect response. It uses the `.config` path. Both are backed by an object of the same layout so no code changes other than accessing config instead of containerConfig where necessary.

see also docker/cli#2868

Change was successfully was tested on our company infrastructure. There were no automatic unit tests added.


MarkEWaite (approved):

basil (approved):
Thanks for the PR!

PR #957, opened 2023-05-23
Fix some issues with `dockerNode` mixed with `node`

jglick:
Closes #948, adapting to https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180. I would not be surprised if there are some more corner cases (like https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/292) affecting something you should never do—hold onto two (or more) executors at once—which generally results from misunderstanding, especially in a Declarative Pipeline which may obscure the control flow. Also doing the equivalent of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/128 since I noticed that had been forgotten here. See also #557.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,17 +81,6 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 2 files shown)


basil (approved):
Thanks for the PR!

PR #967, opened 2023-05-30
Add sidebar to Docker build action

basil:
It seemed really weird to me that this page wouldn't have a sidebar like everything else. I looked at other run actions and they generally seemed to have a sidebar. To test this I loaded the **Docker Build Data** page and verified it had a sidebar.

CC @jenkinsci/sig-ux


basil:
Thanks, I fixed that pre-existing problem in commit 5afb80d735f5b60ed367b8fd378d14812e06aac4

PR #971, opened 2023-06-01
Improving uniqueness of templateID string (via template name)

### LINKED ISSUE #816: Wrong calculation of running Template agents when using the same image on multiple templates
- [X] docker-plugin version : **1.2.1**
 - [X] jenkins version : **2.249.2**
 - [X] docker engine version : **19.03.13**

In [`com.nirima.jenkins.plugins.docker.DockerCloud` in `canAddProvisionedSlave()`](https://github.com/jenkinsci/docker-plugin/blob/c3a38d7034132cb407832a4ae0372f44339a2b03/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java#L650), an estimation of running template agents is done.

This is done by counting the running containers using the same image as the template image, which is wrong since multiple templates of the same cloud can use the same Docker image. In my case, I have 2 templates with the same Docker image, but different container capacities and different agent labels (for limiting the builds).

It may be fixed by adding a template label to each container and add a filter on this label to count running template agents?

cpfeiffer:
Just as in #816 and #874, we want to distinguish templates that use the same docker image.

We do this pragmatically by using the user-supplied template name, so that the user is in control to make it unique.

So this fixes #816 as long as the user provides distinguished template names, which is explained by the help texts.

### Testing done

- unit testcases were extended and pass
- code has been used successfully in production for 10 months

### Submitter checklist

- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
+++ b/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java
@@ -16,6 +16,7 @@ import com.github.dockerjava.api.command.PullImageCmd;
 import com.github.dockerjava.api.command.PushImageCmd;
 import com.github.dockerjava.api.command.StartContainerCmd;
 import com.github.dockerjava.api.model.AuthConfig;
+import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.api.model.Version;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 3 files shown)


cpfeiffer:
Fair enough. We chose to be defensive here. So far, the template name has been completely optional and defaulted to just "docker". We didn't change that behavior, but gave the user an option to fix this issue. Suddenly asking this to be unique might upset a few of the 38k users of this plugin. I wouldn't mind, though.

basil:
To avoid bothering people who do not need to be bothered, we could only issue a warning when (a) there is more than one template and (b) two templates resolve to the same effective value (e.g., two blank templates, one blank template and one template named "docker", two templates named "docker", two templates named "my-template", etc). The code depends on this for correctness, and you are warning people about it in the help, but I think the warning should be more prominent than that to avoid supportability challenges. I do not think this is a big ask since you are already pushing for this in the help text.

basil:
FYI, I have removed myself from the list of maintainers for this plugin, so even if the requested changes are made and I approve this PR, I would still not have permissions to do a release. If this PR is important to you, I would recommend adopting this plugin.

PR #1016, opened 2023-09-12
Require Jenkins 2.414.3 or newer

car-roll:
## Require Jenkins 2.414.3 or newer

Also removes tests that relied on an empty cloud name being allowed.

was originally downstream of https://github.com/jenkinsci/jenkins/pull/8310 but that has now been abandoned

since Jenkins 2.403, it is no longer possible to create a cloud with an empty name. This fixes tests that create empty clouds.

### Testing done

N/A - changes only in test automation are well covered by ci.jenkins.io

```[tasklist]
### Submitter checklist
```


src/test/java/com/nirima/jenkins/plugins/docker/DockerCloudTest.java
```diff
@@ -58,34 +58,14 @@ public void testConstructor_0_10_2() {
 
     private static final String LOG_MESSAGE = "Docker cloud requires a non-blank name after Jenkins 2.402";
 
-    @Issue("JENKINS-70729") // Warn if cloud name is empty
```
car-roll, line 61: @MarkEWaite I removed these two tests because the base class (`Cloud`) now throws an `IllegalArgumentException` that does not allow the creation of `DockerCloud`.

src/test/java/com/nirima/jenkins/plugins/docker/DockerCloudTest.java
```diff
@@ -58,34 +58,14 @@ public void testConstructor_0_10_2() {
...
-        MatcherAssert.assertThat(lr.getMessages(), IsIterableContaining.hasItem(LOG_MESSAGE));
-    }
-
     @Issue("JENKINS-70729") // Handle null or empty cloud name
     @Test
     public void testCopyConstructor() {
         lr.record(DockerCloud.class.getName(), Level.ALL).capture(16);
         DockerCloud cloud =
-                new DockerCloud(null, new DockerAPI(new DockerServerEndpoint("uri", "credentialsId")), List.of());
+                new DockerCloud("tmp", new DockerAPI(new DockerServerEndpoint("uri", "credentialsId")), List.of());
```
car-roll, line 66: @MarkEWaite I kept the CopyConstructor test since it covers backwards compatibility, but I had to modify it to allow an "empty" named DockerCloud

src/test/java/com/nirima/jenkins/plugins/docker/DockerCloudTest.java
MarkEWaite: Thanks @car-roll !  Has the base class been extended to handle upgrades on systems that already have a Docker cloud named with an empty string?  If not, that seems like it will be needed in order to not break upgrades.

src/test/java/com/nirima/jenkins/plugins/docker/DockerCloudTest.java
MarkEWaite: Thanks @car-roll .  That makes sense to me.  Assigning directly to the `name` attribute is a nice technique.  I would not have thought of that.

src/test/java/com/nirima/jenkins/plugins/docker/DockerCloudTest.java
car-roll: So the base class has not been extended to support empty string Cloud names, but support is still there as I did not touch the `readResolve` in `DockerCloud`. 

Just to clarify what I meant, I believe the code that these two tests are testing (`testConstructorWithEmptyName` and `testConstructorWithNullName`) are testing the constructor itself
https://github.com/jenkinsci/docker-plugin/blob/ccec39ca5e6f72ff98e7d1d33367dd104603df4c/src/main/java/com/nirima/jenkins/plugins/docker/DockerCloud.java#L136
However, with this new jenkins version, we are now failing on the `super(name)` call, so that log message will now never be called. Test-wise, I do not think I can reproduce this scenario with the new Jenkins version.

The only time now we will have to deal with a null name constructor should be in existing cloud names, which is tested in `testCopyConstructor`.

car-roll:
Note: https://github.com/jenkinsci/jenkins/pull/8310 has now been closed, but i think the test changes for this PR will be needed whenever the jenkins version is bumped past 2.403. This is when cloud name is not allowed to be empty
https://github.com/jenkinsci/jenkins/blob/5665665b6b7bf45cc3102338477540029a4d8f62/core/src/main/java/hudson/slaves/Cloud.java#L121

MarkEWaite (approved):
Thanks!

MarkEWaite:
@car-roll I assume this does not require a new release but can be included in the next release whenever that happens.  Is that correct?

car-roll:
@MarkEWaite yes, that's correct. It was more about just not having someone else going through resolving those same issues I went through if I closed this PR completely

PR #1021, opened 2023-09-28
Remove `NettyDockerCmdExecFactoryCompat`

basil:
As far as I can tell nothing in the ecosystem is using this class anymore. This plugin has long ago switched from Netty to Apache HttpClient 5.x.


MarkEWaite (approved):
Thanks

PR #1116, opened 2024-11-13
Allow randomizing cloud selection

cavedon:
When multiple Docker clouds are defined, the plugin will always create agents from the first available cloud, not allowing to distribute the load on all the clouds that are defined.

Possible solutions to this issue are:
- Docker Swarm Standalone, which is deprecated
- Docker Engine Swarm API, which is not supported by docker-plugin.

This changes adds a setting that enables randomizing the order in which Docker clouds are tried when creating a new agent.

### Testing done

Tested on a Jenkins instance with two Docker clouds defined, with the setting enabled and disables.
I did not find existing unit tests coverage for FastNodeProvisionerStrategy where to add my tests...

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes


MarkEWaite:
I configured two Docker clouds, one a copy of the other, and watched the allocation.  As far as I could tell, it was always choosing the first cloud, whether or not the randomization checkbox was enabled.

krisstern:
Stale

PR #1125, opened 2025-01-25
Skip Docker-based tests when Docker is not available

basil:
Skip tests rather than failing when Docker is not available to allow this plugin to be included in BOM.

CC @darinpope


MarkEWaite (approved):
Thanks!

PR #1131, opened 2025-02-25
Avoid excessive `Closure`s

jglick:
Depends on https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/762.


src/main/resources/io/jenkins/docker/pipeline/DockerAgentScript.groovy
```diff
@@ -2,26 +2,24 @@ package io.jenkins.docker.pipeline
...
-        return {
-            try {
-                script.dockerNode(describable.asArgs) {
-                    CheckoutScript.doCheckout(script, describable, null, body).call()
-                }
-            } catch (Exception e) {
-                script.getProperty("currentBuild").result = Utils.getResultFromException(e)
-                throw e
+    void run(Closure body) {
+        try {
```
jglick, line 16: (ignore whitespace)

MarkEWaite (approved):
Thanks!  I've tested it in my relatively simple environment and confirmed that my declarative Pipeline samples work as expected.  They launch a container agent through the incremental of this change and report no error.

PR #1134, opened 2025-03-03
Compatibility for old `DockerAgentScript`

jglick:
Amending #1131 using https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/770. Verified interactively by starting an `agent dockerContainer` build using 1.9.0; upgrading to trunk and seeing that the build failed to resume; then resetting the build record and updating to this PR and seeing that it resumed without error.

### DIFF
--- a/src/main/java/io/jenkins/docker/pipeline/DockerAgent.java
+++ b/src/main/java/io/jenkins/docker/pipeline/DockerAgent.java
@@ -8,6 +8,7 @@ import hudson.model.Descriptor;
 import hudson.model.Item;
 import hudson.util.ListBoxModel;
 import io.jenkins.docker.connector.DockerComputerConnector;
+import java.net.URL;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


MarkEWaite:
The change looks very reasonable to me.  I didn't test it myself, but your description of the testing seems very good.  When it is ready for review, would you be OK if I asked the original submitter of the issue to test it?

jglick:
> the original submitter of the issue

Who would that be? I know of a complaint from someone using `docker-workflow`, but not this plugin.

MarkEWaite:
> Who would that be? I know of a complaint from someone using `docker-workflow`, but not this plugin.

My mistake.  I confused the two conditions.  Let me test this pull request so that we have a second verification it is behaving as expected.

jglick:
(I would say this is timely, but OTOH the relevant feature is currently marked as experimental.)

MarkEWaite (approved):
Thanks!  Merging for a release.

MarkEWaite:
Should release within the next hour.  Thanks again @jglick

MarkEWaite:
Released as [1274.vc0203fdf2e74](https://github.com/jenkinsci/docker-plugin/releases/tag/1274.vc0203fdf2e74)

src/main/java/io/jenkins/docker/pipeline/DockerAgent.java
```diff
@@ -127,4 +129,14 @@ public List<Descriptor<? extends DockerComputerConnector>> getAcceptableConnecto
                     .getAcceptableConnectorDescriptors();
         }
     }
+
+    @Extension(optional = true)
```
jglick, line 133: https://github.com/jenkinsci/docker-plugin/blob/511a5c372240f6f6ed68a4274e94a81bc69ceb95/src/main/java/io/jenkins/docker/pipeline/DockerAgent.java#L112-L113 might be wrong: `pipeline-model-extensions` is an optional dep. Use `@OptionalExtension` (also here on `Compat`)?

